PR #25, opened 2015-02-23
Layout test

mattfred:


ryancarlson:
@mattfred, is this one ready to go?

ryancarlson:
or ready for review, I should say?

mattfred:
Yes. I was waiting for Anne's final review, but now is fine. I need to add the option to turn off notifications as well.

mattfred:
Sorry this PR is kind of messy. I really think it is about time for this app to be rebuilt. It runs kind of slow and I think performance can be improved by rebuilding from scratch.

mattfred:
Another thing that would be improved by rebuilding would be to remove all of the unused logic in MainPW. Logic for switching languages is still there, which I believe some of is needed and some is not. Also translator logic is still there and it can be removed. I don't know that just going though and deleting the logic here is the best idea though, because it is used throughout and it no longer needs to. Rebuilding at some point would be much cleaner.

mattfred:
@ryancarlson I think this is finally done. I did merge the new splash icon branch into this one as well. It also fixes that display name issue I showed you earlier.

mattfred:
Oh I almost forgot. We do still need the new API endpoints to edit the drafts as well.

PR #147, opened 2017-01-24
Ryan dev

Yinten:


frett:
there wasn't a need to open a new PR, the discussion could have continued on the previous PR. I'm going to switch back to that one so we have the review history.

PR #154, opened 2017-02-08
Remove deprecated Apache using Retrofit2

Yinten:


src/main/java/org/keynote/godtools/android/api/converters/simplexml/SimpleXmlConverterFactory.java
```diff
@@ -0,0 +1,87 @@
...
+import retrofit2.Converter;
+import retrofit2.Retrofit;
+
+/**
+ * A {@linkplain Converter.Factory converter} which uses Simple Framework for XML.
+ * <p>
+ * This converter only applies for class types. Parameterized types (e.g., {@code List<Foo>}) are
+ * not handled.
+ */
+public final class SimpleXmlConverterFactory extends Converter.Factory {
```
frett: what was your reason for copying this converter into our source instead of just adding a dependency on the library?

`compile 'com.squareup.retrofit2:converter-simplexml:2.1.0`

https://github.com/square/retrofit/tree/master/retrofit-converters/simplexml

PR #314, opened 2017-08-09
Gt 364 Adobe Analytics

ryancarlson:


library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
```diff
@@ -42,8 +61,12 @@
     public static final String CATEGORY_CONTENT_EVENT = "Content Event";
     private Tracker mTracker = null;
 
+    private WeakReference<Activity> activity;
```
frett: Android naming convention for properties is:
`private WeakReference<Activity> mActivity`

library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
```diff
@@ -42,8 +61,12 @@
...
+    private String previousScreenName;
```
frett: `mPreviousScreenName`

frett:
I also just noticed that when you moved the config json to the analytics module, you didn't put it in an assets folder

ryancarlson:
Good catch on the config and folder

ryancarlson:
@frett made the suggested changes... waiting on travis to confirm checkstyles. Anything else you would like to see?

PR #316, opened 2017-08-21
Gt 421 us kgp share link

ryancarlson:
- Translates code from kgp-us to kgp for "Know God Personally"
- Adds page number to URLS for i > 0
- But no page numbers for US KGP

### DIFF
--- a/library/tract-renderer/src/main/java/org/cru/godtools/tract/model/Manifest.java
+++ b/library/tract-renderer/src/main/java/org/cru/godtools/tract/model/Manifest.java
@@ -114,6 +114,11 @@ public final class Manifest extends Base implements Styles {
     }
 
     @NonNull
+    public String getCodeForShareActivity() {
+        return mCode.equals("kgp-us") ? "kgp" : mCode;
+    }
+
+    @NonNull
     public Locale getLocale() {
         return mLocale;
     }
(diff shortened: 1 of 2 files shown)


library/tract-renderer/src/main/java/org/cru/godtools/tract/model/Manifest.java
```diff
@@ -114,6 +114,11 @@ public String getCode() {
     }
 
     @NonNull
+    public String getCodeForShareActivity() {
+        return mCode.equals("kgp-us") ? "kgp" : mCode;
```
frett: I think I want to keep this custom sharing override logic localized to the TractActivity for now.

ryancarlson:
@frett made the change you suggested. are these still the travis failures we had been seeing?

frett:
part of the travis log says checkstyle rule errors

frett:
```
[ant:checkstyle] [ERROR] /Users/frett/development/android/godtools-android/library/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java:601:35: String literal expressions should be on the left side of an equals comparison. [EqualsAvoidNull]
```

ryancarlson:
thanks!

PR #321, opened 2017-12-14
Fix <content:image> rendering

frett:
Depends on https://github.com/CruGlobal/android-gto-support/pull/167

we have no need of the `TractPicassoImageView` specific behavior, but we do need the underlying `SimplePicassoImageView` behavior.

### DIFF
--- a/library/tract-renderer/src/main/res/layout/tract_content_image.xml
+++ b/library/tract-renderer/src/main/res/layout/tract_content_image.xml
@@ -1,10 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
-<org.cru.godtools.tract.widget.TractPicassoImageView xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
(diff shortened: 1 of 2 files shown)


library/tract-renderer/src/main/res/layout/tract_content_image.xml
```diff
@@ -1,10 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
-<org.cru.godtools.tract.widget.TractPicassoImageView xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
+<org.ccci.gto.android.common.picasso.view.SimplePicassoImageView xmlns:android="http://schemas.android.com/apk/res/android"
     android:id="@+id/image"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
+    android:scaleType="centerCrop"
```
frett: when wrapping content, `scaleType="centerCrop"` will resize the image to fill the view, but not crop anything.

PR #322, opened 2017-12-21
track the SystemLanguage in Crashlytics

frett:


library/base-app/src/main/java/org/cru/godtools/base/app/BaseGodToolsApplication.java
```diff
@@ -23,5 +27,7 @@ public void onCreate() {
     private void initializeCrashlytics() {
         Fabric.with(this, new Crashlytics());
         Crashlytics.setBool("InstantApp", InstantApps.isInstantApp(this));
+        Crashlytics.setString("SystemLanguageRaw", Locale.getDefault().toString());
+        Crashlytics.setString("SystemLanguage", LocaleCompat.toLanguageTag(Locale.getDefault()));
```
frett: long term I'm going to leave `SystemLanguage` in place, but in case using `LocaleCompat.toLanguageTag()` causes a crash I'm also including a raw string first to hopefully catch what is wrong.

PR #323, opened 2017-12-21
Fix Travis

frett:


frett:
it looks like adobe analytics is causing a crash in the instant app code. I'll look into it at some point.

PR #327, opened 2018-02-06
GT-461 - Setup Snowplow Tracker

wrandall22:
This PR sets up the initial tracking for SnowPlow. It should track screen views and application crashes, as well as the EveryStudent search.

### DIFF
--- a/library/analytics/proguard-rules-adobe.pro
+++ b/library/analytics/proguard-rules-adobe.pro
@@ -1,5 +1,1 @@
 -dontwarn com.adobe.mobile.**
-
-# Rules for SnowPlow
--keep class com.google.android.gms.ads.identifier.AdvertisingIdClient {*;}
--keep class com.google.android.gms.ads.identifier.AdvertisingIdClient$Info {*;}
--- a/library/analytics/proguard-rules-adobe.pro
+++ b/library/analytics/proguard-rules-adobe.pro
@@ -1,1 +1,5 @@
 -dontwarn com.adobe.mobile.**
+
+# Rules for SnowPlow
+-keep class com.google.android.gms.ads.identifier.AdvertisingIdClient {*;}
+-keep class com.google.android.gms.ads.identifier.AdvertisingIdClient$Info {*;}
--- a/library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
+++ b/library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
@@ -136,12 +136,9 @@ public class AnalyticsService {
     @AnyThread
     private void trackScreenViewInSnowPlow(@NonNull final String screen)
     {
-        final Activity activity = mActiveActivity != null ? mActiveActivity.get() : null;
-        if (activity != null) {
-            mAnalyticsExecutor.execute(() -> {
-                mSnowPlowTracker.track(ScreenView.builder().name(screen).build());
-            });
-        }
+        mAnalyticsExecutor.execute(() -> {
+            mSnowPlowTracker.track(ScreenView.builder().name(screen).build());
+        });
     }
 
     /**
--- a/library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
+++ b/library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
@@ -17,6 +17,7 @@ import com.google.android.gms.analytics.Tracker;
 import com.snowplowanalytics.snowplow.tracker.DevicePlatforms;
 import com.snowplowanalytics.snowplow.tracker.Emitter;
 import com.snowplowanalytics.snowplow.tracker.emitter.RequestCallback;
+import com.snowplowanalytics.snowplow.tracker.events.ScreenView;
 import com.snowplowanalytics.snowplow.tracker.utils.LogLevel;
 
 import org.ccci.gto.android.common.compat.util.LocaleCompat;
@@ -78,6 +79,9 @@ public class AnalyticsService {
 
     /* Adobe Analytics */
     private final Executor mAdobeAnalyticsExecutor = Executors.newSingleThreadExecutor();
+    /* SnowPlow */
+    private final Executor mSnowPlowExecutor = Executors.newSingleThreadExecutor();
+
     @Nullable
     private WeakReference<Activity> mActiveActivity;
     private String mPreviousScreenName;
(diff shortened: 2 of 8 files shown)


wrandall22:
[Documentation for dependencies](https://github.com/snowplow/snowplow/wiki/Android-tracker-Setup)
[Documentation for API](https://github.com/snowplow/snowplow/wiki/Android-Tracker)

library/analytics/proguard-rules-adobe.pro
```diff
@@ -1 +1,5 @@
--dontwarn com.adobe.mobile.**
\ No newline at end of file
+-dontwarn com.adobe.mobile.**
+
+# Rules for SnowPlow
```
frett: split these out into a separate proguard file. it helps keep things organized as more libraries add varying rules.

frett: and you would just add a second `consumerProguardFiles` line under `defaultConfig`

library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
```diff
@@ -65,9 +76,13 @@
     public static final String CATEGORY_CONTENT_EVENT = "Content Event";
 
     private Tracker mTracker = null;
+    private com.snowplowanalytics.snowplow.tracker.Tracker mSnowPlowTracker = null;
 
     /* Adobe Analytics */
     private final Executor mAdobeAnalyticsExecutor = Executors.newSingleThreadExecutor();
+    /* SnowPlow */
+    private final Executor mSnowPlowExecutor = Executors.newSingleThreadExecutor();
```
frett: I'd be fine with using the same executor for both Adobe Analytics and SnowPlow. the main purpose of the executor is to just push processing off of the main thread and to serialize the events it's tracking.

library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
```diff
@@ -118,6 +136,18 @@ private void trackScreenViewInAdobe(@NonNull final String screen) {
         }
     }
 
+    @AnyThread
+    private void trackScreenViewInSnowPlow(@NonNull final String screen)
+    {
+        final Activity activity = mActiveActivity != null ? mActiveActivity.get() : null;
+        if (activity != null) {
```
frett: it looks like you don't need to check for the presence of activity at all.

library/everystudent/src/main/java/org/cru/godtools/everystudent/EveryStudent.java
```diff
@@ -126,6 +126,12 @@ public boolean onOptionsItemSelected(MenuItem item) {
         return false;
     }
 
+    @Override
+    protected void onResume() {
+        super.onResume();
+        mTracker.setActiveActivity(this);
```
frett: good catch here, it was an oversight since this was legacy code that wasn't part of the current common framework

frett:
it looks like SnowPlow doesn't require the Activity object at all. The current tracking of that in the AnalyticsService is to just make Adobe Analytics happy, it's not required by Google Analytics either.

library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
```diff
@@ -55,6 +61,11 @@
     private static final String ADOBE_NOT_LOGGED_IN = "not logged in";
     private static final String ADOBE_GODTOOLS = "GodTools";
 
+    /* SnowPlow value constants */
+    private static final String SNOWPLOW_APP_ID = "GodTools";
+    private static final String SNOWPLOW_NAMESPACE = "GodToolsSnowPlowAndroidTracker";
+    private static final String SNOWPLOW_ENDPOINT = "snowplow.cru.org";
```
frett: is there any desire to keep development tracking separate from release tracking?

library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
wrandall22: Not sure why I did this, at some point I was thinking that the activity was required somewhere, maybe I was thinking of [here](https://github.com/snowplow/snowplow/wiki/Android-Tracker#2315-sets-up-an-activity-lifecycle-handler)

library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
```diff
@@ -118,6 +134,18 @@ private void trackScreenViewInAdobe(@NonNull final String screen) {
...
     }
 
+    @AnyThread
+    private void trackScreenViewInSnowPlow(@NonNull final String screen)
+    {
+        final Activity activity = mActiveActivity != null ? mActiveActivity.get() : null;
+        if (activity != null) {
+            mAnalyticsExecutor.execute(() -> {
+                mSnowPlowTracker.track(ScreenView.builder().name(screen).build());
+                mPreviousScreenName = screen;
```
frett: `mPreviousScreenName` is something specific for Adobe Analytics. do you use it for Snowplow anywhere? if not I would just leave updating it in the adobe analytics executor task, and not have it here.

library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
wrandall22: There will be a stage environment, so we may want to do that for development. Currently there is one environment.

library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
wrandall22: Nope, I can remove it.

library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
frett: alright, let's put these as build config in the build script. let me find you a reference :)

library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
frett: k, let's drop the activity conditional around the task :)

library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
frett: well, at least the endpoint parameter :)

frett: https://github.com/CruGlobal/godtools-android/blob/develop/library/analytics/build.gradle#L9
https://github.com/CruGlobal/godtools-android/blob/develop/library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java#L76

frett:
also, run checkstyle against the code:
`./gradlew check --parallel`

PR #328, opened 2018-02-12
Remove "now" & "later" buttons in favor of one "Okay" button during onboarding tour

ryancarlson:
The "Now" button which took the user to the pick languages activity proved frustrating/distracting for users. Now just one "Okay" button.

### DIFF
--- a/app/src/main/res/layout/fragment_tour_languages.xml
+++ b/app/src/main/res/layout/fragment_tour_languages.xml
@@ -38,12 +38,6 @@
             android:layout_height="wrap_content">
 
             <Button
-                android:id="@+id/action_switch_language"
-                android:layout_width="wrap_content"
-                android:layout_height="wrap_content"
-                android:text="@string/action_tour_now" />
-
-            <Button
                 android:id="@+id/action_tour_close"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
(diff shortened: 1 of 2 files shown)


ryancarlson:
@frett can you review please?

ryancarlson:
`gradlew check` fails with 85 issues. i doubt those were all added in this PR :)

app/src/main/res/layout/fragment_tour_languages.xml
```diff
@@ -51,7 +45,7 @@
                 android:layout_marginStart="8dp"
                 android:layout_toEndOf="@id/action_switch_language"
                 android:layout_toRightOf="@id/action_switch_language"
```
frett: the layout of this button is now incorrect because you didn't remove layout attributes related to the removed button. Also, this leaves the `RelativeLayout` as useless, so you can remove it and put the remaining Button in it's place.

these attributes should be removed:
`layout_marginLeft`
`layout_marginStart`
`layout_toEndOf`
`layout_toRightOf`

frett: this is probably what triggered the fatal lint error

frett:
most of those errors are just warnings and not fatal, if the build is actually failing now there is a new fatal error introduced

app/src/main/res/layout/fragment_tour_languages.xml
ryancarlson: Thanks for explaining!

PR #329, opened 2018-02-12
Gt 379 call to action custom color

ryancarlson:
Use control-color attribute to color the arrow.
Falls back to page.primary-color and manifest.primary-color in that order

### DIFF
--- a/library/tract-renderer/src/main/java/org/cru/godtools/tract/model/CallToAction.java
+++ b/library/tract-renderer/src/main/java/org/cru/godtools/tract/model/CallToAction.java
@@ -36,7 +36,7 @@ public final class CallToAction extends Base {
     @Nullable
     private Text mLabel;
 
-    @Nullable
+    @Nullable @ColorInt
     private Integer mControlColor;
 
     @NonNull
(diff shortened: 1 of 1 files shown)


ryancarlson:
@frett

library/tract-renderer/src/main/java/org/cru/godtools/tract/model/CallToAction.java
```diff
@@ -34,6 +36,9 @@
     @Nullable
     private Text mLabel;
 
+    @Nullable
+    private Integer mControlColor;
```
frett: `@ColorInt`

it helps static analysis for lint checking

library/tract-renderer/src/main/java/org/cru/godtools/tract/model/CallToAction.java
```diff
@@ -43,6 +48,9 @@
 
     @ColorInt
     private static int getArrowColor(@Nullable final CallToAction callToAction) {
+        if (callToAction != null && callToAction.mControlColor != null) {
+            return callToAction.mControlColor;
+        }
         return Styles.getPrimaryColor(callToAction != null ? callToAction.getPage() : null);
     }
```
frett: I'd split this into 2 functions and rename to `getControlColor` now that we have an actual name for the attribute :)
```
private int getControlColor() {
    return mControlColor != null ? mControlColor : Styles.getPrimaryColor(getPage());
}

private int getControlColor(@Nullable final CallToAction callToAction) {
    return callToAction != null ? callToAction.getControlColor() : Styles.getPrimaryColor(null);
}
```

ryancarlson: 👍

PR #331, opened 2018-02-23
disable www.knowgod.com deep links to fix app link behavior

frett:
https://www.knowgod.com/.well-known/assetlinks.json issues a 301 redirect. This is invalid according to the verification file requirements listed here: https://developer.android.com/training/app-links/verify-site-associations.html#publish-json

### DIFF
--- a/library/tract-renderer/src/main/AndroidManifest.xml
+++ b/library/tract-renderer/src/main/AndroidManifest.xml
@@ -15,7 +15,7 @@
 
                 <data android:scheme="http" />
                 <data android:scheme="https" />
-                <data android:host="@string/tract_deeplink_host_1" />
+                <!--<data android:host="@string/tract_deeplink_host_1" />-->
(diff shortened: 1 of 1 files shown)


ryancarlson:
so ..string1 redirects, but ..string2 does not?

frett:
host1 is `www.knowgod.com`
host2 is `knowgod.com`

www.knowgod.com is a 301 redirect to knowgod.com. knowgod.com responds with the actual file. the 301 redirect on www.knowgod.com doesn't meet the validation requirements.

the other option is to configure https://www.knowgod.com/.well-known/assetlinks.json to directly return the file. But that would take some reconfiguration of S3 and cloudfront since `www.knowgod.com` is set to redirect everything to `knowgod.com`, and there isn't a direct way to whitelist a single file as far as I could tell in the 5 mins of research I did.

ryancarlson:
👍

PR #344, opened 2018-04-12
deploy test builds to fabric beta

frett:


.travis.yml
```diff
@@ -7,11 +7,15 @@ env:
...
+    # FABRIC_API_SECRET={API Key}
+    - secure: QmSJoNcdg6djHWmtACAjWcANnvibicZFwu8EDYw6d3BTV9s5q8IFOh4ay6jFYjBKWKUNbdMwqn43Z9MBZTB+G4z0hCuemxdETX5yXKNsoQRORzeNZCzWmcWjAEy2ZfvccDoTINai+Gz1ckFiWUV++ONK7NPzPE5zeuQA4kPKb9w=
+    # BETA_KEYSTORE_PASSWORD={password}
+    - secure: ZZREqPHD1TMUgfHUMws3EVy1f0q/qSkdcowSEXb4S//EGolDgkDCL0oWAz9TzKe6RGjj+PQBzY66JgCC5UJ7gT+RxOgSetQwuR+thQ/bW6P1zH5PmiAxfB28jBXC2OpTX7Ad7G2iOjMMFj4WwmfKX/dPP8/WUGhWtGPaF/oEfms=
   matrix:
     - ANDROID_TARGET=android-15 ANDROID_ABI=armeabi-v7a
 #    - ANDROID_TARGET=android-21 ANDROID_ABI=armeabi-v7a
 #    - ANDROID_TARGET=android-23 ANDROID_ABI=armeabi-v7a
-    - ANDROID_TARGET=android-24 ANDROID_ABI=armeabi-v7a
+    - ANDROID_TARGET=android-24 ANDROID_ABI=armeabi-v7a DEPLOY_BETA=true
```
ryancarlson: any reason to not build on android-26? This is now two versions behind, I think.

frett: it's actually 3 versions behind. But, the arm test images are only available up through android-24 the last time I checked.

gradle.properties
```diff
@@ -14,3 +14,9 @@ androidKeystorePath=non-existant-keystore-dont-create-me.store
 androidKeystoreStorePassword=
 androidKeystoreKeyAlias=godtools
 androidKeystoreKeyPassword=
+
+# crashlytics beta signing config
+crashlyticsBetaKeystorePath=crashlytics/beta.keystore
+crashlyticsBetaKeystoreStorePassword=
+crashlyticsBetaKeystoreKeyAlias=godtools
+crashlyticsBetaKeystoreKeyPassword=
```
ryancarlson: The echo commands above will fill this value in.. or overwrite the key/value? My initial guess is that the echo would add extra lines.

frett: neither, it actually writes to ~/.gradle/gradle.properties

settings in this file override settings in local projects

frett: the blank lines are here to allow the build script to compile without missing properties

.travis.yml
ryancarlson: Ah, okay. Can you check and bump it if the later ones are available?

gradle.properties
ryancarlson: got it, thanks

.travis.yml
frett: there is a google apis ARM system image for android-25. I'd have to update some of the build script to support using that variant, I'll work that up in another PR.

system images for newer versions of android are x86 based, but those images don't work in travis due to the travis architecture.

.travis.yml
```diff
@@ -63,5 +67,20 @@ after_failure:
   - cat /home/travis/build/CruGlobal/godtools-android/build/reports/lint-results*.xml
   - cat /home/travis/build/CruGlobal/godtools-android/*/build/reports/lint-results*.xml
 
+before_deploy:
+  - echo crashlyticsBetaKeystoreStorePassword=$BETA_KEYSTORE_PASSWORD >> .gradle/gradle.properties
+  - echo crashlyticsBetaKeystoreKeyPassword=$BETA_KEYSTORE_PASSWORD >> .gradle/gradle.properties
```
frett: I did mis-type the target file :) let me fix that

.travis.yml
ryancarlson: sounds good

PR #346, opened 2018-04-12
Track several Analytics actions

frett:
Actions tracked
- Share tract
- clicking URL buttons
- toggling between languages when viewing a tract

### DIFF
--- a/library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
+++ b/library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
@@ -32,6 +32,7 @@ public interface AnalyticsService {
     /* Action event names */
     String ACTION_SHARE = "Share Icon Engaged";
     String ACTION_EXIT_LINK = "Exit Link Engaged";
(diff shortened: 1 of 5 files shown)


library/analytics/src/main/java/org/cru/godtools/analytics/AnalyticsService.java
```diff
@@ -27,9 +29,10 @@
     String SCREEN_PRIVACY_POLICY = "Privacy Policy";
     String SCREEN_COPYRIGHT = "Copyright Info";
 
-    /* Custom dimensions */
-    int DIMENSION_TOOL = 1;
-    int DIMENSION_LANGUAGE = 2;
+    /* Action event names */
+    String ACTION_SHARE = "Share Icon Engaged";
```
ryancarlson: Is there any reason to not make these and the ones above `static final` Strings?

ryancarlson: I wondered about that right after I posted the comment. It's been awhile since I've done java :)

PR #350, opened 2018-04-16
Deep Link updates for JF

frett:
Within the Android TractActivity we already store an array of languages we are attempting to show. This PR leverages this arbitrary list of languages to display to implement the primary language fallback functionality.

1. We parse a list of primary language candidates (with fallbacks) from the deep link
2. We trigger a background download for all of these potential languages
3. We pick which of the primary language options we actually make visible in the UI via `updateHiddenLanguages()`
   - If one of the potential primary languages is currently active we use that as the primary language. (This prevents unexpected state changes in the UI for the end user)
   - Otherwise the first potential primary language that we haven't determined to be missing is used.
   - This state is refreshed every time we refresh the language toggle UI.
     - This automatically happens whenever some new data is made available to the TractActivity
- We also update which language is the active language if the current active language is not found

### DIFF
--- a/library/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
+++ b/library/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
@@ -21,6 +21,7 @@ import android.support.v4.view.ViewPager;
 import android.support.v7.app.ActionBar;
 import android.support.v7.content.res.AppCompatResources;
 import android.support.v7.widget.Toolbar;
+import android.text.TextUtils;
 import android.util.SparseArray;
 import android.view.Menu;
 import android.view.MenuItem;
@@ -60,6 +61,7 @@ import org.keynote.godtools.android.db.GodToolsDao;
 import org.keynote.godtools.android.model.Tool;
 
 import java.io.IOException;
+import java.util.List;
 import java.util.Locale;
 
 import butterknife.BindView;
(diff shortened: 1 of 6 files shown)


library/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
```diff
@@ -320,8 +342,26 @@ private boolean isDeepLinkValid(@Nullable final Uri data) {
...
+        final Locale uriLocale = LocaleCompat.forLanguageTag(data.getPathSegments().get(0));
+        final List<Locale> rawPrimaryLanguages = Stream.of(data.getQueryParameters(PARAM_PRIMARY_LANGUAGE))
+                .flatMap(lang -> Stream.of(TextUtils.split(lang, ",")))
+                .map(String::trim)
+                .filterNot(TextUtils::isEmpty)
+                .map(LocaleCompat::forLanguageTag)
+                .toList();
+
+        // process the primary languages specified in the uri
+        if (!rawPrimaryLanguages.isEmpty()) {
```
ryancarlson: Preference: This may be a swift/ruby convention, but I now prefer in a case like this to flip the logic check here and in the case that `rawPrimaryLanguages` is empty, return the uriLocal. It indents less code is more readable in my opinion

library/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
```diff
@@ -335,34 +375,74 @@ private boolean validStartState() {
...
+                    mHiddenLanguages[i] = false;
+
+                    // hide any previously identified primary languages
+                    if (primaryLanguage != -1) {
+                        mHiddenLanguages[primaryLanguage] = true;
+                    }
+
+                    // track our current primary language
+                    primaryLanguage = i;
+                } else if (i < mPrimaryLanguages) {
```
ryancarlson: How could this block get called? `i >= mPrimaryLanguages` would be gated on line 414.

ryancarlson: Never mind.. sleep deprivation

PR #353, opened 2018-04-17
Reduce indentation by continuing early

ryancarlson:


frett:
I took the `state == STATE_NOT_FOUND` early continue option, going to keep the primary language logic in it's own if block to support parallel language fallbacks if we ever implement them.

PR #364, opened 2018-04-25
tweak some styles for the parallel language, and extract the styling to a common style

frett:
For styles I try to keep layout/positioning within the view XML, but extract other style properties to the styles xml


frett:
screenshot_20180425-134402

PR #365, opened 2018-04-25
reduce minimum fling velocity for card swipe vs scroll

frett:


frett:
https://jira.cru.org/browse/GT-431

PR #371, opened 2018-05-03
Language Settings Feature discovery

frett:
The Language Settings feature discovery triggers after 10 seconds on the home screen if it hasn't triggered before and if the user hasn't already visited Language Settings or set a Parallel Language.

Feature Discovery is setup in a manner that will allow us to provide more context sensitive prompts in the future as we want to highlight feature usage for a user.

### DIFF
--- a/app/src/main/java/org/keynote/godtools/android/activity/MainActivity.java
+++ b/app/src/main/java/org/keynote/godtools/android/activity/MainActivity.java
@@ -336,6 +336,10 @@ public class MainActivity extends BasePlatformActivity implements ToolsFragment.
             case FEATURE_LANGUAGE_SETTINGS:
                 assert mToolbar != null : "canShowFeatureDiscovery() verifies mToolbar is not null";
                 if (mToolbar.findViewById(R.id.action_switch_language) != null) {
+                    // purge any pending feature discovery triggers since we are showing feature discovery now
+                    purgeQueuedFeatureDiscovery(FEATURE_LANGUAGE_SETTINGS);
+
+                    // show language settings feature discovery
                     final TapTarget target = TapTarget.forToolbarMenuItem(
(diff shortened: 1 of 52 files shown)


frett:
I just realized this may show over the navigation drawer. I need to refactor slightly to make sure it doesn't display if they have the nav drawer open.

frett:
Alright, it now won't display if the nav drawer is open.

app/src/main/java/org/keynote/godtools/android/activity/MainActivity.java
```diff
@@ -245,10 +266,97 @@ private void loadInitialFragmentIfNeeded() {
...
+                    mFeatureDiscovery =
+                            TapTargetView.showFor(this, target, new LanguageSettingsFeatureDiscoveryListener());
+                    mFeatureDiscoveryActive = feature;
+                } else {
+                    // TODO: we currently don't (can't?) distinguish between when the menu item doesn't exist and when
+                    // TODO: the menu item just hasn't been drawn yet.
+
+                    // the toolbar action isn't available yet.
+                    // re-attempt this feature discovery on the next iteration of the main Looper.
+                    mToolbar.post(() -> showFeatureDiscovery(feature, force));
```
ryancarlson: Could this end up in an infinite loop type situation? I don't see any "sanity check" variable to make sure this doesn't get called more than 'n' times.

frett: it'll end up in an infinite loop as long as the activity is resumed and the language settings button never appears. But I believe in that scenario it will fire only once per frame.

frett: I was originally trying to find a callback to tie into that fired after the action button was available, but I couldn't find one.

PR #437, opened 2018-08-01
Added Search functionality to God Tools

gyasistory:
With this changed I just added a searchView and updated the adapter with the search results.  I wasn't able to get the stream to filter my list properly so I just used a for each loop and recreated a new list to pass to the adapter.  Looking forward to any feed back you can provide.

### DIFF
--- a/app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
+++ b/app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
@@ -1,5 +1,6 @@
 package org.keynote.godtools.android.fragment;
 
+import android.app.SearchManager;
 import android.content.Context;
 import android.os.Bundle;
 import android.support.annotation.CallSuper;
@@ -10,7 +11,10 @@ import android.support.v4.content.Loader;
 import android.support.v7.widget.DividerItemDecoration;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
+import android.support.v7.widget.SearchView;
 import android.view.LayoutInflater;
+import android.view.Menu;
+import android.view.MenuInflater;
 import android.view.View;
 import android.view.ViewGroup;
 
@@ -24,6 +28,7 @@ import org.cru.godtools.model.Language;
 import org.cru.godtools.sync.GodToolsSyncService;
 import org.keynote.godtools.android.content.LanguagesLoader;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 
@@ -59,6 +64,7 @@ public class LanguagesFragment extends BaseFragment implements LanguagesAdapter.
     }
 
     /* BEGIN lifecycle */
+    //region BEGIN lifecycle
 
     @Override
     public void onCreate(@Nullable final Bundle savedInstanceState) {
@@ -68,6 +74,9 @@ public class LanguagesFragment extends BaseFragment implements LanguagesAdapter.
         if (args != null) {
             mPrimary = args.getBoolean(EXTRA_PRIMARY, mPrimary);
         }
+
+        // Allow Fragment to have its own menu
+        setHasOptionsMenu(true);
 
         startLoaders();
     }
@@ -98,12 +107,16 @@ public class LanguagesFragment extends BaseFragment implements LanguagesAdapter.
         if (languages == null) {
             mLanguages = null;
         } else {
-            mLanguages = Stream.of(languages)
-                    .sorted((l1, l2) -> l1.getDisplayName().compareToIgnoreCase(l2.getDisplayName()))
-                    .toList();
+            mLanguages = sortLanguages(languages);
         }
 
         updateLanguagesList();
+    }
+
+    private List<Language> sortLanguages(@NonNull List<Language> languages) {
+        return Stream.of(languages)
+                .sorted((l1, l2) -> l1.getDisplayName().compareToIgnoreCase(l2.getDisplayName()))
+                .toList();
     }
 
     @Override
@@ -120,6 +133,61 @@ public class LanguagesFragment extends BaseFragment implements LanguagesAdapter.
         super.onDestroyView();
     }
 
+    //region Language Search
+
+    @Override
+    public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
+        super.onCreateOptionsMenu(menu, inflater);
+
+        inflater.inflate(R.menu.fragment_language_search, menu);
+
+        // Configuring the SearchView
+        SearchManager searchManager = (SearchManager) getActivity().getSystemService(Context.SEARCH_SERVICE);
+        SearchView searchView = (SearchView) menu.findItem(R.id.app_bar_language_search).getActionView();
+        searchView.setQueryHint(getString(R.string.label_language_search));
+        searchView.setSearchableInfo(searchManager.getSearchableInfo(getActivity().getComponentName()));
+
+        // Will listen for search event and trigger
+        searchView.setOnQueryTextListener(new SearchView.OnQueryTextListener() {
+            @Override
+            public boolean onQueryTextSubmit(String query) {
+                updateLanguageWithSearch(query);
+                return false;
+            }
+
+            @Override
+            public boolean onQueryTextChange(String newText) {
+                updateLanguageWithSearch(newText);
+                return false;
+            }
+        });
+    }
+
+    private void updateLanguageWithSearch(String query) {
+
+        List<Language> queryList = new ArrayList<>();
+
+        // Iterate through list and create new list for adapter
+        if (mLanguages != null) {
+            for(Language language: mLanguages){
+
+                if (language.getDisplayName().toLowerCase().contains(query.toLowerCase())){
+                    queryList.add(language);
+                }
+            }
+        }
+
+        if (mLanguagesAdapter != null) {
+            mLanguagesAdapter.setLanguages(
+                    sortLanguages(queryList)
+            );
+        }
+    }
+
+
+    //endregion
+
+    //endregion
     /* END lifecycle */
 
     private void startLoaders() {
--- a/app/src/main/res/menu/fragment_language_search.xml
+++ b/app/src/main/res/menu/fragment_language_search.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <item
+        android:id="@+id/app_bar_language_search"
+        android:icon="@drawable/ic_search_black_24dp"
+        android:title="Search"
+        app:showAsAction="always|collapseActionView"
+        app:actionViewClass="android.support.v7.widget.SearchView" />
+</menu>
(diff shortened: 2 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -58,6 +58,7 @@
     <string name="text_language_parallel_description">Share GodTools with someone in their native language.</string>
     <string name="text_language_warning">Not every tool is available in every language.</string>
     <string name="label_language_none">None</string>
+    <string name="label_language_search">Enter Parallel Language</string>
```
frett: This will probably be adjusted again, but for now let's change this to "Filter Languages"

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -120,6 +133,61 @@ public void onDestroyView() {
...
+            mLanguagesAdapter.setLanguages(
+                    sortLanguages(queryList)
+            );
+        }
+    }
+
+
+    //endregion
+
+    //endregion
```
frett: I'm guessing there is something inside of Android Studio that supports regions? if so I'll want to see it to better utilize it :)

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -69,6 +75,9 @@ public void onCreate(@Nullable final Bundle savedInstanceState) {
             mPrimary = args.getBoolean(EXTRA_PRIMARY, mPrimary);
         }
 
+        // Allow Fragment to have its own menu
+        setHasOptionsMenu(true);
```
frett: minor nit, my style has usually been to put this call directly under the `super.onCreate()` line.

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -120,6 +133,61 @@ public void onDestroyView() {
...
+        }
+
+        if (mLanguagesAdapter != null) {
+            mLanguagesAdapter.setLanguages(
+                    sortLanguages(queryList)
+            );
+        }
+    }
+
+
```
frett: I think checkstyle will also complain about multiple empty lines next to each other

app/src/main/res/menu/fragment_language_search.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <item
+        android:id="@+id/app_bar_language_search"
+        android:icon="@drawable/ic_search_black_24dp"
+        android:title="Search"
```
frett: need to use the string resource you created here :)

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -120,6 +133,61 @@ public void onDestroyView() {
...
+                }
+            }
+        }
+
+        if (mLanguagesAdapter != null) {
+            mLanguagesAdapter.setLanguages(
+                    sortLanguages(queryList)
+            );
+        }
+    }
```
frett: I actually would try to integrate setting a filtered list into the `updateLanguagesList()` method. doing the logic there will address a potential UI bug where a user filters the list, then languages get reloaded due to sync completing in the background.

frett: how I would probably do it is add a `mQuery` variable under `mLanguages` to track the current search term, then within `updateLanguagesList()` trigger a filter languages method that applies the query before setting the languages to the adapter.

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
gyasistory: Yes regions allow you to collapse chucks of code.

frett: I just tried it out in AS, learned something new today :)

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
frett: Also, you would want to preserve/restore `mQuery` on save/restore state

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -120,6 +133,61 @@ public void onDestroyView() {
...
+    public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
+        super.onCreateOptionsMenu(menu, inflater);
+
+        inflater.inflate(R.menu.fragment_language_search, menu);
+
+        // Configuring the SearchView
+        SearchManager searchManager = (SearchManager) getActivity().getSystemService(Context.SEARCH_SERVICE);
+        SearchView searchView = (SearchView) menu.findItem(R.id.app_bar_language_search).getActionView();
+        searchView.setQueryHint(getString(R.string.label_language_search));
+        searchView.setSearchableInfo(searchManager.getSearchableInfo(getActivity().getComponentName()));
```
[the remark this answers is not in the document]
gyasistory: I can remove it I don't think it plays any significant role for this application.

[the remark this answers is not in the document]
frett: k, I would guess `SearchManager` is used to populate the device search index, which we won't be using for just filtering languages

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -63,12 +78,18 @@ public static Fragment newInstance(final boolean primary) {
...
+        // Allow Fragment to have its own menu
+        setHasOptionsMenu(true);
         final Bundle args = getArguments();
         if (args != null) {
             mPrimary = args.getBoolean(EXTRA_PRIMARY, mPrimary);
         }
-
+        if (savedInstanceState != null) {
+            if (savedInstanceState.containsKey(KEY_SEARCH))
+                mQuery = savedInstanceState.getString(KEY_SEARCH);
```
frett: 2 comments:
- always use braces for if blocks, it helps guard against accidentally executing a statement outside of an if block. I believe there is a checkstyle rule about this.
- in this case if statements shouldn't be required, you can use the `getString` and `getBoolean` methods with default values to simplify this logic.
```
mQuery = savedInstanceState.getString(KEY_SEARCH, mQuery);
mIsSearchViewOpen = savedInstanceState.getBoolean(KEY_IS_SEARCH_OPEN, mIsSearchViewOpen);
```

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -120,6 +145,75 @@ public void onDestroyView() {
...
+                if (language.getDisplayName().toLowerCase().contains(query.toLowerCase())) {
+                    queryList.add(language);
+                }
+            }
+        }
+        return queryList;
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
```
frett: I'm guessing you added this for debugging only, it shouldn't be necessary anymore :)

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -149,12 +243,22 @@ private void setupLanguagesList() {
...
-            mLanguagesAdapter.setLanguages(mLanguages);
+            mLanguagesAdapter.setLanguages(filterLang(mLanguages));
             mLanguagesAdapter.setDisabled(mPrimary ? null : mPrimaryLanguage);
             mLanguagesAdapter.setProtected(mSettings != null ? mSettings.getProtectedLanguages() : null);
         }
     }
 
+    // This will check to see if the language needs to be filtered by search
+    private List<Language> filterLang(List<Language> languages) {
+       if (mQuery == null || mQuery.isEmpty()) {
```
frett: just a minor nit, another option here is `TextUtils.isEmpty(mQuery)`

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -98,14 +119,18 @@ void onLoadLanguages(@Nullable final List<Language> languages) {
         if (languages == null) {
             mLanguages = null;
         } else {
-            mLanguages = Stream.of(languages)
-                    .sorted((l1, l2) -> l1.getDisplayName().compareToIgnoreCase(l2.getDisplayName()))
-                    .toList();
+            mLanguages = sortLanguages(languages);
```
frett: this change is no longer necessary and can be reverted, the `sortLanguages` logic is only triggered here.

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -120,6 +145,75 @@ public void onDestroyView() {
...
+        setSearchView(menu);
+
+    }
+
+    private void setSearchView(Menu menu) {
+        // Configuring the SearchView
+        mSearchItem = menu.findItem(R.id.app_bar_language_search);
+        SearchView mSearchView = (SearchView) mSearchItem.getActionView();
+        mSearchView.setQueryHint(getString(R.string.label_language_search));
+        if (mIsSearchViewOpen) {
```
frett: 👍

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -120,6 +145,75 @@ public void onDestroyView() {
...
+                return false;
+            }
+        });
+    }
+
+    private List<Language> updateLanguageWithSearch(String query) {
+        List<Language> queryList = new ArrayList<>();
+        // Iterate through list and create new list for adapter
+        if (mLanguages != null) {
+            for (Language language : mLanguages) {
```
frett: if you wanted to use a stream I believe this is how it would be written:
```
return Stream.of(mLanguages)
    .filter(l -> l.getDisplayName().toLowerCase().contains(query.toLowerCase()))
    .toList()
```

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
gyasistory: Yeah I need to remove sorry

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
gyasistory: This is what I used and it didn't  work for me.  It keep giving me an empty list.

gyasistory: OK well it is working now. Java is moody sometimes.

app/src/main/java/org/keynote/godtools/android/fragment/LanguagesFragment.java
```diff
@@ -120,6 +137,61 @@ public void onDestroyView() {
...
+            public boolean onQueryTextChange(String newText) {
+                mQuery = newText;
+                updateLanguagesList();
+                return false;
+            }
+        });
+    }
+
+    private List<Language> updateLanguageWithSearch(final String query) {
+        return Stream.of(mLanguages)
```
frett: need to protect against `mLanguages == null`

PR #438, opened 2018-08-03
Just add LaunchMode to Manifiest to prevent duplicate Activities.

gyasistory:


frett:
hmm, I know I mentioned launch modes originally, but actually glancing at the launch mode meanings here: https://developer.android.com/guide/components/activities/tasks-and-back-stack#TaskLaunchModes

it seems that `singleInstance` would lead to the about and settings activities to be their own independent task stack, which is not what we would want.

frett:
I think we might want `singleTask` for those 2 activities

frett:
We would also want that on the main activity if it's not already there. We also need to double check that we are handling the `onNewIntent()` callback if necessary. I don't believe any of those activities need to do anything for it.

frett:
Sanely managing the back stack has always been interesting...

gyasistory:
Will do,

gyasistory:
MainActivity is set to `singleTop` would you like me to change it to `singleTask`

frett:
yeah, set MainActivity to singleTask. I think I finally understand the differences between the different launch modes (hopefully)

PR #439, opened 2018-08-07
Fix Snowplow Screen View URIs

wrandall22:
The `screen_view` portion of the URI was being used as the `host`, in which underscore is an invalid character. This led to the data not getting through to Redshift. This should fix any data we are sending going forward. The other data will need to be fixed using SQL.


gyasistory:
It looks like this was pulled from an older build.  I can't run the application because of a conflict with crashlytics.  

`Android dependency 'com.crashlytics.sdk.android:crashlytics' has different version for the compile (2.9.3) and runtime (2.9.4) classpath. You should manually set the same version via DependencyResolution
`

wrandall22:
Created branch fix-screen-view-uris-dev that branches off of develop

gyasistory:
Yes I think that is is the issue.

wrandall22:
See #440 instead.

PR #441, opened 2018-08-09
Fix Snowplow URIs

wrandall22:
This PR fixes an issue where we were sending URIs like `godtools:/screenview/...` to snowplow instead of `godtools://screenview/...`.


wrandall22:
Thanks Ryan!

PR #442, opened 2018-08-13
Log In for English only

gyasistory:
updated BasePlatformActivity to display the log in for English users only.

Requested by Ann

[GT-511](https://jira.cru.org/browse/GT-511)

### DIFF
--- a/app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
+++ b/app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
@@ -30,7 +30,6 @@ import org.cru.godtools.analytics.model.AnalyticsScreenEvent;
 import org.cru.godtools.base.Settings;
 import org.cru.godtools.base.ui.activity.BaseDesignActivity;
 import org.cru.godtools.base.ui.util.WebUrlLauncher;
-import org.cru.godtools.tract.model.Resource;
 import org.greenrobot.eventbus.Subscribe;
 import org.greenrobot.eventbus.ThreadMode;
 import org.keynote.godtools.android.activity.MainActivity;
@@ -254,7 +253,7 @@ public abstract class BasePlatformActivity extends BaseDesignActivity
      */
     private void updateNavigationDrawerMenu() {
         if (!ConfigurationCompat.getLocales(Resources.getSystem().getConfiguration())
-                .toLanguageTags().startsWith("en")){  // For non English
+                .toLanguageTags().startsWith("en")) {  // For non English
             if (mLoginItem != null) {
                 mLoginItem.setVisible(false);
             }
@@ -265,7 +264,7 @@ public abstract class BasePlatformActivity extends BaseDesignActivity
                 mLogoutItem.setVisible(false);
             }
 
-        }else { // For English 
+        } else { // For English
             if (mLoginItem != null) {
                 mLoginItem.setVisible(mTheKey.getDefaultSessionGuid() == null);
             }
--- a/app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
+++ b/app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
@@ -4,6 +4,7 @@ import android.arch.lifecycle.Lifecycle;
 import android.content.ActivityNotFoundException;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.content.res.Resources;
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.annotation.CallSuper;
@@ -12,6 +13,7 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.design.widget.NavigationView;
 import android.support.v4.app.FragmentManager;
+import android.support.v4.os.ConfigurationCompat;
 import android.support.v4.view.GravityCompat;
 import android.support.v4.widget.DrawerLayout;
 import android.support.v7.app.ActionBarDrawerToggle;
@@ -28,6 +30,7 @@ import org.cru.godtools.analytics.model.AnalyticsScreenEvent;
 import org.cru.godtools.base.Settings;
 import org.cru.godtools.base.ui.activity.BaseDesignActivity;
 import org.cru.godtools.base.ui.util.WebUrlLauncher;
+import org.cru.godtools.tract.model.Resource;
 import org.greenrobot.eventbus.Subscribe;
 import org.greenrobot.eventbus.ThreadMode;
 import org.keynote.godtools.android.activity.MainActivity;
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -13,7 +13,7 @@ buildscript {
         mavenLocal()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.3'
+        classpath 'com.android.tools.build:gradle:3.2.+'
         classpath 'com.github.dcendents:android-maven-gradle-plugin:1.5'
```
gyasistory: Had to update Gradle to get Lint Check to work.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Mon Aug 13 17:10:37 EDT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.5-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.6-all.zip
```
gyasistory: Again this was to get the Lint Check to work.

ui/everystudent/src/main/java/org/cru/godtools/everystudent/EveryStudentView.java
```diff
@@ -51,7 +51,10 @@ public void onCreate(Bundle savedInstanceState) {
         SharedPreferences settings = getSharedPreferences(Constants.PREFS_SETTINGS, 0);
         if (settings.getBoolean("wakelock", true)) {
             PowerManager pm = (PowerManager) getSystemService(Context.POWER_SERVICE);
-            wl = pm.newWakeLock(PowerManager.FULL_WAKE_LOCK, "DoNotDimScreen");
+            // Lint Check requires a prefix. I will use using gdt: - Gyasi Story
```
ryancarlson: Is there some significance to "gdt"? Just my own curiosity.

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
```diff
@@ -242,15 +244,36 @@ private void setupNavigationDrawer() {
...
-        if (mLoginItem != null) {
-            mLoginItem.setVisible(mTheKey.getDefaultSessionGuid() == null);
-        }
-        if (mSignupItem != null) {
-            mSignupItem.setVisible(mTheKey.getDefaultSessionGuid() == null);
-        }
-        if (mLogoutItem != null) {
-            mLogoutItem.setVisible(mTheKey.getDefaultSessionGuid() != null);
+        if (!ConfigurationCompat.getLocales(Resources.getSystem().getConfiguration())
+                .toLanguageTags().startsWith("en")) {  // For non English
```
ryancarlson: Is it possible to create a constant ArrayList of language codes that support login? I had Greg create an Array and hard code "en" in it. Then he just checked if that array contained the current device language. That allows us over time to just add more things into the array rather than adding cases to an if/else or switch statement.

I realize this may be a little different b/c you're doing an `startsWith("")` instead of an exact match, but if there's a way to pull the hard coded "en" out of here into a private constant member Array that would be nice.

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
```diff
@@ -242,15 +244,36 @@ private void setupNavigationDrawer() {
...
+            if (mSignupItem != null) {
+                mSignupItem.setVisible(false);
+            }
+            if (mLogoutItem != null) {
+                mLogoutItem.setVisible(false);
+            }
+
+        } else { // For English
+            if (mLoginItem != null) {
+                mLoginItem.setVisible(mTheKey.getDefaultSessionGuid() == null);
```
ryancarlson: Style note: I would refactor out a local boolean called something like `isLoggedIn` and set it to the result of `mTheKey.getDefaultSessionGuid() == null`. That prevents the same logic's being evaluated three times :)

ui/everystudent/src/main/java/org/cru/godtools/everystudent/EveryStudentView.java
gyasistory: No I was thing of a shorten prefix for GodTools.  It can be changed if you link.

ryancarlson: Nope, I think it's fine. I just wasn't sure if it was an abbreviation for something in the Android SDK or something else.

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
gyasistory: The reason I did it this way is because the call that I used returns a the language and region as a collection in a string.  So the return value will be some thing like "en-US, es-US".  The first language is the main default language and the second is the secondary language.  I did the startsWith("en") to account for all the regions that speak english ("en-uk", etc..).

I will do some research to see if there is a better way to get the default Language, this way I don't have to rely on a string comparison.

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
gyasistory: Good point.

gyasistory:
@ryancarlson I am going to finish the PR for Jeff and then jump on these changes, thanks for the feed back.

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
```diff
@@ -242,16 +247,65 @@ private void setupNavigationDrawer() {
...
+     * and display log in links based on users status.
+     * Updated by:  Gyasi Story
+     */
     private void updateNavigationDrawerMenu() {
-        if (mLoginItem != null) {
-            mLoginItem.setVisible(mTheKey.getDefaultSessionGuid() == null);
-        }
-        if (mSignupItem != null) {
-            mSignupItem.setVisible(mTheKey.getDefaultSessionGuid() == null);
+        if (!containsLoginLanguage()) {  // For non English
```
gyasistory: I extracted the language check out to method to make it flexible in the future.

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
```diff
@@ -242,16 +247,65 @@ private void setupNavigationDrawer() {
...
+     * get more support for more languages you can add them to {@code languages}.
+     *
+     * @return Boolean value if login Language
+     *
+     * Updated by: Gyasi Story
+     */
+    private boolean containsLoginLanguage() {
+        // List of supported Login Languages
+        String[] languages = {
+                Locale.ENGLISH.getLanguage()
```
gyasistory: As our notification starts to support more languages we can just add them here.

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
```diff
@@ -242,16 +247,65 @@ private void setupNavigationDrawer() {
...
+        for (int i = 0; i < localeListCompat.size(); i++) {
+            for (String lang : languages) {
+                if (localeListCompat.get(i).getLanguage().equals(lang)) {
+                    return true;
+                }
+            }
         }
+
+        return false;
+
```
frett: would this work instead?
```
LocaleListCompat localeListCompat = ConfigurationCompat
                .getLocales(Resources.getSystem().getConfiguration());
return localeListCompat.getFirstMatch(languages) != null;
```

gyasistory: @frett That didn't work because it default to English instead of returning a null value.

frett: k, the stream idea I had won't work either since there isn't a direct way to get a raw list of languages out of LocaleListCompat. so, this should be good as is

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
```diff
@@ -242,16 +252,58 @@ private void setupNavigationDrawer() {
...
+     */
+    private boolean containsLoginLanguage() {
+        LocaleListCompat localeListCompat = ConfigurationCompat
+                .getLocales(Resources.getSystem().getConfiguration());
+        String matchLanguage = localeListCompat.getFirstMatch(mLoginLanguages).getLanguage();
+        for (String lang : mLoginLanguages) {
+            if (matchLanguage.equals(lang)) {
+                return true;
+            }
         }
```
gyasistory: @frett I changed the loop so it will only iterate through one collection instead of two.  Let me know if you are OK with this before I merge to development (didn't realize you already approved).

PR #454, opened 2018-08-31
Aem download manger

gyasistory:
This is all of the functionality for pulling Articles from a Manifest and saving it to the database.


gyasistory:
Found a possible issue with Android 15.  The Uri conversion doesn't seem to work.

app/build.gradle
```diff
@@ -171,11 +171,15 @@ dependencies {
 
 // make crashlytics upload task dependent on the assemble task for all build variants
 afterEvaluate {
+
+//    crashlyticsUploadDistributionProductionDebug.dependsOn assembleProductionDebug
+
```
frett: unnecessary change, can be removed

app/build.gradle
```diff
@@ -171,11 +171,15 @@ dependencies {
...
     android.applicationVariants.all { variant ->
         def assembleTask = project.tasks.getByName("assemble${variant.name.capitalize()}")
         def crashlyticsTask = project.tasks.getByName("crashlyticsUploadDistribution${variant.name.capitalize()}")
         crashlyticsTask.dependsOn assembleTask
     }
+
```
frett: unnecessary change

ui/articles-aem-renderer/src/androidTest/java/org/cru/godtools/articles/aem/db/ArticleRoomDatabaseTest.java
```diff
@@ -95,20 +100,70 @@ public void verifyGetAllArticles() {
...
     @Test
     public void verifyManifestHasArticles() {
         assertTrue("Manifest 0 should have articles", mAssociationDao.getTestableArticlesByManifestID("0").size() > 0);
         assertTrue("Manifest 1 should have articles", mAssociationDao.getTestableArticlesByManifestID("1").size() > 0);
         assertTrue("Manifest 2 should have articles", mAssociationDao.getTestableArticlesByManifestID("2").size() > 0);
     }
+
+
+    @Test
+    public void verifyArticleParseLogic() {
```
frett: This should be in a separate `ArticleParserTest` class.

ui/articles-aem-renderer/src/androidTest/java/org/cru/godtools/articles/aem/db/ArticleRoomDatabaseTest.java
```diff
@@ -95,20 +100,70 @@ public void verifyGetAllArticles() {
...
+
+        try {
+            Uri url = Uri.parse("https://stage.cru.org/content/experience-fragments/questions_about_god/english.999.json");
```
frett: by depending on a web resource this test is very brittle. You should embed a raw json file and load that to perform your parsing tests. That way if the web content ever changes it doesn't break your test

ui/articles-aem-renderer/src/androidTest/java/org/cru/godtools/articles/aem/db/ArticleRoomDatabaseTest.java
```diff
@@ -95,20 +100,70 @@ public void verifyGetAllArticles() {
...
+        }
+
+    }
+
+    @Test
+    public void verifyAttachmentsAreSaved() {
+        for (Article article : mSavedArticles) {
+            for (Attachment attachment : mAttachmentDao.getTestableAttachmentsByArticle(article.mkey)) {
+                try {
+                    AEMDownloadManger
```
frett: you are testing AEMDownloadManager functionality, so this should be in an `AEMDownloadManagerTest` test class instead of in the `ArticleRoomDatabaseTest` class

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -0,0 +1,241 @@
...
+
+    public static final String ATTACHMENT_LIST_KEY = "attachment_list_key";
+
+    private JSONObject articleJSON;
+
+    private List<Attachment> attachmentList = new ArrayList<>();
+
+    private List<Article> articleList = new ArrayList<>();
+
+    private String createdTag = "jcr:created";
```
frett: these should be `private static final` to make them constants.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -0,0 +1,241 @@
...
+    private String contentTag = "jcr:content";
+    private String lastModifiedTag = "cq:lastModified";
+    private String uuidTag = "jcr:uuid";
+    private String titleTag = "jcr:title";
+    private String rootTag = "root";
+    private String fileTag = "fileReference";
+    private String baseUrl = "https://stage.cru.org";
```
frett: this shouldn't be here, where are you using it?

frett: you should be relying on the AEM import url for whatever functionality this is providing

library/db/build.gradle
```diff
@@ -8,4 +8,5 @@ dependencies {
     api "org.ccci.gto.android:gto-support-db-stream:${deps.gtoSupport}"
 
     implementation "com.crashlytics.sdk.android:crashlytics:${deps.crashlytics}"
+
```
frett: extraneous whitespace, shouldn't have been committed

ui/articles-aem-renderer/build.gradle
```diff
@@ -5,14 +5,24 @@ android {
...
     annotationProcessor "android.arch.lifecycle:compiler:${deps.androidLifecycle}"
     annotationProcessor "android.arch.persistence.room:compiler:${deps.roomVersion}"
+
+    // for convert Uri to Json
+    implementation "org.apache.directory.studio:org.apache.commons.io:${deps.ioCommons}"
+
+    implementation "com.jakewharton.timber:timber:${deps.timber}"
+
 }
+
```
frett: extraneous whitespace

ui/articles-aem-renderer/build.gradle
```diff
@@ -5,14 +5,24 @@ android {
...
+    implementation project(':library:xml-model')
 
-    implementation "android.arch.lifecycle:extensions:${deps.androidLifecycle}"
-    implementation "android.arch.persistence.room:runtime:${deps.roomVersion}"
-    implementation "com.android.support:appcompat-v7:${deps.androidSupport}"
-
+    // For Testing
     androidTestImplementation "com.android.support.test:runner:${deps.androidTestingSupport}"
     androidTestImplementation "android.arch.persistence.room:testing:${deps.roomVersion}"
+    testImplementation "android.arch.core:core-testing:${deps.androidLifecycle}"
```
frett: this dependency does nothing since it is targeting JVM unit tests and not the android tests.

ui/articles-aem-renderer/src/androidTest/java/org/cru/godtools/articles/aem/db/ArticleRoomDatabaseTest.java
```diff
@@ -95,20 +100,70 @@ public void verifyGetAllArticles() {
...
+            for (Attachment attachment : mAttachmentDao.getTestableAttachmentsByArticle(article.mkey)) {
+                assertFalse(attachment.mAttachmentFilePath != null);
+            }
+        }
+    }
 }
+
+
+
+
```
frett: extraneous trailing whitespace

frett:
we should use OkHttp3 for requests, that should handle the SSL protocol correctly

frett:
plus, we are already utilizing OkHttp3 for all other web requests (via Retrofit)

gyasistory:
OK I will try using that.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
gyasistory: I am just using them to parse the json data.  I wanted to make them member variable so I can easily change them if needed.

app/build.gradle
```diff
@@ -171,6 +171,9 @@ dependencies {
 
 // make crashlytics upload task dependent on the assemble task for all build variants
 afterEvaluate {
+
+    crashlyticsUploadDistributionProductionDebug.dependsOn assembleProductionDebug
+
```
frett: this code should be removed, not uncommented. it was replaced by the `android.applicationVariants.all` code right below it

ui/articles-aem-renderer/src/androidTest/java/org/cru/godtools/articles/aem/ArticleParseTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+            OkHttpClient client = new OkHttpClient();
+
+            Request request = new Request.Builder()
+                    .url(url.toString())
+                    .build();
+
+            Response response = client.newCall(request).execute();
+
+            String result = response.body().string();
```
frett: for tests you shouldn't be relying on fetching content from the web, it makes a test very brittle. you should include the json as a file and parse that. This makes the test self-contained and not dependent on an external system. see:
https://github.com/CruGlobal/godtools-android/blob/develop/library/xml-model/src/androidTest/java/org/cru/godtools/xml/model/ManifestIT.java#L18-L23
https://github.com/CruGlobal/godtools-android/blob/develop/library/xml-model/src/androidTest/assets/tests/manifest-empty.xml

ui/articles-aem-renderer/src/androidTest/java/org/cru/godtools/articles/aem/ArticleParseTest.java
```diff
@@ -0,0 +1,53 @@
...
+            HashMap<String, Object> results = new ArticleParser(jsonObject).execute();
+
+            assertTrue(results.containsKey(ArticleParser.ATTACHMENT_LIST_KEY));
+            assertTrue(results.containsKey(ArticleParser.ARTICLE_LIST_KEY));
+
+            List<Article> resultsArticles = (List<Article>) results.get(ArticleParser.ARTICLE_LIST_KEY);
+            assertTrue(resultsArticles.size() > 0);
+            List<Attachment> resultAttachments = (List<Attachment>) results.get(ArticleParser.ATTACHMENT_LIST_KEY);
+            assertTrue(resultAttachments.size() > 0);
+        } catch (JSONException | IOException | IllegalStateException e) {
```
frett: what you can do instead of catching the exceptions, is just declare the test with `throws Exception`, then you won't have to catch it only to propagate it as a `fail`. within tests you shouldn't worry about graceful error handling at all, tests should throw exception when something unexpected happens.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -0,0 +1,233 @@
...
+    /**
+     * This execute function allows you to change the initial jsonObject and run the execution.
+     *
+     * use <code>ARTICLE_LIST_KEY</code> and <code>ATTACHMENT_LIST_KEY</code> to access the Collections
+     * out of the HashMaps
+     *
+     * @param jsonObject = the object to be parsed.
+     * @return = return a list of <code>Article</code> and <code>Attachments</code> in HashMap
+     */
+    public HashMap<String, Object> execute(JSONObject jsonObject) {
```
frett: I don't think you ever use this version of `execute`

frett: from an overall class design perspective, this class has potential for data leakage if the same Parser object is used multiple times due to how the instance variables are used. There can be the intention that it will never be used in that way, but a developer a couple years down the road may not understand that constraint and mis-use the class.

Using instance variables is fine for tracking state, but I would make it so that `articleJSON` is final and only set in the constructor, and ensure you only parse the json one time when creating the object.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -0,0 +1,233 @@
...
+     * @param evaluatingObject = the Json Object to be evaluated.
+     */
+    private void jsonObjectHandler(JSONObject evaluatingObject) {
+        // Create loop through Keys
+        Iterator<String> keys = evaluatingObject.keys();
+        while (keys.hasNext()) {
+            String nextKey = keys.next();
+            if (isArticleOrCategory(nextKey)) {
+                try {
+                    JSONObject returnedObject = evaluatingObject.getJSONObject(nextKey);
```
frett: I probably would have used this structure instead:
```java
JSONObject returnedObject = evaluatingObject.optJSONObject(nextKey);
if(returnedObject != null) {
  ...
}
```

frett: then you wouldn't have needed to use a `catch`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -0,0 +1,233 @@
...
+                    if (isObjectCategory(returnedObject)) {
+                        jsonObjectHandler(returnedObject);
+                    } else {
+                        // Get inner Article Object
+                        parseArticleObject(returnedObject);
+                    }
+                } catch (Exception e) {
```
frett: catching all `Exceptions` like this can hide bugs

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -0,0 +1,233 @@
...
+        // Create loop through Keys
+        Iterator<String> keys = evaluatingObject.keys();
+        while (keys.hasNext()) {
+            String nextKey = keys.next();
+            if (isArticleOrCategory(nextKey)) {
+                try {
+                    JSONObject returnedObject = evaluatingObject.getJSONObject(nextKey);
+                    if (isObjectCategory(returnedObject)) {
+                        jsonObjectHandler(returnedObject);
+                    } else {
```
frett: I don't know for sure, but I would assume there might be other types of "objects" that may be present in the json. it's probably safer to use something like:
```java
switch(returnedObject.optString("jcr:primaryType", "")) {
  case "sling:OrderedFolder":
    jsonObjectHandler(returnedObject);
    break;
  case "{{page_type}}":
    parseArticleObject(returnedObject);
    break;
}
```

frett: also, `jcr:primaryType`, `sling:OrderedFolder`, and whatever the page type is should be static private constants defined at the top of this class

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -0,0 +1,233 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * This method parses an Article Json Object into the Database. On completion it will add
+     * articles to <code>articleList</code>
+     *
+     * @param articleObject = article JsonObject
```
frett: the actual javadoc format for parameters doesn't use an `=`, it should just be:
```
* @param articleObject article JsonObject
```
example: https://www.oracle.com/technetwork/articles/java/index-137868.html#format

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -0,0 +1,233 @@
...
+                    .getJSONObject(CONTENT_TAG).getString(LAST_MODIFIED_TAG));
+        }
+        retrievedArticle.mkey = articleObject.getJSONObject(CONTENT_TAG).getString(UUID_TAG);
+        retrievedArticle.mTitle = articleTagObject.getJSONObject(CONTENT_TAG)
+                .getString(TITLE_TAG);
+
+        JSONObject articleRootObject = articleTagObject.getJSONObject(CONTENT_TAG)
+                .getJSONObject(ROOT_TAG);
+
+        retrievedArticle.mContent = articleRootObject.getJSONObject("text").getString("text");
```
frett: this line can be removed, the actual content will be downloaded elsewhere

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -0,0 +1,233 @@
...
+     *
+     * @param articleRootObject = the root json Object of Article
+     * @param articleKey = the uuid of the article
+     */
+    private void getAttachmentsFromRootObject(JSONObject articleRootObject, String articleKey) {
+        // Iterate through keys
+        Iterator<String> keys = articleRootObject.keys();
+        while (keys.hasNext()) {
+            String nextKey = keys.next();
+            if (nextKey.contains("image")) {
```
frett: you shouldn't be looking for images based on the key name, you should look based on this attribute:
```
"sling:resourceType": "wcm/foundation/components/image"
```

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
gyasistory: Would it make more since to make execute a static method that takes in a local JsonObject, and use the local object to parse?

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
```diff
@@ -0,0 +1,159 @@
...
+            String[] urlSplit = attachment.mAttachmentUrl.split("/");
+            String filename = urlSplit[urlSplit.length - 1];
+            File articleFile = new File(context.getFilesDir(), "articles");
+            if (!articleFile.exists()) {
+                articleFile.mkdir();
+            }
+            articleFile = new File(articleFile, filename);
+            FileOutputStream outputStream = new FileOutputStream(articleFile);
+            URL url = new URL(attachment.mAttachmentUrl);
+            outputStream.write(IOUtils.toByteArray(url));
```
frett: this should be using `OkHttp`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -0,0 +1,194 @@
...
+ */
+public class ArticleParser {
+
+    public static final String ARTICLE_LIST_KEY = "article_list_key";
+
+    public static final String ATTACHMENT_LIST_KEY = "attachment_list_key";
+
+    private static List<Attachment> attachmentList = new ArrayList<>();
+
+    private static List<Article> articleList = new ArrayList<>();
```
frett: these 2 lists are still going to cause problems. They aren't cleared when parsing a second article json object.

frett:
you will need to resolve merge conflicts with the develop branch

frett:
@gyasistory you can look at the individual commits I made to get an idea of my thought process.

PR #461, opened 2018-09-05
Category parsing

frett:


frett:
The category id will be the link into which articles to open. At some point we will have categories ids defined on the articles we get from AEM.

PR #462, opened 2018-09-06
Category UI

frett:


ui/article-renderer/src/main/java/org/cru/godtools/article/activity/CategoriesActivity.java
```diff
@@ -0,0 +1,63 @@
...
+import org.cru.godtools.article.fragment.CategoriesFragment;
+import org.cru.godtools.base.tool.activity.BaseSingleToolActivity;
+
+import java.util.Locale;
+
+public class CategoriesActivity extends BaseSingleToolActivity {
+    private static final String TAG_MAIN_FRAGMENT = "mainFragment";
+
+    public static void start(@NonNull final Context context, @NonNull final String toolCode,
+                             @NonNull final Locale language) {
```
gyasistory: I never thought of creating a static method to start an activity it makes so much since.

frett: it's a really nice convenience method :) it ensures you are passing the correct data in the correct locations.

PR #463, opened 2018-09-07
Updated the Adobe Analytics call to include missing requirements.

gyasistory:
This PR covers the missing Adobe Analytics tags

### DIFF
--- a/library/analytics/src/main/java/org/cru/godtools/analytics/AdobeAnalyticsService.java
+++ b/library/analytics/src/main/java/org/cru/godtools/analytics/AdobeAnalyticsService.java
@@ -89,7 +89,7 @@ public final class AdobeAnalyticsService implements AnalyticsService {
         return sInstance;
     }
 
-    /* BEGIN tracking methods */
+    // region Tracking Methods
 
     @UiThread
     @Override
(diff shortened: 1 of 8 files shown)


library/analytics/src/main/java/org/cru/godtools/analytics/AdobeAnalyticsService.java
```diff
@@ -202,6 +211,55 @@ private void trackState(@NonNull final String screen, @Nullable final Locale con
         if (contentLocale != null) {
             data.put(KEY_CONTENT_LANGUAGE, toLanguageTag(contentLocale));
         }
+
+        switch (screen) {
+            case "Home":
```
frett: use the screen constants for this switch, that way if they are ever updated the changes propagate here.

library/analytics/src/main/java/org/cru/godtools/analytics/AdobeAnalyticsService.java
```diff
@@ -202,6 +211,55 @@ private void trackState(@NonNull final String screen, @Nullable final Locale con
...
+                //No site Section for this screen
+
+                break;
+            case "Find Tools":
+                data.put(KEY_SITE_SECTION, "tools");
```
frett: `"tools"` should be a constant value

frett:
I still personally feel including the bulk of this logic in the `ScreenAnalyticsEvent` and the child event provides a more flexible implementation that is easier to maintain/update long term

frett:
@gyasistory updated this to move the site section logic to the screen event model. I also created an `EveryStudentAnalyticsScreenEvent` to centralize the legacy everystudent functionality

PR #466, opened 2018-09-10
[GT-509] Make it possible to toggle tabs by clicking images

frett:
This is dependent on https://github.com/CruGlobal/mobile-content-api/pull/163

There is a test page that illustrates this use-case, it is the "Tabs Select Listener testing" page in the "Renderer Test" tract in the stage environment


frett:
I'm waiting on merging this until the xml spec is approved and merged. See: https://github.com/CruGlobal/mobile-content-api/pull/163

PR #467, opened 2018-09-10
GT 512 Duplicate action bars

gyasistory:
https://jira.cru.org/browse/GT-512

### DIFF
--- a/ui/base-tool/src/main/res/values/styles.xml
+++ b/ui/base-tool/src/main/res/values/styles.xml
@@ -5,7 +5,7 @@
         <item name="toolbarStyle">@style/Widget.GodTools.Tool.Toolbar.AppBar</item>
     </style>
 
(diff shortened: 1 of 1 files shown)


ui/base-tool/src/main/res/values/styles.xml
```diff
@@ -5,9 +5,10 @@
         <item name="toolbarStyle">@style/Widget.GodTools.Tool.Toolbar.AppBar</item>
     </style>
 
-    <style name="Theme.GodTools.Tool.AppBar">
+    <style name="Theme.GodTools.Tool.AppBar" parent="Theme.AppCompat.NoActionBar">
```
frett: don't include the parent attribute, it actually breaks how we are inheriting style data

frett: without the parent attribute it automatically inherits from "Theme.GodTools.Tool", which is the expected behavior.

PR #469, opened 2018-09-11
AEM download manager wireframe

frett:
- make the download manager a singleton
- create wireframe methods for the various tasks


frett:
@gyasistory this should be ready for review

frett:
yup, couldn't get too much further until room db updates are applied. But doing the wireframe will allow both of us to work on this if necessary.

PR #470, opened 2018-09-11
Room upgrade & enable stetho

frett:


ui/articles-aem-renderer/src/androidTest/java/org/cru/godtools/articles/aem/db/AEMDownloadMangerTest.java
```diff
@@ -12,6 +13,7 @@
 public class AEMDownloadMangerTest extends DBBaseTest {
 
     @Test
+    @Ignore
```
gyasistory: Is there a reason you are ignoring this test.

frett: It was causing Travis to fail due to the test making a network request

PR #471, opened 2018-09-12
Article list view

gyasistory:
Create List View for Articles Section

### DIFF
--- a/ui/article-renderer/build.gradle
+++ b/ui/article-renderer/build.gradle
@@ -22,6 +22,4 @@ dependencies {
     implementation "com.jakewharton:butterknife:${deps.butterknife}"
 
     annotationProcessor "com.jakewharton:butterknife-compiler:${deps.butterknife}"
-
-    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
 }
--- a/ui/article-renderer/build.gradle
+++ b/ui/article-renderer/build.gradle
@@ -23,4 +23,5 @@ dependencies {
 
     annotationProcessor "com.jakewharton:butterknife-compiler:${deps.butterknife}"
 
+    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
 }
--- a/ui/article-renderer/build.gradle
+++ b/ui/article-renderer/build.gradle
@@ -11,6 +11,8 @@ dependencies {
 
     implementation project(':ui:base-tool')
 
+    implementation project(':ui:articles-aem-renderer')
+
     implementation "androidx.constraintlayout:constraintlayout:${deps.constraintLayout}"
     implementation "com.android.support:recyclerview-v7:${deps.androidSupport}"
 
@@ -20,4 +22,5 @@ dependencies {
     implementation "com.jakewharton:butterknife:${deps.butterknife}"
 
     annotationProcessor "com.jakewharton:butterknife-compiler:${deps.butterknife}"
+
 }
--- a/ui/article-renderer/build.gradle
+++ b/ui/article-renderer/build.gradle
@@ -9,10 +9,10 @@ android {
 dependencies {
     implementation project(':library:base')
 
+    implementation project(':ui:articles-aem-renderer')
     implementation project(':ui:base-tool')
 
-    implementation project(':ui:articles-aem-renderer')
-
+    implementation "android.arch.lifecycle:extensions:${deps.androidLifecycle}"
     implementation "androidx.constraintlayout:constraintlayout:${deps.constraintLayout}"
     implementation "com.android.support:recyclerview-v7:${deps.androidSupport}"
 
--- a/ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
+++ b/ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
@@ -1,0 +1,61 @@
+package org.cru.godtools.article.activity;
+
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.v4.app.FragmentManager;
+
+import org.cru.godtools.article.fragment.ArticlesFragment;
+import org.cru.godtools.base.tool.activity.BaseSingleToolActivity;
+import org.cru.godtools.article.R;
+import java.util.Locale;
+
+public class ArticlesActivity extends BaseSingleToolActivity{
+
+    private static final String CATEGORY_KEY = "category_key";
+    private String mCategoryID;
+
+    //region Initialization
+    public  static  void  start(@NonNull final Context context, @NonNull final  String toolCode,
+                                @NonNull final Locale language, @NonNull final String categoryID){
+        Bundle args = new Bundle();
+        populateExtras(args, toolCode, language);
+        args.putString(CATEGORY_KEY, categoryID);
+        Intent intent = new Intent(context, ArticlesActivity.class).putExtras(args);
+        context.startActivity(intent);
+    }
+
+    public ArticlesActivity(boolean immersive) {
+        super(immersive);
+    }
+    //endregion
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_article_articles);
+        if (getIntent().hasExtra(CATEGORY_KEY)){
+            mCategoryID = getIntent().getStringExtra(CATEGORY_KEY);
+        }
+    }
+
+
+    @Override
+    protected void onStart() {
+        super.onStart();
+        loadInitialFragmentIfNeeded();
+    }
+
+    private void loadInitialFragmentIfNeeded() {
+        FragmentManager fm = getSupportFragmentManager();
+
+        if (fm.findFragmentByTag(ArticlesFragment.TAG) == null){
+           fm.beginTransaction().replace(R.id.articles_frame, ArticlesFragment
+                   .newInstance(mCategoryID != null ? mCategoryID : "")).commit();
+        }
+
+
+    }
+}
--- a/ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
+++ b/ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
@@ -28,17 +28,17 @@ public class ArticlesActivity extends BaseSingleToolActivity {
     }
 
     public ArticlesActivity() {
-        super(true);
+        super(false);
     }
     //endregion
 
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        setContentView(R.layout.activity_article_articles);
         if (getIntent().hasExtra(CATEGORY_KEY)) {
             mCategoryID = getIntent().getStringExtra(CATEGORY_KEY);
         }
+        setContentView(R.layout.activity_article_articles);
     }
 
     @Override
--- a/ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
+++ b/ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
@@ -27,8 +27,10 @@ public class ArticlesActivity extends BaseSingleToolActivity{
         context.startActivity(intent);
     }
 
-    public ArticlesActivity(boolean immersive) {
-        super(immersive);
+
+
+    public ArticlesActivity() {
+        super(true);
     }
     //endregion
 
@@ -53,7 +55,7 @@ public class ArticlesActivity extends BaseSingleToolActivity{
 
         if (fm.findFragmentByTag(ArticlesFragment.TAG) == null){
            fm.beginTransaction().replace(R.id.articles_frame, ArticlesFragment
-                   .newInstance(mCategoryID != null ? mCategoryID : "")).commit();
+                   .newInstance(mTool, mLocale, mCategoryID != null ? mCategoryID : "")).commit();
         }
 
 
--- a/ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
+++ b/ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
@@ -12,22 +12,20 @@ import org.cru.godtools.base.tool.activity.BaseSingleToolActivity;
 import org.cru.godtools.article.R;
 import java.util.Locale;
 
-public class ArticlesActivity extends BaseSingleToolActivity{
+public class ArticlesActivity extends BaseSingleToolActivity {
 
     private static final String CATEGORY_KEY = "category_key";
     private String mCategoryID;
 
     //region Initialization
-    public  static  void  start(@NonNull final Context context, @NonNull final  String toolCode,
-                                @NonNull final Locale language, @NonNull final String categoryID){
+    public static void start(@NonNull final Context context, @NonNull final String toolCode,
+                                @NonNull final Locale language, @NonNull final String categoryID) {
         Bundle args = new Bundle();
         populateExtras(args, toolCode, language);
         args.putString(CATEGORY_KEY, categoryID);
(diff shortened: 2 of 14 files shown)


ui/article-renderer/build.gradle
```diff
@@ -20,4 +22,6 @@ dependencies {
     implementation "com.jakewharton:butterknife:${deps.butterknife}"
 
     annotationProcessor "com.jakewharton:butterknife-compiler:${deps.butterknife}"
+
+    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
```
frett: we are already including constraint-layout in this module. We are using the version in the AndroidX namespace per https://developer.android.com/topic/libraries/support-library/androidx-overview

ui/article-renderer/src/main/res/layout/fragment_articles.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<android.support.constraint.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
frett: there is no point in having this ConstraintLayout here

ui/article-renderer/src/main/res/layout/list_item_article.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
frett: `RelativeLayout` would give better performance for this layout since you aren't using any automatic size distribution that `LinearLayout` is optimized for.

ui/article-renderer/src/main/res/layout/list_item_article.xml
```diff
@@ -0,0 +1,26 @@
...
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:orientation="vertical"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
```
frett: should have `wrap_content` for the layout_height

ui/article-renderer/src/main/res/layout/list_item_article.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
```
frett: you should use databinding for this layout. see: https://github.com/CruGlobal/godtools-android/blob/develop/ui/article-renderer/src/main/res/layout/list_item_category.xml

ui/article-renderer/src/main/java/org/cru/godtools/article/adapter/ArticleAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+
+import org.cru.godtools.article.R;
+import org.cru.godtools.articles.aem.model.Article;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.List;
+import java.util.Locale;
+
+public class ArticleAdapter extends RecyclerView.Adapter<ArticleAdapter.ArticleViewHolder> {
```
frett: a lot of this logic disappears when you switch to utilizing databinding. see: https://github.com/CruGlobal/godtools-android/blob/develop/ui/article-renderer/src/main/java/org/cru/godtools/article/adapter/CategoriesAdapter.java

ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java
```diff
@@ -0,0 +1,68 @@
...
+        super.onCreateView(inflater, container, savedInstanceState);
+
+        return inflater.inflate(R.layout.fragment_articles, container, false);
+    }
+
+    @Override
+    public void onActivityCreated(@Nullable Bundle savedInstanceState) {
+        super.onActivityCreated(savedInstanceState);
+
+        RecyclerView recyclerView = Objects.requireNonNull(getActivity())
```
frett: try to limit the amount of random code initialization that happens within the various lifecycle methods. you should organize the logic into helper methods like `setupRecyclerView()`. This helps keep fragments and activities maintainable long term by reducing complexity of the lifecycle event methods.

frett: Also, you should utilize Butterknife to bind views you need within fragments and activities. Extend `BaseFragment` and see https://github.com/CruGlobal/godtools-android/blob/develop/ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/CategoriesFragment.java#L27-L29 for how to simply get the `RecyclerView`

ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java
```diff
@@ -0,0 +1,72 @@
...
+        mArticlesView.setLayoutManager(new LinearLayoutManager(getActivity()));
+
+        ArticleViewModel viewModel = ArticleViewModel.getInstance(getActivity());
+
+        String manifestKey = getArguments() != null ? getArguments().getString(MANIFEST_Key) : "";
+
+        viewModel.getArticlesByManifest(manifestKey).observe(this, articles -> {
+            // This will be triggered by any change to the database
+            mAdapter.setArticles(articles);
+        });
```
frett: you should move the RecyclerView initialization logic to a private `setupArticlesView()` method, it helps reduce complexity of the the various lifecycle methods.

It also makes it much more clear what some logic is supposed to be doing. looking at a list of methods, `onActivityCreated()` doesn't indicate that it's initializing the list of articles for this fragment. But `setupArticlesView()` would be very indicative that this is where the logic to setup the articles view lives

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/view_model/ArticleViewModel.java
```diff
@@ -18,6 +20,16 @@
...
     private final ManifestAssociationRepository mManifestRepository;
 
+    /**
+     * This allows other models to get an instance of this ViewModel with out adding dependencies
+     * to their .gradle file
+     * @param activity FragmentActivity
+     * @return this
+     */
+    public static ArticleViewModel getInstance(FragmentActivity activity){
+        return ViewModelProviders.of(activity).get(ArticleViewModel.class);
```
frett: because you are calling this from a fragment you should use the Fragment itself and not the activity it was attached to.

ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java
```diff
@@ -0,0 +1,72 @@
...
+        return inflater.inflate(R.layout.fragment_articles, container, false);
+    }
+
+    @Override
+    public void onActivityCreated(@Nullable Bundle savedInstanceState) {
+        super.onActivityCreated(savedInstanceState);
+
+        mAdapter = new ArticleAdapter();
+        Objects.requireNonNull(mArticlesView).setAdapter(mAdapter);
+        mArticlesView.setLayoutManager(new LinearLayoutManager(getActivity()));
```
frett: use `requireContext()` instead of `getActivity()`

frett:
as a side note, I'm not fully sure what the current `ArticleViewModel` gains us, or if we are using it correctly.

it appears that `ViewModel` is supposed to be maintained across activity/fragment configuration changes. Within the model you should probably be caching the data that you want to reuse. How you are currently using it is not caching the `LiveData` object for re-use. you are creating a new `LiveData` object every time the fragment is initialized.

frett:
Also, it appears that the ContentLoader framework that we are using elsewhere within GodTools uses a `ViewModel` at it's core to persist loaders across configuration changes.

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public ArticlesActivity() {
+        super(true);
+    }
+    //endregion
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_article_articles);
```
frett: I generally set the content view after I read any intent extras and restore any activity saved state. That way you can utilize the activity config and state when the view is being setup.

frett: see: https://github.com/CruGlobal/godtools-android/blob/develop/ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java#L226

ui/base/src/main/res/values/colors.xml
```diff
@@ -4,10 +4,10 @@
     <color name="tintable">#FF000000</color>
     <color name="tintable_20">#33000000</color>
     <color name="black">#000000</color>
-    <color name="white">#FFFFFF</color>
+    <color name="white">#ffffff</color>
```
gyasistory: This was auto corrected.

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
```diff
@@ -0,0 +1,58 @@
...
+                                @NonNull final Locale language, @NonNull final String categoryID) {
+        Bundle args = new Bundle();
+        populateExtras(args, toolCode, language);
+        args.putString(CATEGORY_KEY, categoryID);
+        Intent intent = new Intent(context, ArticlesActivity.class).putExtras(args);
+        context.startActivity(intent);
+    }
+
+    public ArticlesActivity() {
+        super(true);
```
frett: I was opting to not make any of the articles UI immersive, so make this `super(false)`

ui/article-renderer/src/main/java/org/cru/godtools/article/adapter/ArticleAdapter.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * This method is called to set the list of articles.  Used by Live Data to update list
+     *
+     * @param articles list of Articles
+     */
+    public void setArticles(List<Article> articles) {
+        mArticles = articles;
+        notifyDataSetChanged();
+    }
+}
```
frett: this functionally looks good, I would just re-organize the order of methods so that similar types of methods are together.

for example, put all the data setters together, keep all the lifecycle event methods (`on...`) together. put the `Callback` interface near the top of the class so it's easy to see for other developers when looking at the adapter.

when ordering the lifecycle event methods, try to order them in the sequence they will be called when possible, this allows other developers to quickly look through the lifecycle events to get an idea of the lifecycle of that component. In this component the sequence would be `create`, `bind`, `recycle`

ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private void setupArticleRecyclerView() {
+        if (mArticlesView != null) {
+            mAdapter = new ArticleAdapter();
+            mAdapter.setCallbacks(this);
+            mAdapter.setToolManifest(mManifest);
+            DividerItemDecoration itemDecoration = new DividerItemDecoration(getContext(),
+                    DividerItemDecoration.VERTICAL);
+            itemDecoration.setDrawable(Objects.requireNonNull(getActivity()).getResources()
+                    .getDrawable(R.drawable.divider));
```
frett: let's configure this drawable via a theme/style. I try to keep any sort of styling outside of java code if at all possible. Obviously things driven from the tool manifest xml cannot be handled via XML and has to be programmatically set (unless we figure out how to do it via databinding later), but this divider doesn't fall into that category.

ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java
```diff
@@ -0,0 +1,95 @@
...
+
+            mArticlesView.addItemDecoration(itemDecoration);
+            mArticlesView.setLayoutManager(new LinearLayoutManager(getActivity()));
+            mArticlesView.setAdapter(mAdapter);
+
+            ArticleViewModel viewModel = ArticleViewModel.getInstance(getActivity());
```
frett: I think I'm going to create a commit on your branch with what I'm thinking of for the fragment view model

ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java
```diff
@@ -0,0 +1,95 @@
...
+
+            String manifestKey = getArguments() != null ? getArguments().getString(MANIFEST_KEY) : "";
```
frett: as a personal preference I always process all arguments within `onCreate()` and assign them to instance variables. Then when initializing components I just use the instance variables.

ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java
```diff
@@ -0,0 +1,95 @@
...
+            viewModel.getArticlesByManifest(manifestKey).observe(this, articles -> {
+                // This will be triggered by any change to the database
+                mAdapter.setArticles(articles);
+            });
+        }
+    }
+
+    @Override
+    public void onArticleSelected(Article article) {
+        Timber.d("You selected %s as your article", article.mTitle);
```
frett: yay for temporary debugging!

did you know that you can add a `tag` to the message with code like this:
```
Timber.tag("Articles Fragment").d("something happened!!");
```

obviously this isn't necessary for this temporary debugging message, but it is beneficial for permanent logging messages so you can track down log messages to component quicker.

ui/article-renderer/src/main/res/layout/activity_article_articles.xml
```diff
@@ -0,0 +1,27 @@
...
+        android:background="@android:color/transparent"
+        app:elevation="0dp">
+
+        <android.support.v7.widget.Toolbar
+            android:id="@+id/appbar"
+            android:layout_width="match_parent"
+            android:layout_height="?attr/actionBarSize"
+            app:elevation="0dp" />
+    </android.support.design.widget.AppBarLayout>
+</android.support.design.widget.CoordinatorLayout>
```
frett: hmm, I think I need to extract a simple tool activity layout that can be re-used by our multiple activities.

ui/article-renderer/src/main/res/layout/list_item_article.xml
```diff
@@ -0,0 +1,46 @@
...
+            android:id="@+id/article_title_text_view"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            android:text="@{article.mTitle}"
+            android:layout_margin="0dp"
+            android:padding="0dp"
+            android:textAppearance="@style/TextAppearance.GodTools.Large"
+            android:textColor="@{manifest.getPrimaryColor()}"
```
frett: oh yeah! I forgot databinding would do automagically wire simple properties!

ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java
frett: I haven't, let me take a look at this and see if I can figure out how to accomplish it. I just know the theory of how it's supposed to work ;)

gyasistory: @frett I was able to move the Layout Manager to the XML Layout, but everything I am seeing about setting the divider is doing it in code.  Everything I see for doing in XML is for other types of views like ListView.  I even tried using the divider tag with no success.  I am going to leave it in code and keep doing some research.

ui/article-renderer/src/main/res/layout/activity_article_articles.xml
frett: I did this in develop, we can change this after this PR is merged.

PR #472, opened 2018-09-13
Translation room db updates

frett:


frett:
@gyasistory this is ready for review

PR #479, opened 2018-09-18
Categories activity callbacks

frett:


ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/CategoriesFragment.java
```diff
@@ -62,7 +65,10 @@ protected void onManifestUpdated() {
 
     @Override
     public void onCategorySelected(@Nullable final Category category) {
-        Timber.d(category != null ? category.getId() : null);
+        final Callbacks callbacks = FragmentUtils.getListener(this, Callbacks.class);
+        if (callbacks != null) {
```
gyasistory: This is my first time looking at your FragmentUtils it is very helpful.

frett: I think I originally wrote the `getListener` method 4+ years ago

PR #483, opened 2018-09-20
Sync aem import task

gyasistory:
The PR is for attaching the logic for parsing an AEM api into the import Task.

### DIFF
--- a/ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
+++ b/ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
@@ -196,8 +196,8 @@ public class AEMDownloadManger {
     void syncAemImportTask(@NonNull final Uri baseUri, final boolean force) {
         if (!force) {
             AemImport aemImport = mAemDb.aemImportDao().getAemImport(baseUri);
-            if (aemImport != null){
-                if (!aemImport.isStale()){
+            if (aemImport != null) {
+                if (!aemImport.isStale()) {
                     return; // Don't import Aem if not forced and aemImport is not stale
                 }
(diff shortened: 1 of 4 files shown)


ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
```diff
@@ -193,7 +194,11 @@ void syncStaleAemImportsTask() {
      */
     @WorkerThread
     void syncAemImportTask(@NonNull final Uri baseUri, final boolean force) {
-        // TODO
```
gyasistory: I was not sure how to handle the force param.  It seems to be handle in the calling Task method.

frett: This method may be called on an AEM import we already synced once, force indicates if we should sync no matter what. if `force` is false, then we want to observe `AemImport.isStale()` to determine if we should actually sync.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
```diff
@@ -75,12 +75,13 @@
     private final Map<Uri, SyncAemImportTask> mSyncAemImportTasks = Collections.synchronizedMap(new HashMap<>());
 
     private AEMDownloadManger(@NonNull final Context context) {
-        mAemDb = ArticleRoomDatabase.getInstance(context);
-        mDao = GodToolsDao.getInstance(context);
+        mContext = context;
```
frett: `mContext = context.getApplicationContext()` to be safe

PR #484, opened 2018-09-20
Use url for article key

gyasistory:
This is to convert from using the Uuid for the Article Key to the Article url.

### DIFF
--- a/ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
+++ b/ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
@@ -6,6 +6,8 @@ import org.cru.godtools.articles.aem.model.Article;
 import org.cru.godtools.articles.aem.model.Attachment;
 import org.json.JSONObject;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
@@ -13,6 +15,8 @@ import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
+
+import timber.log.Timber;
 
 /**
  * This class handles parsing any AEM json calls into DOA objects.
@@ -31,16 +35,15 @@ public class ArticleParser {
     private static final String ORDER_FOLDER_TAG = "sling:OrderedFolder";
     private static final String PAGE_TAG = "cq:Page";
 
-    private static final String ARTICLE_BASE_URL = BASE_URL + "/content/experience-fragments/questions_about_god";
-
     //region Public Executor
     /**
      * This executes the parsing of the local JsonObject.
      *
      * @return return a list of {@link Article}
      */
-    public static List<Article> parse(@NonNull final JSONObject articleJSON) {
-        return jsonObjectHandler(articleJSON, new ArrayList<>());
+    public static List<Article> parse(@NonNull final JSONObject articleJSON, String url) {
+        String baseURL = getArticleBaseUrl(url);
+        return jsonObjectHandler(articleJSON, new ArrayList<>(), baseURL);
     }
     //endregion public Executor
 
@@ -52,10 +55,12 @@ public class ArticleParser {
      *
      * @param evaluatingObject the Json Object to be evaluated.
      * @param urlParams
+     * @param baseURL
      * @return All the articles that were parsed
      */
     @NonNull
-    private static List<Article> jsonObjectHandler(JSONObject evaluatingObject, List<String> urlParams) {
+    private static List<Article> jsonObjectHandler(JSONObject evaluatingObject,
+                                                   List<String> urlParams, String baseURL) {
         final List<Article> articles = new ArrayList<>();
 
         // Create loop through Keys
@@ -68,12 +73,12 @@ public class ArticleParser {
                     urlParams.add(nextKey);
                     switch (returnedObject.optString(PRIMARY_TYPE_TAG)) {
                         case ORDER_FOLDER_TAG:
-                            jsonObjectHandler(returnedObject, urlParams);
+                            articles.addAll(jsonObjectHandler(returnedObject, urlParams, baseURL));
                             // This param has been used and needs to be removed before looping.
 
                             break;
                         case PAGE_TAG:
-                            articles.add(parseArticleObject(returnedObject, urlParams));
+                            articles.add(parseArticleObject(returnedObject, urlParams, baseURL));
                             // This param has been used and needs to be removed before looping.
 
                             break;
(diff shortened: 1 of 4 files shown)


ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -31,14 +31,16 @@
     private static final String ORDER_FOLDER_TAG = "sling:OrderedFolder";
     private static final String PAGE_TAG = "cq:Page";
 
+    private static final String ARTICLE_BASE_URL = BASE_URL + "/content/experience-fragments/questions_about_god";
```
frett: we can't assume this is the base url, you should pass in whatever AEM import base URL you are actually parsing

frett: or base the url on that

gyasistory: @frett is the "/content/experience-fragments/questions_about_god" always going to be the same?

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -62,16 +65,20 @@
...
 
                             break;
                         case PAGE_TAG:
-                            articles.add(parseArticleObject(returnedObject));
+                            articles.add(parseArticleObject(returnedObject, urlParams));
+                            // This param has been used and needs to be removed before looping.
 
                             break;
                     }
+                    urlParams = new ArrayList<>();
```
frett: resetting the `urlParams` here will lose any parent paths that were traversed already.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -62,16 +65,20 @@
             if (isArticleOrCategory(nextKey)) {
                 JSONObject returnedObject = evaluatingObject.optJSONObject(nextKey);
                 if (returnedObject != null) {
+                    urlParams.add(nextKey);
```
frett: if you don't remove the `nextKey` urlParams you add here though you will have a corrupted url.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -31,14 +31,16 @@
...
+
     //region Public Executor
     /**
      * This executes the parsing of the local JsonObject.
      *
      * @return return a list of {@link Article}
      */
     public static List<Article> parse(@NonNull final JSONObject articleJSON) {
-        return jsonObjectHandler(articleJSON);
+        return jsonObjectHandler(articleJSON, new ArrayList<>());
```
frett: it's probably better to not use a `List` at all, but instead pass in a `Uri` object.

frett: The initial `Uri` object can be the `AemImport` base url

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
frett: using a `Uri` object you can just use this line of code: `final Uri newUrl = url.buildUpon().appendPath(nextKey).build();`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
frett: using a `Uri` object you won't need to reset it after processing anymore.

gyasistory: I had to reset the urlParams because I can have more than one Category or Article in the same Iteration of Keys.  So I am reseting the value before I loops to the next set of Keys.

frett: unfortunately the sequence of my comments wasn't preserved. I think using a `List` to keep track of the recursive parsing path is highly error prone, and how you have it right now will still produce errors.

gyasistory: Yeah I didn't like using the list either but wasn't sure how to handle it another way.  I will try working on using a URI object.  To see if I can get it working with that way.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -49,10 +56,12 @@
      * and Category Object will be Recursively placed back into this method.
      *
      * @param evaluatingObject the Json Object to be evaluated.
+     * @param urlParams
      * @return All the articles that were parsed
      */
     @NonNull
-    private static List<Article> jsonObjectHandler(JSONObject evaluatingObject) {
+    private static List<Article> jsonObjectHandler(JSONObject evaluatingObject,
+                                                   Uri urlParams) {
```
frett: I would name the parameter `baseUri` instead of `urlParams`. I feel baseUri is more descriptive of what the value actually is.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -172,6 +190,21 @@ private static long getDateLongFromJsonString(String dateString) {
...
+
+    /**
+     * Using "https://stage.cru.org/content/experience-fragments/questions_about_god/english.9999.json"
+     * as an example url this method will remove ".999.json" from the url leaving the base url.
+     * @param url
+     * @return
+     */
+    private static String getArticleBaseUrl(String url) {
+        if (url.endsWith(".json")) {
+            url = url.substring(0, url.lastIndexOf("."));
```
frett: I don't think this will actually be necessary since we should never be passing in a url that has `.###.json`.

But if we did need to sanitize the url, this logic is brittle. If the url is just `https://stage.cru.org/english.json`, This will  throw a `StringIndexOutOfBoundsException`. How you would probably want to handle this is through a regular expression.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/ArticleParser.java
```diff
@@ -136,7 +148,13 @@ private static Article parseArticleObject(JSONObject articleObject) {
     @NonNull
     private static List<Attachment> getAttachmentsFromRootObject(JSONObject articleRootObject, String articleKey) {
```
frett: instead of `String articleKey`, I would pass in the current `baseUri` that you have been building.

frett:
@gyasistory I'm merging this since it's been approved already

PR #486, opened 2018-09-21
Login feature tracking

frett:


frett:
that's correct. I need to do better at actually describing PRs. This was work I had done when we started working on the signup banner, but we stalled that work temporarily.

As you were approaching adding in some new features for card bounce, I wanted to get the tweaks I had made to feature tracking merged in.

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
```diff
@@ -215,11 +224,17 @@ public boolean onNavigationItemSelected(@NonNull final MenuItem item) {
...
-        stopLanguagesChangeListener();
+        stopSettingsChangeListener();
     }
 
     /* END lifecycle */
 
+    @MainThread
+    @Subscribe(threadMode = ThreadMode.MAIN)
+    public final void theKeyEvent(@NonNull final TheKeyEvent event) {
+        onTheKeyEvent(event);
```
frett: having an overridable lifecycle event is used to refresh the signup banner UI when the login state changes.

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.java
```diff
@@ -120,15 +121,23 @@ protected void onSetupActionBar() {
...
         mEventBus.register(this);
         loadLanguages(false);
         updateNavigationDrawerMenu();
     }
 
-    @MainThread
-    @Subscribe(threadMode = ThreadMode.MAIN)
-    public void onTheKeyEvent(@NonNull final TheKeyEvent event) {
+    @CallSuper
+    public void onSettingsUpdated(@Nullable final SharedPreferences preferences, @Nullable final String key) {
```
frett: I made this overridable so that the signup banner UI can be refreshed when the login feature preference changes.

PR #488, opened 2018-09-21
Articles fragment -> activity callbacks

frett:
This adds in the callback from the articles fragment to the articles activity.


ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
```diff
@@ -57,14 +63,23 @@ protected void onStart() {
...
-           fm.beginTransaction().replace(R.id.articles_frame, ArticlesFragment
-                   .newInstance(mTool, mLocale, mCategoryId)).commit();
+        if (fm.findFragmentByTag(TAG_MAIN_FRAGMENT) != null) {
+            return;
         }
+
+        assert mTool != null : "if mTool was null the activity would have already finished";
+        fm.beginTransaction()
+                .replace(R.id.frame, ArticlesFragment.newInstance(mTool, mLocale, mCategoryId))
+                .commit();
```
gyasistory: We didn't add the TAG in the replace call

frett: good catch!

PR #490, opened 2018-09-25
Android Studio 3.2 & build tools update

frett:
Upgrade gradle plugin to support Android Studio 3.2. I also had to fix the initial content generation script


frett:
@gyasistory merged

PR #492, opened 2018-09-25
GT-463 Android -on first use, make first card jump

gyasistory:
https://jira.cru.org/browse/GT-463

Setting it so the first card jumps if never opened.


ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -51,15 +56,18 @@
         void onActiveCardChanged(@Nullable View activeCard);
     }
 
-    private final GestureDetectorCompat mGestureDetector;
+    private Boolean mShouldAnimateCard = true;
```
frett: you should use the primitive `boolean` and not the boxed `Boolean`

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -69,9 +77,18 @@ public boolean onFling(final MotionEvent e1, final MotionEvent e2, final float v
     private View mActiveCard;
     private int mActiveCardPosition = 0;
 
+    private ObjectAnimator mBounceAnimator;
```
frett: it probably would make the logic much easier if you just set the bounce animator on the already existing `mAnimation` field

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -82,30 +99,38 @@ public void onAnimationEnd(final Animator animation) {
...
 
+    //region Initialization
     public PageContentLayout(@NonNull final Context context) {
         this(context, null);
+        initialization(context);
     }
 
     public PageContentLayout(@NonNull final Context context, @Nullable final AttributeSet attrs) {
         this(context, attrs, 0);
+        initialization(context);
```
frett: `initialization()` would have already been called at this point.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -82,30 +99,38 @@ public void onAnimationEnd(final Animator animation) {
         }
     };
 
+    //region Initialization
     public PageContentLayout(@NonNull final Context context) {
         this(context, null);
+        initialization(context);
```
frett: `initialization()` would have already been called twice at this point.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
gyasistory: How would I be able to stop it when a different animation runs.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
gyasistory: So these aren't overloaded Constructors?

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -82,30 +99,38 @@ public void onAnimationEnd(final Animator animation) {
...
     }
 
     public PageContentLayout(@NonNull final Context context, @Nullable final AttributeSet attrs,
                              final int defStyleAttr) {
         super(context, attrs, defStyleAttr);
+        initialization(context);
+    }
+
+    private void initialization(@NonNull Context context) {
```
frett: I think I would rather not have an `initialization()` method and be able to keep the singletons you are using as `final`.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -406,6 +431,69 @@ protected LayoutParams generateLayoutParams(final ViewGroup.LayoutParams p) {
...
+            mBounceAnimator.start();
+        }
+    }
+
+    /**
+     *  This method will check to see if the Animate Card feature has been detected and start or stop the
+     *  animation if needed.
+     */
+    public void shouldAnimateCard() {
+        mShouldAnimateCard = !mSettings.isFeatureDiscovered(Settings.FEATURE_ANIMATE_CARD);
```
frett: 👎 checking if you need to animate the card or not should be in the `PageViewHolder`

frett: and the `PageViewHolder` should be responsible for updating whether the `PageContentLayout` is animating the first card or not.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -736,4 +824,28 @@ public LayoutParams(@NonNull final LayoutParams p) {
...
+     *  This runnable object is used to create a loop that will run the animation.
+     */
+    private Runnable bounceAnimationRunnable = new Runnable() {
+        @Override
+        public void run() {
+            mShouldAnimateCard = !mSettings.isFeatureDiscovered(Settings.FEATURE_ANIMATE_CARD);
+            if (mShouldAnimateCard) {
+                animateFirstCardView();
+
+                mHandler.postDelayed(bounceAnimationRunnable, 10000);
```
frett: `10000` should be a constant at the top of this class so it's easy to adjust.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
gyasistory: I see now those aren't super they are calls to other constructors That is what confused me.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -406,6 +431,69 @@ protected LayoutParams generateLayoutParams(final ViewGroup.LayoutParams p) {
...
+     * @param targetView the Card View
+     */
+    @UiThread
+    private void bounceCardAnimation(final View targetView) {
+
+        mBounceAnimator = ObjectAnimator.ofFloat(targetView, "translationY",
+                targetView.getY(), targetView.getY() + 30, targetView.getY());
+        mBounceAnimator.setInterpolator(BounceUtil::getBounceInOut);
+        mBounceAnimator.setStartDelay(500);
+        mBounceAnimator.setDuration(1500);
```
frett: These durations should be constants at the top of the class so they are easy to change

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -406,6 +431,69 @@ protected LayoutParams generateLayoutParams(final ViewGroup.LayoutParams p) {
...
+
+    /**
+     * This method will create a new instance of the bounce animation.
+     * @param targetView the Card View
+     */
+    @UiThread
+    private void bounceCardAnimation(final View targetView) {
+
+        mBounceAnimator = ObjectAnimator.ofFloat(targetView, "translationY",
+                targetView.getY(), targetView.getY() + 30, targetView.getY());
```
frett: is `30` a px or dp value?

the bounce amount should be a `dimen` defined in a values file.

library/base/src/main/java/org/cru/godtools/base/Settings.java
```diff
@@ -25,6 +25,7 @@
     // feature discovery
     public static final String FEATURE_LANGUAGE_SETTINGS = "languageSettings";
     public static final String FEATURE_LOGIN = "login";
+    public static final String FEATURE_ANIMATE_CARD = "animateCard";
```
frett: I would create 2 features, and call them "tractCardSwiped" and "tractCardClicked" (or similar).

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
frett: also, I would initialize this to `false`, default behavior shouldn't be to bounce.

frett: and rename this to `mBounceFirstCard`.

`mShouldAnimateCard` doesn't indicate if this boolean is about the bouncing animation or the swiping animation.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
frett: This logic right here already cancels any currently running animation: https://github.com/CruGlobal/godtools-android/blob/b7cff6e64839faa7e59a4c8b41bc08d2073cf469/ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java#L316-L321

frett: This also means you wouldn't need to add the `onStartAnimation` callback.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
frett: they are calling `this` and not `super`

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
gyasistory: It is a pixel value as a float.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
gyasistory: OK I will switch it back I was trying to prevent duplicate code.

library/base/src/main/java/org/cru/godtools/base/Settings.java
frett: another option if we don't want to distinguish between swipes and clicks, is just "tractCard". I'm envisioning features are more about the feature itself and not the animation that draws attention to the feature.

So, the feature in this case is cards within tracts, how we draw attention to that is currently via a bounce animation. That could change based on the overlay tests that Ann is running with Brenton. But the feature itself is not going to change.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
frett: it's a toss up, I like `final` because it makes the values immutable, which I value over a few lines of duplicated code.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
frett: we don't want to use a raw pixel value because it's a different size on every device.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -736,4 +824,28 @@ public LayoutParams(@NonNull final LayoutParams p) {
             this.childType = p.childType;
         }
     }
+
+    //region bounce thread calls
+    /**
+     *  This runnable object is used to create a loop that will run the animation.
+     */
+    private Runnable bounceAnimationRunnable = new Runnable() {
```
frett: I would need to check, but this `Runnable` might leak in the `Handler` if the `PageContentLayout` is destroyed.

frett: https://www.androiddesignpatterns.com/2013/01/inner-class-handler-memory-leak.html

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -434,24 +434,24 @@ protected LayoutParams generateLayoutParams(final ViewGroup.LayoutParams p) {
...
     private void animateFirstCardView() {
-        // Cards were in the wrong location need to reset animation each time.
-        for (int i = 0; i < getChildCount(); i++) {
-            final View child = getChildAt(i);
-            final LayoutParams lp = (LayoutParams) child.getLayoutParams();
-            switch (lp.childType) {
-                case CHILD_TYPE_CARD:
-                    bounceCardAnimation(child);
-                    return; // Stop after first Card View
+        if (mActiveCard == null) {
```
frett: something that I picked up from @ryancarlson which he got from Swift (or Rails) is short-circuiting conditions instead of deeply nested if blocks. What I mean by that is:
```
if (mActiveCard != null) {
    return;
}
```
at the beginning of the method so you don't have to nest the rest of the logic in an if block.

This gives the benefit of less code churn as conditionals and amount of them change over time within a method.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -736,4 +816,46 @@ public LayoutParams(@NonNull final LayoutParams p) {
...
+        private void verifyAndSendDelayedMessage(int message) {
+            if (!hasMessages(message)) {
+                sendEmptyMessageDelayed(message, DELAY_BETWEEN_BOUNCE_ANIMATION);
+            }
+        }
+
+        @Override
+        public void handleMessage(Message msg) {
+            super.handleMessage(msg);
+
```
frett:
```
switch(msg.what) {
    case BOUNCE_MESSAGE:
```

switch execution behavior on what message we receive. That way if we ever add in some other type of message processing we can assign it a unique what value and re-use this Handler.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -736,4 +816,46 @@ public LayoutParams(@NonNull final LayoutParams p) {
...
+            PageContentLayout layout = mPageContentLayout.get();
+
+            if (layout != null) {
+                if (layout.mBounceFirstCard) {
+                    layout.animateFirstCardView();
+                    layout.mBounceFirstCard = !layout.isBounceFeatureDiscovered();
+                    verifyAndSendDelayedMessage(BOUNCE_MESSAGE);
+                } else {
+                    if (hasMessages(BOUNCE_MESSAGE)) {
```
frett: no need to check for messages before removing it :)

ui/tract-renderer/src/main/java/org/cru/godtools/tract/viewmodel/PageViewHolder.java
```diff
@@ -116,6 +116,9 @@ void onVisible() {
         } else {
             mHeroViewHolder.markVisible();
         }
+        if (!mPageContentLayout.isBounceFeatureDiscovered()) {
+            mPageContentLayout.setAnimateCard(true);
```
frett: I'm going to rework this logic to enable/disable the bounce on the PCL. I also want to monitor for the card feature to be discovered here.

frett: and I have an idea of one more thing to do with feature discovery :)

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -736,4 +816,46 @@ public LayoutParams(@NonNull final LayoutParams p) {
...
+        @Override
+        public void handleMessage(Message msg) {
+            super.handleMessage(msg);
+
+            PageContentLayout layout = mPageContentLayout.get();
+
+            if (layout != null) {
+                if (layout.mBounceFirstCard) {
+                    layout.animateFirstCardView();
+                    layout.mBounceFirstCard = !layout.isBounceFeatureDiscovered();
```
frett: This is no longer necessary, updates to discovered features happen externally

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -406,6 +430,65 @@ protected LayoutParams generateLayoutParams(final ViewGroup.LayoutParams p) {
...
+     /**
+     * This method will create a new instance of the bounce animation.
+     * @param targetView the Card View
+     */
+    @UiThread
+    private void bounceCardAnimation(final View targetView) {
+        int bounceHeight = getResources().getDimensionPixelSize(R.dimen.card_bounce_height);
+        mAnimation = ObjectAnimator.ofFloat(targetView, View.Y, targetView.getY() - bounceHeight);
+        mAnimation.setInterpolator(mBounceInterpolator);
+        mAnimation.setStartDelay(BOUNCE_ANIMATION_START_DELAY);
```
frett: I don't believe you need this start delay anymore. We are now delaying the animation via the handler.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -406,6 +430,65 @@ protected LayoutParams generateLayoutParams(final ViewGroup.LayoutParams p) {
...
+        this.mBounceFirstCard = animate;
+        if (mBounceFirstCard) {
+            mHandler.sendEmptyMessageDelayed(BOUNCE_MESSAGE, DELAY_FIRST_BOUNCE_ANIMATION);
+        }
+    }
+
+    /**
+     * This method will stop the bounce Animation by setting the AnimateCard variable to false.
+     */
+    private void stopBounceAnimation() {
```
frett: I don't think this method is needed anymore. `mBounceFirstCard` is managed externally only.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -406,6 +430,65 @@ protected LayoutParams generateLayoutParams(final ViewGroup.LayoutParams p) {
...
+        for (int i = 0; i < getChildCount(); i++) {
+            final View child = getChildAt(i);
+            final LayoutParams lp = (LayoutParams) child.getLayoutParams();
+            switch (lp.childType) {
+                case CHILD_TYPE_CARD:
+                    bounceCardAnimation(child);
+                    return; // Stop after first Card View
+            }
+        }
+        stopBounceAnimation();
```
frett: not necessary anymore

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -406,6 +430,65 @@ protected LayoutParams generateLayoutParams(final ViewGroup.LayoutParams p) {
...
+
+    /**
+     * This method will either create a new bounce Animation or start the one that is already created.
+     */
+    @UiThread
+    private void animateFirstCardView() {
+        if (mActiveCard != null) {
+            return;
+        }
+
```
frett: we should also skip bouncing the first card if another animation is running.
```
if (mAnimation != null) {
    return;
}
```

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -406,6 +430,65 @@ protected LayoutParams generateLayoutParams(final ViewGroup.LayoutParams p) {
...
+                case CHILD_TYPE_CARD:
+                    bounceCardAnimation(child);
+                    return; // Stop after first Card View
+            }
+        }
+        stopBounceAnimation();
+
+    }
+
+    public void setAnimateCard(boolean animate) {
```
frett: rename to `setBounceFirstCard(boolean animate)`, so that we know this is related to the bounce animation.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -406,6 +430,65 @@ protected LayoutParams generateLayoutParams(final ViewGroup.LayoutParams p) {
...
+     * @param targetView the Card View
+     */
+    @UiThread
+    private void bounceCardAnimation(final View targetView) {
+        int bounceHeight = getResources().getDimensionPixelSize(R.dimen.card_bounce_height);
+        mAnimation = ObjectAnimator.ofFloat(targetView, View.Y, targetView.getY() - bounceHeight);
+        mAnimation.setInterpolator(mBounceInterpolator);
+        mAnimation.setStartDelay(BOUNCE_ANIMATION_START_DELAY);
+        mAnimation.setDuration(mBounceInterpolator.getTotalDuration(BOUNCE_ANIMATION_DURATION_FIRST_BOUNCE));
+        mAnimation.addListener(mAnimationListener);
```
frett: I think we will need a separate animation listener for bounce animations that doesn't dispatch the active card change event.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
gyasistory: Completely forgot about that good catch.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
gyasistory: Ok

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
gyasistory: Ok

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
gyasistory: In the animateFirstCardView() I am calling stopBounceAnimation so the loop doesn't continue on a page with no cards.  If you would like to remove the method I can just set the boolean value to false instead of calling the method.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
frett: or something similar, just needs to be named something indicative of it relating to bouncing the first card.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
frett: hmm, I think there might be an overloaded meaning here. the `mBounceFirstCard` boolean should just indicate if this view should animate the first card when one is available. if you are clearing the flag here the view won't start animating the first card if one is added to the view later.

frett: I feel that we should never be changing the value of `mBounceFirstCard` within the `PageContentLayout`, any changes to that value should be initiated externally.

frett: want to talk this through?

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -48,10 +53,20 @@
         ViewTreeObserver.OnGlobalLayoutListener {
     private static final int FLING_SCALE_FACTOR = 20;
 
+    private static final int DELAY_FIRST_BOUNCE_ANIMATION = 1000;
+    private static final int DELAY_BETWEEN_BOUNCE_ANIMATION = 10000;
+    private static final int BOUNCE_ANIMATION_START_DELAY = 500;
+    private static final long BOUNCE_ANIMATION_DURATION_FIRST_BOUNCE = 300;
```
frett: haha, I just re-read the name I gave this. it's a poor name because it can be mis-interrupted. I'm going to adjust this name before I merge.

PR #494, opened 2018-09-27
stop any PCL animation currently running if we are changing the active card without animating.

frett:
I really need to work up some UI tests for this logic to verify how things work together

### DIFF
--- a/ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
+++ b/ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
@@ -12,6 +12,7 @@ import android.os.Parcelable;
 import android.support.annotation.IdRes;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.support.annotation.UiThread;
 import android.support.v4.os.ParcelableCompat;
 import android.support.v4.os.ParcelableCompatCreatorCallbacks;
 import android.support.v4.view.AbsSavedState;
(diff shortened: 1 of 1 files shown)


ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -269,6 +269,7 @@ public void changeActiveCard(final int cardPosition, final boolean animate) {
         changeActiveCard(getChildAt(mCardPositionOffset + cardPosition), animate);
     }
 
+    @UiThread
```
gyasistory: The import for UiThread was missing.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
```diff
@@ -12,6 +12,7 @@
 import android.support.annotation.IdRes;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.support.annotation.UiThread;
 import android.support.v4.os.ParcelableCompat;
```
gyasistory: I added it.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/widget/PageContentLayout.java
frett: thanks, i cherry-picked the commit from somewhere it already existed. just got missed, but it would explain the travis failure :)

frett:
going to let travis finish, then I'll merge

PR #495, opened 2018-09-28
Store articles in db

frett:
- Add some synchronization around syncing an AemImport
- utilize Retrofit to fetch AEM json for articles
- create an AemImportRepository that encompasses all logic to store a processed AemImport in the database

### DIFF
--- a/ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/view_model/ArticleViewModel.java
+++ b/ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/view_model/ArticleViewModel.java
@@ -8,6 +8,7 @@ import android.support.annotation.NonNull;
 import android.support.v4.app.FragmentActivity;
 
 import org.cru.godtools.articles.aem.db.ArticleRepository;
+import org.cru.godtools.articles.aem.db.ArticleRoomDatabase;
 import org.cru.godtools.articles.aem.db.AttachmentRepository;
 import org.cru.godtools.articles.aem.model.Article;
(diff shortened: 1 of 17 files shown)


frett:
@gyasistory this is ready for review

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/view_model/ArticleViewModel.java
```diff
@@ -31,7 +32,7 @@ public static ArticleViewModel getInstance(FragmentActivity activity) {
     public ArticleViewModel(@NonNull Application application) {
         super(application);
 
-        mArticleRepository = new ArticleRepository(application);
+        mArticleRepository = ArticleRoomDatabase.getInstance(application).articleRepository();
         mAttchReposistory = new AttachmentRepository(application);
```
gyasistory: So it you would like for us to get repositories from the Database class instead of calling them directly.  That makes since so you know what database you are working with if add multiple databases.

frett:
@gyasistory 👍 I would have noticed eventually when travis kept failing ;)

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/view_model/ArticleViewModel.java
frett: yeah, it also was necessary once I started annotating repositories as `@Dao` so that I could make entire methods `@Transactional`.

PR #499, opened 2018-10-04
WIP: Download article content

frett:
- track the UUID for an article to determine when we need to re-download an article
- Add logic to download a specific article if it hasn't been downloaded or has changed since the last download


frett:
The only reason I keep deleting the old room db schema is because we haven't hit a stable point for it yet where keeping the old schema is worthwhile.

PR #505, opened 2018-10-05
Create AEM web view fragment

gyasistory:
This is the initial set for Web View displaying an AEM Article


ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+    //endregion WebClient
+
+    //region Runnable and Handler
+
+    private Runnable mWebViewRunnable = new Runnable() {
+        @Override
+        public void run() {
+            AsyncTask.execute(() -> {
+                mArticle = getArticleFromKey();
```
gyasistory: Is there a better way to run on the background thread than AsyncTask.execute()

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
```diff
@@ -82,4 +98,17 @@ private void loadInitialFragmentIfNeeded() {
                 .replace(R.id.frame, ArticlesFragment.newInstance(mTool, mLocale, mCategoryId), TAG_MAIN_FRAGMENT)
                 .commit();
     }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        switch (item.getItemId()) {
+            case android.R.id.home:
+                onBackPressed();
```
frett: this breaks the up navigation paradigm

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
```diff
@@ -67,6 +72,17 @@ protected void onStart() {
...
+                .findFragmentByTag(AEMArticleItemFragment.TAG);
+
+        if (fragment == null) {
+            fragment = AEMArticleItemFragment.newInstance(mTool, mLocale, article.uri.toString());
+        } else {
+            fragment.setArticle(article);
+        }
+
+        getSupportFragmentManager().beginTransaction().replace(R.id.frame, fragment,
+                AEMArticleItemFragment.TAG).addToBackStack(AEMArticleItemFragment.TAG).commit();
```
frett: do this in a separate activity, don't overload the ArticlesActivity

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,177 @@
...
+            mArticleKey = Uri.parse(args.getString(ARTICLE_KEY_TAG));
+        }
+    }
+
+    @Nullable
+    @Override
+    public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container,
+                             @Nullable Bundle savedInstanceState) {
+        View view = inflater.inflate(R.layout.fragment_aem_article_item, container, false);
+        ButterKnife.bind(this, view);
```
frett: this is already handled in `onViewCreated` in the base fragment.

frett: so, this method can simply be:
```
return inflater.inflate(R.layout.fragment_aem_article_item, container, false);
```

ui/articles-aem-renderer/src/main/res/layout/fragment_aem_article_item.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:orientation="vertical"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
```
frett: `ConstraintLayout` is not necessary

frett: the `ConstraintLayout` contains a single view that is set to take up all available space. This can be accomplished by making the `WebView` `match_parent` for the width and height and eliminate the extraneous ConstraintLayout

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
gyasistory: OK thanks didn't know that

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
gyasistory: @frett I did this because the back button in the action bar wasn't functioning.  Is there a call in your baseToolsFragment that will resolve this.

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
gyasistory: Should the Activity be in the article-renderer or in articles-aem-renderer

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
frett: that's because it needed the parent activity defined in the android manifest. I did that in this PR: https://github.com/CruGlobal/godtools-android/pull/502

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
frett: articles-aem-renderer

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
gyasistory: I am going to assume that this will be handle if I use a separate Activity.

gyasistory: Ok

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
gyasistory: Ok

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
```diff
@@ -55,6 +59,8 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         }
 
         setContentView(R.layout.activity_generic_fragment);
+
+        Objects.requireNonNull(getSupportActionBar()).setHomeButtonEnabled(true);
```
frett: you shouldn't need to actually call this

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
```diff
@@ -67,6 +73,10 @@ protected void onStart() {
     public void onArticleSelected(@Nullable final Article article) {
         Timber.tag("ArticlesActivity")
                 .d("Article selected: %s", article != null ? article.title : null);
+
+        assert article != null;
```
frett: asserting the article isn't null shouldn't be how you handle a null article being sent to this method.

The `assert mTool != null` works because the activity has a check state when it starts to make sure we have a `mTool` defined, otherwise the activity would never reach the state that this method would be called.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
```diff
@@ -260,6 +260,15 @@ void syncAemImportTask(@NonNull final Uri baseUri, final boolean force) {
...
             enqueueDownloadArticle(article.uri, false);
+            //TODO: Attachments weren't being saved this needs to be refined
+            for (Attachment attachment : article.mAttachments) {
+                new AttachmentRepository(mContext).insertAttachment(attachment);
+                try {
+                    saveAttachmentToStorage(attachment);
+                } catch (IOException e) {
+                    Timber.e(e, "syncAemImportTask: ");
+                }
+            }
```
frett: this is now invalid with how we are actually downloading attachments

ui/articles-aem-renderer/src/main/AndroidManifest.xml
```diff
@@ -1,2 +1,11 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="org.cru.godtools.article.aem" />
+    xmlns:tools="http://schemas.android.com/tools"
+    package="org.cru.godtools.article.aem" >
+    <application>
+        <activity
+            android:name="org.cru.godtools.articles.aem.activity.AemArticleItemActivity"
+            android:theme="@style/Theme.GodTools.Tool.AppBar"
+            android:parentActivityName="org.cru.godtools.article.activity.ArticlesActivity"
+            tools:targetApi="jelly_bean" />
```
frett: you also need to include the parent activity meta-data xml node so that it works on ICS

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/activity/AemArticleItemActivity.java
```diff
@@ -0,0 +1,70 @@
...
+        super.onStart();
+        loadFragmentIfNeeded();
+    }
+    // endregion
+
+    @MainThread
+    private void loadFragmentIfNeeded() {
+        final FragmentManager fm = getSupportFragmentManager();
+
+        if (fm.findFragmentByTag(AEMArticleItemFragment.TAG) != null) {
```
frett: for the fragment tag I would prefer you use a tag name that is local to this activity and not a constant out of the fragment itself.

The purpose of the tag is to identify a specific fragment out of the fragment manager. In theory the fragment manager may have multiple instances of the same fragment, so the distinguisher shouldn't be based on the type of fragment, but on where in the containing activity the fragment you are looking for is attached.

This is why I usually use a local static `TAG_MAIN_FRAGMENT`.

Also, the tag used for this shouldn't be confused with the `TAG` used for log messages. They both are called tag, but are 2 different things.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
frett: actually, name it `findAsync` since it will asynchronously find the article

gyasistory: I changed it I just haven't updated that commit yet

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,157 @@
...
+        return inflater.inflate(R.layout.fragment_aem_article_item, container, false);
+    }
+
+    @Override
+    public void onActivityCreated(@Nullable Bundle savedInstanceState) {
+        super.onActivityCreated(savedInstanceState);
+
+        mAemWebView.setWebViewClient(mWebViewClient);
+
+        mHandler.post(mWebViewRunnable);
```
frett: there isn't a need to use a `Runnable` or `Handler` anymore. You should use a `ViewModel` to track the LiveData object, similar to how it's handled in the `ArticlesFragment`.

frett: https://github.com/CruGlobal/godtools-android/blob/articles/ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java#L68
https://github.com/CruGlobal/godtools-android/blob/articles/ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java#L116-L124

frett: https://github.com/CruGlobal/godtools-android/blob/articles/ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java#L138

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,157 @@
...
+        this.mArticle = article;
+
+        setFragmentViews();
+    }
+
+    private void setFragmentViews() {
+        if (mArticle.content != null) {
+            loadWebViewData();
+        }
+        Objects.requireNonNull(getActivity()).setTitle(mArticle.title);
```
frett: use `requireActivity()` instead of `Objects.requireNonNull(getActivity())`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
gyasistory: Ok

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public void setArticle(Article article) {
+        this.mArticle = article;
+
+        setFragmentViews();
+    }
+
+    private void setFragmentViews() {
+        if (mArticle.content != null) {
```
frett: `mArticle` could be null

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,136 @@
...
+        getArticleFromKey();
+    }
+    //endregion LifeCycle
+
+    //region Article Data
+    private void getArticleFromKey() {
+        mViewModel = ViewModelProviders.of(this).get(AemArticleWebViewModel.class);
+
+        if (!mViewModel.initialized) {
+            ArticleRoomDatabase db = ArticleRoomDatabase.getInstance(requireActivity().getApplicationContext());
```
frett: no need to do `getApplicationContext()`, we already unwrap the Context object within `getInstance`. Also, you could do: `ArticleRoomDatabase.getInstance(requireContext())`.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
gyasistory: good catch

ui/article-renderer/src/main/java/org/cru/godtools/article/activity/ArticlesActivity.java
```diff
@@ -70,6 +73,9 @@ protected void onStart() {
     public void onArticleSelected(@Nullable final Article article) {
         Timber.tag("ArticlesActivity")
                 .d("Article selected: %s", article != null ? article.title : null);
+
+        AemArticleItemActivity.start(this, Objects.requireNonNull(mTool), mLocale,
+                Objects.requireNonNull(article).uri.toString());
```
frett: `Objects.requreNonNull` will throw an exception if the parameter is null, this is the same as leaving the normal `NullPointerException`. what you really should be doing is `if (article != null) {}`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/activity/AemArticleItemActivity.java
```diff
@@ -0,0 +1,71 @@
...
+    // endregion
+
+    //region Life Cycle
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        Bundle extras = getIntent().getExtras();
+        if (extras != null && extras.containsKey(EXTRA_ARTICLE_KEY)) {
+            mArticleKey = extras.getString(EXTRA_ARTICLE_KEY);
```
frett: instead I would do:
```java
if (extras != null) {
   mArticleKey = extras.getString(EXTRA_ARTICLE_KEY, mArticleKey);
}
```

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,135 @@
...
+        if (mArticle != null && mArticle.content != null) {
+            loadWebViewData();
+        }
+        requireActivity().setTitle(mArticle.title);
+    }
+
+    private void loadWebViewData() {
+        StringBuilder builder = new StringBuilder(mArticle.content);
+        builder.insert(mArticle.content.indexOf("<head>") + 7,
+                "<style> img { max-width: 100%; } </style>");
```
frett: move the logic to modify the content into a separate method. that will allow us to test the content modification logic independent of loading that data into the `WebView`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,135 @@
...
+        this.mArticle = article;
+
+        setFragmentViews();
+    }
+
+    private void setFragmentViews() {
+        if (mArticle != null && mArticle.content != null) {
+            loadWebViewData();
+        }
+        requireActivity().setTitle(mArticle.title);
```
frett: `mArticle` can be null here

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    private void loadWebViewData() {
+        StringBuilder builder = new StringBuilder(mArticle.content);
```
frett: `mArticle` can be null here

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,135 @@
...
+        builder.insert(mArticle.content.indexOf("<head>") + 7,
+                "<style> img { max-width: 100%; } </style>");
+        mAemWebView.loadDataWithBaseURL("https://" + mArticleKey.getHost(), builder.toString(),
```
frett: the base URL should be `mArticleKey`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,135 @@
...
+
+        getArticleFromKey();
+    }
+    //endregion LifeCycle
+
+    //region Article Data
+    private void getArticleFromKey() {
+        mViewModel = ViewModelProviders.of(this).get(AemArticleWebViewModel.class);
+
+        if (!mViewModel.initialized) {
```
frett: hmm, this actually can be simplified. I probably can simplify the other `ViewModel` as well. instead of checking `initialized`, you can just check `mViewModel.getArticle == null`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,135 @@
...
+            return super.shouldInterceptRequest(view, url);
+        }
+    }
+    //endregion WebClient
+
+    private AemArticleWebViewModel mViewModel;
+
+    public static class AemArticleWebViewModel extends ViewModel {
+
+        LiveData<Article> getArticle;
```
frett: I would just name this `article`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,135 @@
...
+            } else if (url.endsWith(".css")) {
+                //TODO: Load CSS WebResource
+            }
+
+            return super.shouldInterceptRequest(view, url);
+        }
+    }
+    //endregion WebClient
+
+    private AemArticleWebViewModel mViewModel;
```
frett: move the field to the top of the class with the other fields

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,135 @@
...
+    public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container,
+                             @Nullable Bundle savedInstanceState) {
+        return inflater.inflate(R.layout.fragment_aem_article_item, container, false);
+    }
+
+    @Override
+    public void onActivityCreated(@Nullable Bundle savedInstanceState) {
+        super.onActivityCreated(savedInstanceState);
+
+        mAemWebView.setWebViewClient(mWebViewClient);
```
frett: This should be in `onViewCreated()`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,135 @@
...
+
+        getArticleFromKey();
```
frett: initializing the view model should happen at the end of `onCreate()`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
frett: You are configuring views, the point that they are ready to be configured is in `onViewCreated()`. Also, putting the logic there will allow the view to be configured if the view is recreated during the lifecycle of the fragment.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
frett: Also, this update function may be triggered when there isn't a view created, so the whole method body should be wrapped in a `mAemWebView != null`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,149 @@
...
+        if (mArticle == null || mAemWebView == null || mArticleKey == null) {
+            return;
+        }
+        String data = getUpdatedHtml(mArticle.content);
+        mAemWebView.loadDataWithBaseURL( mArticleKey.toString(), data,
+                "text/html", null, null);
+    }
+
+    @NonNull
+    private String getUpdatedHtml(String content) {
```
frett: perfect 👍

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,149 @@
...
+    //endregion LifeCycle
+
+    //region Article Data
+    private void getArticleFromKey() {
+        mViewModel = ViewModelProviders.of(this).get(AemArticleWebViewModel.class);
+
+        if (mViewModel.article == null) {
+            ArticleRoomDatabase db = ArticleRoomDatabase.getInstance(requireContext());
+            mViewModel.article = db.articleDao().liveFind(mArticleKey);
+            mViewModel.article.observe(this, this::setArticle);
```
frett: move the observe statement out of this if block. If the fragment is recreated due to configuration change, the observe won't be run on the re-created fragment.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Override
+    public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceState) {
+        super.onViewCreated(view, savedInstanceState);
+
+        if (mAemWebView == null || mViewModel == null) {
+            return;
+        }
+        mAemWebView.setWebViewClient(mWebViewClient);
+        mViewModel.article.observe(this, this::setArticle);
```
frett: not the right spot for this observe statement, since it's not the view instance that is observing, but the fragment itself.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
gyasistory: Not sure why this is still showing onActivityCreated this is in onViewCreated.

frett:
I'm going to deal with the merge conflicts for this PR

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -0,0 +1,151 @@
...
+    @Override
+    public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container,
+                             @Nullable Bundle savedInstanceState) {
+        super.onCreateView(inflater, container, savedInstanceState);
+
+        View view = inflater.inflate(
+                R.layout.fragment_aem_article_item,
+                container,
+                false);
+        setViewModel();
```
frett: this should be in `onCreate()`, not `onCreateView()`

frett:
@gyasistory I'm going to merge this in

PR #514, opened 2018-10-12
Use saved resources 4 web view

gyasistory:
Use locally saved resources instead of pulling them from the web.


frett:
now that the main WebView PR has been merged, this PR needs to be cleaned up and simplified

frett:
when you resolved the merge you reverted a bunch of the changes that had been made to the fragment

gyasistory:
OK thanks I will take a look.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -138,21 +155,62 @@ private String injectCss(@NonNull final String content) {
...
             }
 
             return super.shouldInterceptRequest(view, url);
         }
+
+        private WebResourceResponse getResponseFromFile(@NonNull String mimeType, @NonNull String url)
+                throws FileNotFoundException, NoSuchElementException {
+
+            Resource resource = Stream.of(mResources)
+                    .filter(r -> r.getUri().toString().equals(url)).findFirst().get();
```
frett: this is executing on a background thread, so you can just directly query the database for the specific resource and not deal with needing LiveData or a list of all resources attached to this article.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -138,21 +155,62 @@ private String injectCss(@NonNull final String content) {
...
+
+            if (resource == null || resource.getLocalFileName().isEmpty()) {
+                throw new FileNotFoundException(String.format("No local file for %s", url));
```
frett: we shouldn't crash if a resource isn't cached locally

frett: disregard for now, you catch the exception in the calling code

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/db/ResourceDao.kt
```diff
@@ -32,4 +33,11 @@ interface ResourceDao {
         WHERE a.articleUri = :uri
         """)
     fun getAllForArticle(uri: Uri): List<Resource>
+
+    @Query("""
+        SELECT r.*
+        FROM resources AS r JOIN articleResources AS a ON a.resourceUri = r.uri
+        WHERE a.articleUri = :uri
+        """)
+    fun getAllLiveForArticle(uri: Uri): LiveData<List<Resource>>
```
frett: you shouldn't need this LiveData version of the db query

gyasistory: Not using the Live Data will make me have to call the data in a background thread.  If you want me to make that change I will.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/db/ResourceDao.kt
```diff
@@ -1,14 +1,13 @@
...
-import android.arch.lifecycle.LiveData
 import android.arch.persistence.room.Dao
 import android.arch.persistence.room.Insert
 import android.arch.persistence.room.OnConflictStrategy
 import android.arch.persistence.room.Query
 import android.net.Uri
 import android.support.annotation.WorkerThread
 import org.cru.godtools.articles.aem.model.Resource
-import java.util.Date
+import java.util.*
```
frett: I had to change my code style settings within Android Studio to prevent the code formatter from doing wildcard imports.

frett: ktlint will complain about wildcard imports.

Using a wildcard import can lead to unexpected results when new code uses an unexpected class that was included via a wildcard.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -138,21 +147,62 @@ private String injectCss(@NonNull final String content) {
...
+                Timber.d(e);
             }
 
             return super.shouldInterceptRequest(view, url);
         }
+
+        private WebResourceResponse getResponseFromFile(@NonNull String mimeType, @NonNull String url)
+                throws FileNotFoundException {
+
+            Resource resource = ArticleRoomDatabase.getInstance(requireActivity()).resourceDao()
```
frett: we use the `ArticleRoomDatabase` in multiple locations, let's centralize getting the singleton in `onCreate()`.
so:
```java
public void onCreate() {
   super.onCreate();
   mAemDb = ArticleRoomDatabase.getInstance(requireContext());
```

and then this becomes `mAemDb.resoruceDao().find(Uri.parse(url))`

gyasistory: good catch

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -138,21 +147,62 @@ private String injectCss(@NonNull final String content) {
...
         @Override
         public WebResourceResponse shouldInterceptRequest(WebView view, String url) {
-            if (url.endsWith(".jpg")) {
-                //TODO: Load local Image WebResource
-            } else if (url.endsWith(".css")) {
-                //TODO: Load CSS WebResource
+            String extension = MimeTypeMap.getFileExtensionFromUrl(url);
+
+            if (extension.isEmpty()) {
+                return super.shouldInterceptRequest(view, url);
```
frett: We may want to implement a "Not Found" response behavior when we don't recognize the resource that we are trying to load.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -138,21 +147,62 @@ private String injectCss(@NonNull final String content) {
...
             return super.shouldInterceptRequest(view, url);
         }
+
+        private WebResourceResponse getResponseFromFile(@NonNull String mimeType, @NonNull String url)
+                throws FileNotFoundException {
+
+            Resource resource = ArticleRoomDatabase.getInstance(requireActivity()).resourceDao()
+                    .find(Uri.parse(url));
+
+            if (resource == null || resource.getLocalFileName().isEmpty()) {
```
frett: `resource.getLocalFileName().isEmpty()` could cause a `NullPointerException`. How I would probably implement this method is:
```java
final Resource resource = mAemDb.resoruceDao().find(Uri.parse(url));
if(resource == null) {
   return null;
}

final File localFile = resource.getLocalFile(requireContext());
if (localFile == null || !localFile.isFile()) {
   return null;
}

final FileInputStream inputStream = new FileInputStream(localFile);
...
```

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
gyasistory: @frett I think I can just remove it with the switch statement in place it really isn't needed.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -138,21 +150,61 @@ private String injectCss(@NonNull final String content) {
...
+                        return getResponseFromFile(mimeType, url);
+                    case "css":
+                        mimeType = "text/css";
+                        return getResponseFromFile(mimeType, url);
+                }
+            } catch (FileNotFoundException e) {
+                Timber.d(e);
             }
 
             return super.shouldInterceptRequest(view, url);
```
frett: this should also be handled by the not found behavior

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -138,21 +150,72 @@ private String injectCss(@NonNull final String content) {
...
+                    response = getResponseFromFile(mimeType, url);
+                    break;
+                case "css":
+                    mimeType = "text/css";
+                    response = getResponseFromFile(mimeType, url);
+                    break;
+            }
+
+            if (response == null) {
+                return super.shouldInterceptRequest(view, url);
```
frett: you should return a not found `WebResourceResponse` and not let the WebView trigger an actual web request

gyasistory: So you don't want anything to load if we haven't  saved the file also this runs for files that are not covered in the switch statement like ".ico"

frett: yeah, I would also change the switch to always attempt to load the resource from disk, use "application/octet-stream" as the mime-type if you don't have a preset mime-type

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -142,13 +152,63 @@ private String injectCss(@NonNull final String content) {
...
+                    mimeType = "text/css";
+                    response = getResponseFromFile(mimeType, url);
+                    break;
+                default:
+                    mimeType = "application/octet-stream";
+                    response = getResponseFromFile(mimeType, url);
+            }
+
+            if (response == null) {
+                return returnWebResponse(mimeType, new ByteArrayInputStream("".getBytes()));
```
frett: what I meant about returning a Not Found response is return a 404 Response, not just an empty resource

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
```diff
@@ -142,13 +152,63 @@ private String injectCss(@NonNull final String content) {
...
+                    response = getResponseFromFile(mimeType, url);
+                    break;
+                case "css":
+                    mimeType = "text/css";
+                    response = getResponseFromFile(mimeType, url);
+                    break;
+                default:
+                    mimeType = "application/octet-stream";
+                    response = getResponseFromFile(mimeType, url);
+            }
```
frett: your mime-type resolution logic can be simplified. create a constant map that lists extensions with their mimetype.
```java
private static final Map<String, String> MIME_TYPES = ImmutableMap.of(
    "gif", "image/gif",
    "css", "text/css",
    ...
);
```

Then just do a lookup by the extension:
```java
String mimeType = MIME_TYPES.get(extension);
if(mimeType == null) {
    mimeType = "application/octet-stream";
}
```

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/fragment/AEMArticleItemFragment.java
frett: hmm, the response code part of the API is only available starting in Lollipop. From Lollipop on we should return a 404. otherwise the data from the web response object should probably just be null (assuming it doesn't cause issues on older versions of android)

PR #515, opened 2018-10-12
Article Categories support

frett:
- parse categories out of the AEM JSON
- Update ArticlesFragment to filter the list of articles on category if a category is provided


ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/support/AemJsonParser.java
```diff
@@ -94,7 +100,18 @@ private static Article parseArticle(@NonNull final Uri url, @NonNull final JSONO
...
 
-            if (json.has(TAG_CONTENT) && content.has(LAST_MODIFIED_TAG)) {
+            // parse any categories
+            final JSONArray tags = content.optJSONArray(TAG_TAGS);
+            final List<String> categories = new ArrayList<>();
+            for (int i = 0; i < tags.length(); i++) {
+                final String tag = tags.optString(i, "");
+                if (tag.startsWith(PREFIX_TAG_CATEGORY)) {
+                    categories.add(tag.substring(PREFIX_TAG_CATEGORY.length()));
+                }
```
gyasistory: You are looking for  categories in this area.
postman
But there is no categories tag listed I am not sure if this is just some data that isn't suppose to be there or not but I thought the category  was listed as shown below.
postman

frett: it's supposed to be in the `cq:tags` node. the problem is that AEM has decided to cache an old version of the json that isn't accurate. This json accurately reflects what is expected: https://stage.cru.org/content/experience-fragments/questions_about_god/english.9998.json

gyasistory: OK I see it, is there a way around that. Maybe we are going to have to randomly generate the .9999.json call to make sure we are getting fresh data.  I can make that my next Task :)

frett: don't change it yet, caching is good and should be utilized if possible. Plus I'm hoping that our articles are stable and won't require being updated frequently.

frett: and if we do implement cache-busting I would probably do it as `?_={secondsSinceEpoch}`

PR #519, opened 2018-10-12
Launch articles for debug builds

frett:
This change will launch the new articles renderer on debug builds, but leave the legacy everystudent UI in place for release builds


frett:
I'm going to integrate this into your webview PR before we merge it

PR #520, opened 2018-10-15
Articles cache busting

frett:


ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/api/AemApi.java
```diff
@@ -15,12 +15,13 @@
...
 import retrofit2.converter.scalars.ScalarsConverterFactory;
 import retrofit2.http.GET;
+import retrofit2.http.Query;
 import retrofit2.http.Streaming;
 import retrofit2.http.Url;
 
 public interface AemApi {
     @GET
-    Call<JSONObject> getJson(@Url String uri);
+    Call<JSONObject> getJson(@Url String uri, @Query("_") long timestamp);
```
gyasistory: Trying to understand the Query, does it add "?_=1539707467263" to the end of the url?

frett: Yeah, the param name is pointless, but I didn't want it to potentially interfere with other functionality

PR #522, opened 2018-10-15
Use the Category name as the ArticlesActivity title

frett:


frett:
it's a proguard issue I believe. I was about to start into figuring out what was failing, but then had to go to a meeting

PR #525, opened 2018-10-16
AEM tags support

frett:
- Parse AEM tags for a category from the manifest xml
- Track all tags for an article in the AEM article cache database
- Load articles based on the AEM tags in a specific category


gyasistory:
I am not seeing the logic to tie the tags to the Category.  
Is that coming in another PR?

android_emulator_-_pixel_2_api_28_-_google_play_5554

frett:
That's configured in the manifest xml here: https://github.com/CruGlobal/mobile-content-api/pull/166

frett:
I also still need to do a different task of supporting multiple experience fragment templates so it will parse the production articles that have the configured tags

PR #526, opened 2018-10-17
GT-527 - add swipe to refresh view

gyasistory:
https://jira.cru.org/browse/GT-527#


ui/article-renderer/src/main/res/layout/fragment_articles.xml
```diff
@@ -1,11 +1,17 @@
...
+
+    <android.support.v7.widget.RecyclerView
+        android:id="@+id/articles_recycler_view"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:background="@color/white"
+        android:theme="@style/ThemeOverlay.Widget.GodTools.RecyclerView.Articles"
+        app:layoutManager="android.support.v7.widget.LinearLayoutManager"
+        tools:listitem="@layout/list_item_article" />
+</android.support.v4.widget.SwipeRefreshLayout>
```
frett: add a trailing line return. it's a good practice/habit to get into because without it various command line utilities (including `git diff`) behave weirdly. See: https://unix.stackexchange.com/a/18789

frett:
This was what I was referring to with the newline: https://github.com/CruGlobal/godtools-android/pull/526/commits/2aa7d01e189adb72843d2579431b0b8994ddecef

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMFutureRunnable.java
```diff
@@ -0,0 +1,17 @@
...
+    private AEMDownloadManger mManager;
+
+    public AEMFutureRunnable(Context context) {
+        mManager = AEMDownloadManger.getInstance(context);
+    }
+
+    @Override
+    public void run() {
+        mManager.extractAemImportsFromManifestsTask();
+        mManager.syncStaleAemImportsTask();
```
gyasistory: @frett I created this class so I didn't have to change the access modifier for any methods.  Let me know if this and the way I handled the Future Task are the headed in the right direction.

frett: I was actually thinking of relaxing the visibility of the necessary enqueue methods and have them return a `ListenableFuture`.

frett: also, the swipe to refresh should be a targeted force sync, not a sync everything

frett: so, it should end up more directly calling `enqueueSyncAemImport(uri, true)`

frett: the task scheduling methods are meant for external consumption, the actual task methods are meant for internal use only. The reason I made them package-private and not private is to prevent creation of synthetic accessor methods.

See: https://medium.com/thoughts-overflow/effects-of-javas-synthetic-accessor-methods-in-android-bb67b3bac22e

frett: it's a minor thing that helps keep method counts low while we still have to deal with old android and want to avoid multidex if possible.

gyasistory: @frett not sure what you mean by a targeted forced sync can you clarify?

frett: since you are viewing a list of articles for a specific manifest, you should only sync the AEM Imports that the manifest contains

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
```diff
@@ -445,6 +452,10 @@ private File streamResource(@NonNull final InputStream in) throws IOException {
 
     // region PriorityRunnable Tasks
 
+    public ThreadPoolExecutor getExecutor() {
+        return mExecutor;
+    }
```
frett: you shouldn't pull this thread executor out of the DownloadManager

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
```diff
@@ -147,6 +147,13 @@ private void enqueueExtractAemImportsFromManifests() {
         }
     }
 
+    @AnyThread
+    public void enqueueSyncManifestAemImports(Manifest manifest) {
```
frett: this should return a `ListenableFuture`

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
```diff
@@ -147,6 +147,13 @@ private void enqueueExtractAemImportsFromManifests() {
...
+        for (Uri uri : manifest.getAemImports()) {
+            enqueueSyncAemImport(uri, true);
```
frett: you will want to update `enqueueSyncAemImport` to return a `ListenableFuture` as well. Then use `Futures.successfulAsList` to combine all the Futures into a single Future

frett: The force flag should be passed in via the `enqueueSyncManifestAemImports` method.

ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java
```diff
@@ -162,26 +162,22 @@ public void onRefresh() {
...
             }
 
             @Override
             public void onFailure(Throwable t) {
-                Timber.d("onFailure() called with: t = [" + t + "]");
+                Timber.d(t);
                 setRefreshing(false);
             }
-        }, Runnable::run);
+        }, manger.getExecutor());
```
frett: `new MainThreadExecutor()` the class can be found in gto-support-core

gyasistory:
@frett I had trouble getting the changes the way you requested.  The main issue I had was with the Executor it either finished before the task or didn't call the finish at all.  So I created an executor that holds all of the sync Task and run a Future Call back when it is done.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
```diff
@@ -147,24 +147,47 @@ private void enqueueExtractAemImportsFromManifests() {
         }
     }
 
+    @AnyThread
+    public ThreadPoolExecutor enqueueSyncManifestAemImports(Manifest manifest, Boolean force) {
```
frett: it should be returning a ListenableFuture, you will create the ListenableFuture that you are returning in the `SyncAemImportTask` constructor. Then you will complete it in the `run` method of the `SyncAemImportTask`

frett: You shouldn't need to modify any of the executors or use any (aside from the MainThreadExecutor) at all to complete this

ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java
```diff
@@ -124,6 +138,47 @@ private void setupViewModel() {
...
+            public void onSuccess(List<Boolean> result) {
+                setRefreshing(false);
+            }
+
+            @Override
+            public void onFailure(Throwable t) {
+                Timber.e(t);
+                setRefreshing(false);
+            }
+        }, new MainThreadExecutor());
```
frett: Because you don't really care whether the future succeeds or fails, it looks like `ListenableFuture.addListener()` will allow you to simplify this to:
```
        future.addListener(() -> setRefreshing(false), new MainThreadExecutor());
```

gyasistory: Ok I will try that.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
```diff
@@ -147,24 +150,36 @@ private void enqueueExtractAemImportsFromManifests() {
...
+    @AnyThread
+    public ListenableFuture<List<Boolean>> enqueueSyncManifestAemImports(Manifest manifest, Boolean force) {
+
+        List<ListenableFuture<Boolean>> futures = new ArrayList<>();
+
+        for (Uri uri : manifest.getAemImports()) {
+            futures.add(enqueueSyncAemImport(uri, force));
+        }
+
+       return Futures.successfulAsList(futures);
```
frett: looks good and simple now :)

ui/article-renderer/src/main/java/org/cru/godtools/article/fragment/ArticlesFragment.java
frett: the next thing to do is address the potential memory leak by creating a static inner class Runnable with a weak reference to the `SwipeToRefreshLayout`. that way if the view goes away before the sync completes the listener doesn't hold a hard reference to the fragment.

frett: I might have something that already does this, let me take a quick look.

frett: nope, don't have anything that does that immediately. although maybe a `WeakRunnable` class could make sense.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
```diff
@@ -455,6 +470,8 @@ private File streamResource(@NonNull final InputStream in) throws IOException {
         volatile boolean mStarted = false;
         volatile boolean mForce = false;
 
+        final SettableFuture<Boolean> mResult = SettableFuture.create();
```
frett: the only reason I made this a `Boolean` Future is in case we wanted to report whether a task actually did something or not.

frett:
I pulled out your WeakRunnable testing logic and utilized the gto support version

PR #527, opened 2018-10-18
Articles list style updates

frett:
Adjust the UI for the articles list so that it pull colors from the Manifest


frett:
Was testing colors in the manifest

PR #530, opened 2018-10-22
Gt 528 tls 1.2 support

gyasistory:
[https://jira.cru.org/browse/GT-528](https://jira.cru.org/browse/GT-528) 

I made two recommended Changes which I found the first was to Check to see if the TLS play service was installed.

The second was to make sure to enable the TLS in OkHttpClient.  

I believe the change in OkHttpClient should be eventually moved to GTO-Support.  Also the not sure if we want to introduce Play Services to the Application or whether or not to keep it in the MainActivity or the BaseApplication class.

### DIFF
--- a/ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/api/AemApi.java
+++ b/ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/api/AemApi.java
@@ -1,16 +1,31 @@
 package org.cru.godtools.articles.aem.api;
 
 import android.net.Uri;
+import android.os.Build;
 import android.support.annotation.NonNull;
+
+import com.google.common.collect.ImmutableList;
 
 import org.ccci.gto.android.common.api.retrofit2.converter.JSONObjectConverterFactory;
 import org.ccci.gto.android.common.okhttp3.util.OkHttpClientUtil;
 import org.json.JSONObject;
 
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.Socket;
+import java.net.UnknownHostException;
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
 import java.util.concurrent.TimeUnit;
 
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocket;
+import javax.net.ssl.SSLSocketFactory;
+
+import okhttp3.ConnectionSpec;
 import okhttp3.OkHttpClient;
 import okhttp3.ResponseBody;
+import okhttp3.TlsVersion;
 import retrofit2.Call;
 import retrofit2.Retrofit;
 import retrofit2.converter.scalars.ScalarsConverterFactory;
@@ -18,6 +33,7 @@ import retrofit2.http.GET;
 import retrofit2.http.Query;
 import retrofit2.http.Streaming;
 import retrofit2.http.Url;
+import timber.log.Timber;
 
 public interface AemApi {
     @GET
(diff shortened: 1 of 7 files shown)


frett:
I don't think the play services change is necessary due to the nature of GodTools. It would make sense if we were dealing with secure user data.

I do want to see the TLS 1.2 changes added to gto-support and referenced from there. Similar to how we currently reference the global network interceptors for OkHttp3.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/OkHttpClientProvider.java
```diff
@@ -0,0 +1,117 @@
...
+            if (trustManagers.length != 1 || !(trustManagers[0]instanceof X509TrustManager)) {
+                throw new IllegalStateException("Unexpected default trust managers: " + Arrays.toString(trustManagers));
+            }
+            X509TrustManager trustManager = (X509TrustManager) trustManagers[0];
+
+            builder.sslSocketFactory(new Tls12SocketFactory(), trustManager);
+
+            ConnectionSpec cs = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS).tlsVersions(TlsVersion.TLS_1_2).build();
+
+            builder.connectionSpecs(Collections.singletonList(cs));
```
frett: this logic is written to only support TLS 1.2, and disable all other protocols. This is not what we want. We only want to add TLS 1.2 to our list of supported protocols if the current OS supports it.

gyasistory: OK let me work on that.

frett:
How I probably would like to see the actual method call to the gto support library would be something along the lines of:
```java
final SSLSocketFactory sslSocketFactory = SSLSocketFactoryBuilder.create()
.addSupportedProtocolsIfAvailable(TlsVersion.TLS_1_2)
.build()

builder.addSocketFactory(sslSocketFactory)
```

The intention being to only enable additional protocols and not just replace what already exists. The Builder could be extended in the future to remove other protocols if needed.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/OkHttpClientProvider.java
```diff
@@ -0,0 +1,122 @@
...
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509TrustManager;
+
+import okhttp3.ConnectionSpec;
+import okhttp3.OkHttpClient;
+import timber.log.Timber;
+
+public class OkHttpClientProvider {
+
+    public static OkHttpClient.Builder createBuilder() {
```
frett: the builder I was referencing was a brand new builder for the SSLSocketFactory only. How `OkHttpClient` is created can stay completely within the `AemApi` method that originally had it.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/OkHttpClientProvider.java
frett: you don't need to modify the connection spec at all, the OkHttp default connectionspec is fine

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/api/AemApi.java
```diff
@@ -31,17 +30,19 @@
     Call<ResponseBody> downloadResource(@Url Uri uri);
 
     static AemApi buildInstance(@NonNull final String uri) {
-        // create OkHttp client
-        final OkHttpClient.Builder builder = new OkHttpClient.Builder()
-                .connectTimeout(60, TimeUnit.SECONDS)
-                .readTimeout(60, TimeUnit.SECONDS);
```
frett: How I imagine this actually working is:
```java
final SSLSocketFactory sslSocketFactory = DynamicSSLSocketFactory.builder()
        .addEnabledProtocols("TLSv1.2")
        .build();

final OkHttpClient.Builder builder = new OkHttpClient.Builder()
        .connectTimeout(60, TimeUnit.SECONDS)
        .readTimeout(60, TimeUnit.SECONDS)
        .sslSocketFactory(sslSocketFactory);
```

frett: I know `sslSocketFactory(SSLSocketFactory)` is deprecated, but I'm fine with using it for now because it pulls the stock TrustManager automatically.

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/DynamicSSLSocketFactory.java
```diff
@@ -0,0 +1,77 @@
...
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocket;
+import javax.net.ssl.SSLSocketFactory;
+
+public class DynamicSSLSocketFactory extends SSLSocketFactory {
```
frett: I'll add a Builder into this class so that we can customize it's functionality over time without breaking prior usage. Also, the SSLSocketFactory will need to be moved to gto-support-util

frett:
@gyasistory I added a Builder to the `DynamicSSLSocketFactory`.

The benefits of the Builder pattern are:
- allows creation of Immutable objects
- allows future development without needing to change old code
- avoids large constructor and allows simple modification of a single property

frett:
We will want to move this class to GTO support and reference it from there

gyasistory:
@frett It looks like everything is done here except moving it to GTO Support.  If so is that what you would like me to do?

frett:
do you want to open up a PR on GTO support with the SSLSocketFactory added to the gto-support-util module? if not I can take care of it.

I'm sticking it in gto-support-util because it isn't dependent on OkHttp and can be used by other code that needs a custom SSLSocketFactory.

gyasistory:
@frett Yes it will give me some practice on working with GTO-Support.

frett:
@gyasistory now that this class is in GTO support, you should update this PR to reference it from there and remove the local copy of the SSLSocketFactory

PR #532, opened 2018-10-22
handle errors encountered loading a manifest

frett:


frett:
reading on how to handle `InterruptedException`: https://programming.guide/java/handling-interrupted-exceptions.html

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/service/AEMDownloadManger.java
```diff
@@ -225,8 +225,17 @@ void extractAemImportsFromManifestsTask() {
...
+                    final Manifest manifest = mManifestManager.getManifest(translation).get();
+                    repository.addAemImports(translation, manifest.getAemImports());
+                } catch (final InterruptedException e) {
+                    // set interrupted and return immediately
+                    Thread.currentThread().interrupt();
+                    return;
+                } catch (final ExecutionException e) {
+                    Timber.tag(TAG)
+                            .d(e.getCause(), "Unable to add aem imports from manifest");
+                }
```
gyasistory: Just curious why not do a return for the second Catch Statement.  Is there a reason you want this to continue.

frett: yes, there are other translations that may still be processed within this method. the `InterruptedException` is special because it means something else is trying to stop the thread executing, and in that scenario you want to exit as quickly as you safely can

PR #543, opened 2018-10-26
GT-544 ensure we only call startActivity from Activity Contexts

gyasistory:
[https://jira.cru.org/browse/GT-544]()

Android P is enforcing an always intended requirement that "you cannot start an activity from a non-activity context unless you pass the intent flag FLAG_ACTIVITY_NEW_TASK. If you attempt to start an activity without passing this flag, the activity does not start, and the system prints a message to the log."

Our preferred mechanism for dealing with this should be to just ensure we are launching activities from other activities. If there is a situation where we can't ensure this we should examine that code to figure out what actually makes sense.

### DIFF
--- a/app/src/main/java/org/cru/godtools/activity/AboutActivity.java
+++ b/app/src/main/java/org/cru/godtools/activity/AboutActivity.java
@@ -20,8 +20,8 @@ import static org.cru.godtools.base.util.LocaleUtils.getDeviceLocale;
 public final class AboutActivity extends BasePlatformActivity {
     private static final String TAG_MAIN_FRAGMENT = "mainFragment";
 
-    public static void start(@NonNull final Activity context) {
-        context.startActivity(new Intent(context, AboutActivity.class));
+    public static void start(@NonNull final Activity activity) {
+        activity.startActivity(new Intent(activity, AboutActivity.class));
     }
 
     /* BEGIN lifecycle */
--- a/app/src/main/java/org/cru/godtools/activity/AboutActivity.java
+++ b/app/src/main/java/org/cru/godtools/activity/AboutActivity.java
@@ -1,6 +1,6 @@
 package org.cru.godtools.activity;
 
-import android.content.Context;
+import android.app.Activity;
 import android.content.Intent;
 import android.os.Bundle;
 
@@ -20,7 +20,7 @@ import static org.cru.godtools.base.util.LocaleUtils.getDeviceLocale;
 public final class AboutActivity extends BasePlatformActivity {
     private static final String TAG_MAIN_FRAGMENT = "mainFragment";
 
-    public static void start(@NonNull final Context context) {
+    public static void start(@NonNull final Activity context) {
         context.startActivity(new Intent(context, AboutActivity.class));
     }
 
--- a/ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
+++ b/ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
@@ -167,11 +167,11 @@ public class TractActivity extends BaseToolActivity
         return new Intent(context, TractActivity.class).putExtras(extras); //.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
     }
 
-    public static void start(@NonNull final Activity context, @NonNull final String toolCode,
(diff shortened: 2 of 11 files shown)


ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
```diff
@@ -163,12 +164,14 @@ public static Intent createIntent(@NonNull final Context context, @NonNull final
                                       @NonNull final Locale... languages) {
         final Bundle extras = new Bundle();
         populateExtras(extras, toolCode, languages);
-        return new Intent(context, TractActivity.class).putExtras(extras);
+        return new Intent(context, TractActivity.class).putExtras(extras); //.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
```
gyasistory: @frett not sure if I should add the flag or not here since it has no available Activity when called.

frett: leave it off, I would say responsibility for adding that flag should come from the caller that is using the intent. This method is just meant to create a base intent to launch this activity.

app/src/main/java/org/cru/godtools/activity/AboutActivity.java
```diff
@@ -20,7 +20,7 @@
 public final class AboutActivity extends BasePlatformActivity {
     private static final String TAG_MAIN_FRAGMENT = "mainFragment";
 
-    public static void start(@NonNull final Context context) {
+    public static void start(@NonNull final Activity context) {
```
frett: small nit, just to be thorough I would rename the `context` param to `activity`

gyasistory: OK

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
frett: I believe this is used locally and by the shortcuts logic

ui/tract-renderer/src/main/java/org/cru/godtools/tract/adapter/ManifestPagerAdapter.java
```diff
@@ -206,7 +206,7 @@ private void checkForModalEvent(@NonNull final Event event) {
             final Manifest manifest = mPage.getManifest();
             for (final Modal modal : mPage.getModals()) {
                 if (modal.getListeners().contains(event.id)) {
-                    ModalActivity.start(mPageView.getContext(), manifest.getManifestName(), manifest.getCode(),
+                    ModalActivity.startWithFlag(mPageView.getContext(), manifest.getManifestName(), manifest.getCode(),
```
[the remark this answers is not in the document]
gyasistory: Any idea how you would like to handle this cause I can think about 3 but I don't really like most of them.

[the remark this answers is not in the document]
frett: probably hand off triggering the modal to the hosting activity via callbacks. Create a `showModal(@NonNull Modal modal)` callback. But let's work on that in a separate PR

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
frett: remove the commented `//.addFlags(...`

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
```diff
@@ -163,12 +164,14 @@ public static Intent createIntent(@NonNull final Context context, @NonNull final
...
     }
 
-    public static void start(@NonNull final Context context, @NonNull final String toolCode,
+    public static void start(@NonNull final Activity activity, @NonNull final String toolCode,
                              @NonNull final Locale... languages) {
-        context.startActivity(createIntent(context, toolCode, languages));
+        final Bundle extras = new Bundle();
+        populateExtras(extras, toolCode, languages);
+        activity.startActivity(new Intent(activity, TractActivity.class).putExtras(extras));
```
frett: you should still utilize the `createIntent()` method so we aren't duplicating this logic

PR #544, opened 2018-10-26
GT-549 Persist AEM Article WebView through configuration changes

gyasistory:
Right now configuration changes will cause AEM Article WebView to reset to the top of the article. This will lead to a poor experience. We should attempt to persist the WebView through configuration changes

[https://jira.cru.org/browse/GT-549]()


ui/articles-aem-renderer/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,7 @@
         <activity
             android:name="org.cru.godtools.articles.aem.activity.AemArticleActivity"
             android:parentActivityName="org.cru.godtools.article.activity.ArticlesActivity"
+            android:configChanges="keyboard|keyboardHidden|orientation|screenLayout|uiMode|screenSize"
```
frett: use retainInstance on the fragment instead

frett: example logic here: https://github.com/CruGlobal/android-thekey/blob/master/thekey-view-dialog-fragment/src/main/java/me/thekey/android/view/dialog/LoginDialogFragment.java#L55

frett: it's older logic, but I'd prefer we avoid overriding configuration changes as much as possible

frett:
basically you will want to do this: https://developer.android.com/guide/topics/resources/runtime-changes#RetainingAnObject

whether it's done using setRetainInstance or a ViewModel class doesn't matter too much. Although using a ViewModel sounds cleaner.

frett:
from earlier on that page:
> It is not recommended to handle configuration changes yourself due to the hidden complexity of handling the configuration changes.

frett:
I would say let's use a ViewModel to persist the WebView (and WebViewClient) through configuration changes.

PR #545, opened 2018-10-29
make tool renderers rotatable for debug builds, but fix in portrait for release builds

frett:
The purpose of this is to provide an easy way to test how layouts adjust to different screen sizes on debug builds


frett:
this doesn't appear to be working anymore

frett:
Found my issue, it helps if I put the resource in the right directory

PR #547, opened 2018-10-29
update the Terms of Use URL

frett:


frett:
This is fairly trivial, so I'm just going to merge it

PR #548, opened 2018-10-29
Adobe ActivityLifecycleCallbacks

frett:
utilize a standard way to monitor when activities resume/pause without needing to add logic to every activity

### DIFF
--- a/library/analytics/src/debug/java/org/cru/godtools/analytics/TimberAnalyticsService.java
+++ b/library/analytics/src/debug/java/org/cru/godtools/analytics/TimberAnalyticsService.java
@@ -1,7 +1,5 @@
 package org.cru.godtools.analytics;
 
-import android.app.Activity;
-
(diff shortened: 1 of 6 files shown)


library/analytics/src/debug/java/org/cru/godtools/analytics/TimberAnalyticsService.java
```diff
@@ -33,18 +31,6 @@ public static synchronized TimberAnalyticsService getInstance() {
...
-        Timber.tag("AnalyticsService")
-                .d("onActivityResume(%s)", activity.getClass());
-    }
-
-    @Override
-    public void onActivityPause(@NonNull final Activity activity) {
-        Timber.tag("AnalyticsService")
-                .d("onActivityPause(%s)", activity.getClass());
-    }
-
```
frett: This was for debugging only, so I'm alright not logging it anymore.

PR #552, opened 2018-10-30
Article model update

frett:
- remove unused date created & date modified fields
- convert Article model to Kotlin
- add shareUrl and shortShareUrl fields
- first Room db migration


frett:
well that makes migrations a pain. there isn't a SQLite way to drop a single column without recreating the table. This problem is arising because we got rid of a couple columns

gyasistory:
Yeah I figured that was the cause.  That is frustrating if there is no way to drop a column.

frett:
@gyasistory added a migration test and for now decided to just keep the old columns to simplify migrations

PR #554, opened 2018-10-30
GT-542 - standardize snowplow data

wrandall22:
This PR standardizes the data we're sending to snowplow to match the standard found [here](https://github.com/CruGlobal/user-data-platform/wiki/tracker-protocol#mobile-app-events).


wrandall22:
Yep, I'm aware, thanks. Not sure how the team wants to handle that.

PR #555, opened 2018-10-30
GT-553 added Linkify to extract url from TextView

gyasistory:
https://jira.cru.org/browse/GT-553

This is what Ann requested about making links in the TextView Clickable.

### DIFF
--- a/app/src/main/res/layout/fragment_tool_details.xml
+++ b/app/src/main/res/layout/fragment_tool_details.xml
@@ -96,6 +96,7 @@
             android:layout_marginLeft="@dimen/tools_details_margin_horiz"
             android:layout_marginRight="@dimen/tools_details_margin_horiz"
             android:layout_marginTop="8dp"
+            android:autoLink="web|email"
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -282,6 +280,7 @@ private void updateViews() {
         bindShares(mShares, mTool);
         if (mDescription != null) {
             mDescription.setText(ModelUtils.getTranslationDescription(getContext(), mLatestTranslation, mTool));
+            Linkify.addLinks(mDescription, Linkify.WEB_URLS);
```
gyasistory: @frett let me know if we need to try to control the link color.

frett: This can be accomplished with the `android:autoLink="email|web"` xml attribute. see: https://github.com/CruGlobal/godtools-android/blob/develop/app/src/main/res/values/styles_about.xml#L6

frett:
on a side note it looks like your code style settings got changed, it shouldn't have reordered all the imports like that

PR #559, opened 2018-11-01
convert AemArticleActivity to Kotlin

frett:


frett:
Feel free to ask questions about the logic, I'm still learning and there are some things that might have a  more understandable implementation available

ui/articles-aem-renderer/src/main/java/org/cru/godtools/articles/aem/activity/AemArticleActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+            state != BaseToolActivity.STATE_LOADED -> state
+            article?.content == null -> BaseToolActivity.STATE_LOADING
+            else -> BaseToolActivity.STATE_LOADED
+        }
+    }
+
+    @MainThread
+    private fun loadFragmentIfNeeded() {
+        // The fragment is already present
+        if (supportFragmentManager.findFragmentByTag(TAG_MAIN_FRAGMENT) != null) return
```
frett: Kotlin feels a lot more like a scripting language, it gives you 20 ways to do the same thing.

I haven't decided how I want to handle simple if blocks yet, if we want to always include braces, or allow simple one-liners like this.

PR #560, opened 2018-11-01
GT-539 AppSee

gyasistory:
https://jira.cru.org/browse/GT-539 

AppSee is a company that offers insight into user behavior. JesusFilm has been testing it and is extending a spot to GodTools in their contract. 

Code will need to be incorporated on Android so that we can learn about user's behavior. I am fine if App See is only incorporated on Android as it is the majority of our users. 

This will not be actionable until at least the beginning of November when Jesus Film gives us the green light and the code that is needed. 

All the following Task must be complete prior to merge

- https://jira.cru.org/browse/GT-554

- https://jira.cru.org/browse/GT-555

- https://jira.cru.org/browse/GT-556

- https://jira.cru.org/browse/GT-557


ui/base/src/main/java/org/cru/godtools/base/ui/activity/BaseActivity.java
```diff
@@ -37,6 +39,7 @@ protected void onCreate(@Nullable final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         mAnalytics = AnalyticsService.getInstance(this);
         mEventBus = EventBus.getDefault();
+        Appsee.start();
```
frett: does this have to be started for every activity? or can it be started once at the main entry point?

I would really prefer that we don't include the Appsee dependency into the base-ui module if possible, this will cause it to be included in our instant app as well as our main app.

frett: Also, Appsee.start() doesn't take the activity context, so maybe a better place to start it is in the main app Application class.

gyasistory: https://www.appsee.com/docs/android/native

In the documentation it says to include it in every activity that is an entry point into the app.

gyasistory: I tried it in the Application and it didn't work.  I did move it to MainActivity

frett:
much better, only 28 lines changed so far :)

gyasistory:
@frett I am done with this PR but I am going to leave it as WIP until I know for sure that Brenton is OK with the data the way it is coming in.

library/analytics/src/main/java/org/cru/godtools/analytics/AppSeeAnalyticService.java
```diff
@@ -0,0 +1,111 @@
...
+import org.greenrobot.eventbus.Subscribe;
+import org.greenrobot.eventbus.ThreadMode;
+
+import java.util.Map;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.UiThread;
+
+public class AppSeeAnalyticService implements AnalyticsService, Application.ActivityLifecycleCallbacks, AppseeListener {
```
frett: you shouldn't need to implement `AnalyticsService`

library/analytics/build.gradle
```diff
@@ -25,6 +25,9 @@ android {
 dependencies {
     implementation project(':library:base')
 
+    implementation "com.appsee:appsee-android:${deps.appSee.android}"
+    implementation "com.crashlytics.sdk.android:crashlytics:${deps.crashlytics}"
+
```
frett: At some point we talked about not including this logic in the analytics module because it would then get included in the instant app

gyasistory: Forgot about that so I will move it to the App module

frett: you can put the logic in the main app module and still have it listen for analytics events

app/build.gradle
```diff
@@ -15,6 +16,7 @@ android {
 
         proguardFile getDefaultProguardFile('proguard-android-optimize.txt')
         proguardFile 'proguard-searchview.pro'
+        consumerProguardFiles 'proguard-appsee.pro'
```
frett: I posted this on another PR I believe. You want `proguardFile` and not `consumerProguardFiles`.

`consumerProguardFiles` is for proguard rules that get bundled with a library so the app consuming the library can apply them.

app/src/main/java/org/cru/godtools/GodToolsApplication.java
```diff
@@ -75,4 +76,12 @@ protected void startServices() {
                 .clientId(THEKEY_CLIENTID)
                 .eventsManager(new EventBusEventsManager());
     }
+
+    @Override
+    protected void configureAnalyticsServices() {
+        super.configureAnalyticsServices();
+
+        // Need to keep AppSee in Gradle file in order to work
+        AppSeeAnalyticService.getInstance(getApplicationContext());
```
frett: you can just pass in `this` instead of needing to call `getApplicationContext()`

library/analytics/src/main/java/org/cru/godtools/analytics/AppSeeAnalyticService.java
```diff
@@ -0,0 +1,111 @@
...
+    private static AppSeeAnalyticService sInstance;
+
+    private AppSeeAnalyticService(Context context) {
+        EventBus.getDefault().register(this);
+        ((Application) context.getApplicationContext()).registerActivityLifecycleCallbacks(this);
+        Appsee.addAppseeListener(this);
+    }
+
+    @NonNull
+    public static synchronized AppSeeAnalyticService getInstance(Context context) {
```
frett: change this method to `start` and have it take an `Application` instead of `Context`

library/analytics/src/main/java/org/cru/godtools/analytics/AppSeeAnalyticService.java
```diff
@@ -0,0 +1,111 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.UiThread;
+
+public class AppSeeAnalyticService implements AnalyticsService, Application.ActivityLifecycleCallbacks, AppseeListener {
+
+    @Nullable
+    private static AppSeeAnalyticService sInstance;
+
+    private AppSeeAnalyticService(Context context) {
```
frett: have the constructor take the `Application` class instead of `Context` so that it simplifies registering the ActivityLifecycleCallbacks

library/analytics/src/main/java/org/cru/godtools/analytics/AppSeeAnalyticService.java
```diff
@@ -0,0 +1,111 @@
...
+    @Override
+    public void onActivityDestroyed(final Activity activity) {}
+    //endregion LifeCycle Callbacks
+
+    //region AppSee LifeCycle Callbacks
+    @Override
+    public void onAppseeSessionStarting(final AppseeSessionStartingInfo appseeSessionStartingInfo) {
+        String crashlyticsAppSeeId = Appsee.generate3rdPartyId("Crashlytics",
+                                                               false);
+        Crashlytics.getInstance().core.setString(
```
frett: this can just be `Crashlytics.setString()`

library/analytics/src/main/java/org/cru/godtools/analytics/AppSeeAnalyticService.java
```diff
@@ -0,0 +1,111 @@
...
+    @Subscribe(threadMode = ThreadMode.MAIN)
+    public void onAnalyticScreenEvent(@NonNull final AnalyticsScreenEvent event) {
+        Appsee.startScreen(event.getScreen());
+    }
+
+    @Subscribe
+    public void onAnalyticActionEvent(@NonNull final AnalyticsActionEvent event) {
+        Map<String, Object> data = (Map<String, Object>) event.getAttributes();
+        if (data != null) {
+            Appsee.addEvent(event.getAction(), data);
```
frett: how do these events show up? do we need to add an Appsee analytics type and filter action events to only send when they are for the Appsee analytics system?

library/analytics/src/main/java/org/cru/godtools/analytics/AppSeeAnalyticService.java
frett: only reason I'm saying `start` vs `getInstance` is because this service is just being started, we never interact directly with our singleton

library/analytics/build.gradle
```diff
@@ -25,6 +25,8 @@ android {
 dependencies {
     implementation project(':library:base')
 
+    implementation "com.crashlytics.sdk.android:crashlytics:${deps.crashlytics}"
```
frett: you can remove this crashlytics dependency from here, it's not being used anymore :)

build.gradle
```diff
@@ -73,6 +77,8 @@ buildscript {
     dependencies {
         classpath 'com.android.tools.build:gradle:3.2.1'
         classpath 'com.google.gms:google-services:4.1.0'
+        //noinspection GradleDependency
```
frett: this noinspection line should be removed, it is outdated and no longer applicable

build.gradle
```diff
@@ -87,7 +93,7 @@ plugins {
 
 group 'org.cru.android'
 version '5.0.19-SNAPSHOT'
-ext.versionCode = (grgit.log(includes:['HEAD']).size() * 2) + 4025200
+ext.versionCode = (grgit.log(includes:['HEAD']).size() * 2) + 4026000
```
frett: also, you reverted a versionCode change from develop, this should go back to the `4025200` number

app/src/main/java/org/cru/godtools/service/AppSeeAnalyticService.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @UiThread
+    @Subscribe(threadMode = ThreadMode.MAIN)
+    public void onAnalyticScreenEvent(@NonNull final AnalyticsScreenEvent event) {
+        if (!event.isForSystem(AnalyticsSystem.APPSEE)) {
+            return;
+        }
+        // AppSee should match Adobe for integration
+        String siteSection = (SITE_SECTION_TOOLS.equals(event.getAdobeSiteSection()) ||
+                SITE_SECTION_MENU.equals(event.getAdobeSiteSection())) ? event.getAdobeSiteSection() :
```
gyasistory: @frett for the Site Section do you think I should put the Screen name behind the Site section if is tools or menu.  For example "menu - privacy policy".  If not all of the tools and menu screens will be tracked as one.

app/src/main/java/org/cru/godtools/service/AppSeeAnalyticService.java
```diff
@@ -0,0 +1,128 @@
...
+
+        String screen = siteSubSection != null ? String.format("%s - %s", siteSection, siteSubSection) : siteSection;
+
+        Appsee.startScreen(screen);
+    }
+
+    @Subscribe
+    public void onAnalyticActionEvent(@NonNull final AnalyticsActionEvent event) {
+        if (!event.isForSystem(AnalyticsSystem.APPSEE)) {
+            return;
```
gyasistory: @frett I also noticed that the events are not listed as AppSee events.  Wasn't sure if that is correct or not.

gyasistory: It is just the Pray question not the language toggle

app/src/main/java/org/cru/godtools/service/AppSeeAnalyticService.java
frett: that's why I'm surprised you are using sitesection and not just using the screen name

app/src/main/java/org/cru/godtools/service/AppSeeAnalyticService.java
frett: yeah, we will want to update those tools to trigger the analytics events for Appsee as well

gyasistory: Is that something done in Android or the API?

frett: API

frett:
@gyasistory why are you using the sitesection & sitesubsection instead of the screen event name? the sitesection and subsection are passed as attributes to Adobe and are not part of the screen name.

app/src/main/java/org/cru/godtools/service/AppSeeAnalyticService.java
```diff
@@ -0,0 +1,117 @@
...
+    @NonNull
+    public static synchronized AppSeeAnalyticService start(Application application) {
+        if (sInstance == null) {
+            sInstance = new AppSeeAnalyticService(application);
+        }
+        return sInstance;
+    }
+
+    @UiThread
+    @Subscribe(threadMode = ThreadMode.MAIN)
```
frett: any specific reason you are processing this on the main thread?

app/src/main/java/org/cru/godtools/service/AppSeeAnalyticService.java
```diff
@@ -0,0 +1,117 @@
...
+    public void onAnalyticScreenEvent(@NonNull final AnalyticsScreenEvent event) {
+        if (!event.isForSystem(AnalyticsSystem.APPSEE)) {
+            return;
+        }
+
+        Appsee.startScreen(event.getScreen());
+    }
+
+    @Subscribe
```
frett: this should specify a threadMode that matches screen events.

gyasistory:
@frett I am sorry I am just now seeing these comment I will work on this now if that is OK with you.

app/src/main/java/org/cru/godtools/service/AppSeeAnalyticService.java
gyasistory: I am still running some test but it looks like I am not getting consistent reports using ASYNC.

frett: k, if ASYNC doesn't work try MAIN_ORDERED for the threadMode

frett: that would keep it on the main thread, but guarantee execution order of events being processed.

PR #561, opened 2018-11-01
GT-555 add App See Key

gyasistory:
https://jira.cru.org/browse/GT-555


app/build.gradle
```diff
@@ -92,13 +92,18 @@ android {
             if (project.hasProperty('crashlyticsBetaBuild')) {
                 signingConfig signingConfigs.crashlyticsBeta
             }
+            //AppSee key
+            manifestPlaceholders = [appSeekey:"f05dbdbcd6524a7089bdcfbac5c9be11"]
```
frett: interesting, this is the first time I've seen this property

app/build.gradle
```diff
@@ -143,6 +148,7 @@ dependencies {
     implementation "org.ccci.gto.android:gto-support-sync:${deps.gtoSupport}"
     implementation "org.ccci.gto.android:gto-support-util:${deps.gtoSupport}"
 
+    implementation "com.appsee:appsee-android:+"
```
frett: I know documentation tends to suggest using wildcard versions, but we will want an explicit version specified. It helps prevent unexpected changes in behavior when a new version of the library is released with an unexpected change

appsee-keep.pro
```diff
@@ -0,0 +1,7 @@
+-keep class com.appsee.** { *; }
+-dontwarn com.appsee.**
+-keep class android.support.** { *; }
+-keep interface android.support.** { *; }
+-keep class androidx.** { *; }
+-keep interface androidx.** { *; }
+-keepattributes SourceFile,LineNumberTable
```
frett: these proguard rules need to really be scoped down, as they stand it will keep a lot of excess methods and functionality we don't use.

app/build.gradle
gyasistory: I knew this was going to be a soar spot.  It wasn't apparently clear what the release version was but I will spend sometime today to see if I can find one.

app/build.gradle
gyasistory: I found it when I was doing some research on how to handle it.

app/build.gradle
frett: you can look to see what version it ended up pulling in. Expand the "External Libraries" item in your project file list

gyasistory: I found it.

app/build.gradle
```diff
@@ -16,6 +16,7 @@ android {
 
         proguardFile getDefaultProguardFile('proguard-android-optimize.txt')
         proguardFile 'proguard-searchview.pro'
+        consumerProguardFiles 'proguard-appsee.pro'
```
frett: should be `proguardFile`.

`consumerProguardFile` indicates that the rules should be bundled with a library project

PR #563, opened 2018-11-01
Utilize Firebase Dynamic Links to generate article share links

frett:


frett:
I'm going to hold off on merging this until we figure out our actual share strategy for articles

frett:
The best we can do with the firebase dynamic links is to provide shorter links for the canonical Url. It is not possible with the current design to have these links deeplink into the android app but still open the canonical URL everywhere else.

PR #573, opened 2018-11-07
GT-552 Invalid Tool for Tools with no data

gyasistory:
https://jira.cru.org/browse/GT-552

### DIFF
--- a/library/xml-model/src/main/java/org/cru/godtools/xml/model/Manifest.java
+++ b/library/xml-model/src/main/java/org/cru/godtools/xml/model/Manifest.java
@@ -61,7 +61,7 @@ public final class Manifest extends Base implements Styles {
     private static final ImageScaleType DEFAULT_BACKGROUND_IMAGE_SCALE_TYPE = ImageScaleType.FILL;
     private static final int DEFAULT_BACKGROUND_IMAGE_GRAVITY = ImageGravity.CENTER;
 
-    private enum Type {
+    public enum Type {
         TRACT, ARTICLE, UNKNOWN;
 
         public static final Type DEFAULT = TRACT;
@@ -171,7 +171,7 @@ public final class Manifest extends Base implements Styles {
         return Text.getText(mTitle);
     }
 
-    @Nullable
+    @NonNull
     public Type getType() {
         return mType;
     }
--- a/library/xml-model/src/main/java/org/cru/godtools/xml/model/Manifest.java
+++ b/library/xml-model/src/main/java/org/cru/godtools/xml/model/Manifest.java
@@ -172,6 +172,11 @@ public final class Manifest extends Base implements Styles {
     }
 
     @Nullable
+    public Type getType() {
+        return mType;
+    }
+
+    @Nullable
     public static String getTitle(@Nullable final Manifest manifest) {
         return manifest != null ? manifest.getTitle() : null;
     }
--- a/ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
+++ b/ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
@@ -558,8 +558,8 @@ public class TractActivity extends BaseToolActivity
                 continue;
             }
 
-            // skip any language that we don't have a manifest for
-            if (mManifests.get(i) == null) {
+            // skip any language that isn't loaded
+            if (determineLanguageState(i) != STATE_LOADED) {
                 continue;
             }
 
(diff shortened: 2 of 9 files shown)


ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
```diff
@@ -417,8 +417,9 @@ protected int determineActiveToolState() {
     }
 
     private int determineLanguageState(final int languageIndex) {
-        if (mManifests.get(languageIndex) != null) {
-            return STATE_LOADED;
+        Manifest manifest = mManifests.get(languageIndex);
+        if (manifest != null) {
+            return manifest.getPages().size() > 0 ? STATE_LOADED : STATE_NOT_FOUND;
```
frett: within the Manifest object there is a `mType` field, expose that via a `getType()` accessor and utilize that value to determine what type of tool this is

library/xml-model/src/main/java/org/cru/godtools/xml/model/Manifest.java
```diff
@@ -171,6 +171,11 @@ public String getTitle() {
         return Text.getText(mTitle);
     }
 
+    @Nullable
```
frett: this can be `@NonNull` since `mType` is declared as NonNull

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
```diff
@@ -417,8 +417,9 @@ protected int determineActiveToolState() {
     }
 
     private int determineLanguageState(final int languageIndex) {
-        if (mManifests.get(languageIndex) != null) {
-            return STATE_LOADED;
+        Manifest manifest = mManifests.get(languageIndex);
+        if (manifest != null) {
+            return Tool.Type.TRACT.equals(manifest.getType()) ? STATE_LOADED : STATE_NOT_FOUND;
```
frett: `Type` is an enum, so you can do `manifest.getType() == Type.TRACT`. Also this is currently the wrong `Type` enum you are comparing against, you need to make the Manifest.Type enum definition public so you can access it here.

frett:
This logic will need to be more robust so it can handle the opposite scenario as well, opening any of the article activities referencing a tract tool.

gyasistory:
@frett That would result in a Article Manifest with no pages correct?

frett:
there wouldn't be categories or any AEM imports. but you really should look at the manifest type to determine if it's going to be valid or not.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
```diff
@@ -417,8 +417,9 @@ protected int determineActiveToolState() {
     }
 
     private int determineLanguageState(final int languageIndex) {
-        if (mManifests.get(languageIndex) != null) {
-            return STATE_LOADED;
+        Manifest manifest = mManifests.get(languageIndex);
+        if (manifest != null) {
+            return Manifest.Type.TRACT == manifest.getType() ? STATE_LOADED : STATE_NOT_FOUND;
         } else if (isSyncToolsDone() && mTranslations.indexOfKey(languageIndex) >= 0 &&
```
gyasistory: @frett so are you saying along with this Type check I should check to see if a category or AemImport exist?

frett:
I need to take a deeper look at this, unfortunately TractActivity is quite complex. I want to make sure there aren't any unforeseen interactions with all the multiple tract toggling logic

gyasistory:
@frett will do

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.java
```diff
@@ -414,9 +414,11 @@ protected int determineActiveToolState() {
...
 
     private int determineLanguageState(final int languageIndex) {
-        if (mManifests.get(languageIndex) != null) {
-            return STATE_LOADED;
-        } else if (isSyncToolsDone() && mTranslations.indexOfKey(languageIndex) >= 0 &&
+        final Manifest manifest = mManifests.get(languageIndex);
+        final boolean syncDone = isSyncToolsDone();
+        if (manifest != null && syncDone) {
+            return Manifest.Type.TRACT == manifest.getType() ? STATE_LOADED : STATE_NOT_FOUND;
+        } else if (syncDone && mTranslations.indexOfKey(languageIndex) >= 0 &&
```
frett: I'm definitely going to want to re-examine this. if we have a manifest we want to show it even if there is still a sync process running.

gyasistory: I can remove it I wasn't sure if we should be displaying the screen if it was still syncing.

frett:
@gyasistory I made the following changes:
- add a `STATE_INVALID_TYPE` constant, right now it operates the same as `STATE_NOT_FOUND`, but this will make it easier to implement a new INVALID UI if desired
- address a couple more locations that directly used `mManifests` to determine if a manifest was valid in `TractActivity`
- create a `isSupportedType()` method for `BaseSingleToolActivity` that will allow implementors to implement `STATE_INVALID_TYPE` in a consistent manner
- renamed `ArticleBaseSingleToolActivity` to `BaseArticleActivity`

Ideally we would put `BaseArticleActivity` in the article-renderer module, but due to our dependencies between article-aem-renderer and article-renderer this isn't possible.

frett:
@gyasistory I'm going to merge this in

PR #580, opened 2018-11-10
we can't specify an orientation for a transparent activity

frett:
this crash: https://fabric.io/cru/android/apps/org.keynote.godtools.android/issues/5be6cfe6f8b88c29635700f5?time=last-ninety-days

I had fixed this in the past, but I re-introduced the bug when I adjusted orientations for all activities.


frett:
I had fixed this in the past, but I re-introduced the bug when I was adjusted orientations for all activities.

PR #588, opened 2018-11-15
convert TranslationRef to Kotlin

frett:


ui/article-aem-renderer/src/main/java/org/cru/godtools/article/aem/model/TranslationRef.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.room.ForeignKey
+import androidx.room.Index
+import androidx.room.PrimaryKey
+import org.cru.godtools.model.Language
+import org.cru.godtools.model.Translation
+import java.util.Locale
+import javax.annotation.concurrent.Immutable
+
+fun Translation?.toTranslationRefKey(): TranslationRef.Key? {
+    return this?.run {
```
frett: @gyasistory something fun you can do in Kotlin, `this` might be null... :)

PR #611, opened 2019-01-10
GT-568 Articles share link

frett:
We now have canonicalUrls in the article json meta-data


frett:
there is an article in the stage environment that has a canonicalUrl defined for it

PR #618, opened 2019-03-20
GT-589 : Update download_initial_content.gradle

gyasistory:


gyasistory:
@frett no it didn't make a difference

frett:
K, wasn't sure of it would have mattered

PR #723, opened 2019-10-07
Firebase app distribution

frett:


frett:
@gyasistory yeah, I did run it in travis ci while working on the solution. once this is merged it should trigger a build and you should end up with an email about it

PR #741, opened 2019-10-29
GT-601  : include footer at bottom of cards to indicate next card

gyasistory:
https://jira.cru.org/browse/GT-601

### DIFF
--- a/ui/tract-renderer/src/main/java/org/cru/godtools/tract/viewmodel/CardViewHolder.java
+++ b/ui/tract-renderer/src/main/java/org/cru/godtools/tract/viewmodel/CardViewHolder.java
@@ -1,5 +1,6 @@
 package org.cru.godtools.tract.viewmodel;
 
+import android.content.Context;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.TextView;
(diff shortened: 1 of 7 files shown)


ui/tract-renderer/src/main/java/org/cru/godtools/tract/viewmodel/CardViewHolder.java
```diff
@@ -114,6 +131,75 @@ private void bindLabel() {
...
+        mCardPositionView.setText(positionText);
+        if (isPrayerForm(mModel) || isPrayerSelection(mModel)) {
+            mPreviousCardView.setVisibility(View.INVISIBLE);
+            mPreviousCardView.setEnabled(false);
+            mCardPositionView.setVisibility(View.INVISIBLE);
+            mNextCardView.setVisibility(View.INVISIBLE);
+            mNextCardView.setEnabled(false);
+            return;
+        }
+        Context localContext = LocaleUtils.localizeContextIfPossible(mCardView.getContext(), locale);
```
frett: did Ann specifically ask to render the labels in the tool language and not the platform language?

gyasistory: That was my understanding but I will verify.

frett: My gut is to leave the navigation labels in the device language since it's meant to help the owner of the device navigate. If I'm presenting it to someone in their native language I may not know what next and previous in their language is

ui/tract-renderer/src/main/java/org/cru/godtools/tract/viewmodel/CardViewHolder.java
```diff
@@ -114,6 +131,75 @@ private void bindLabel() {
...
+            }
+        }
+        return false;
+    }
+
+    private int getCardCount() {
+        int count = 0;
+        if (mModel != null) {
+            for (Card card : mModel.getPage().getCards()) {
+                if (!isPrayerForm(card) || !isPrayerSelection(card)) {
```
wrandall22: Should this be `&&` instead of `||`? Maybe something clearer would be:
```
if (isPrayerForm(card) || isPrayerSelection(card)) {
  continue;
}
count++;
```

ui/tract-renderer/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="previous">previous</string>
+    <string name="next">next</string>
+</resources>
```
wrandall22: Should have newline at end of file?

ui/tract-renderer/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
```
frett: in order for this to not collide with other imported translations in OneSky this file needs to have a unique name, `strings.xml` will collide with the other strings.xml file in this app

ui/tract-renderer/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,5 @@
...
+    <string name="previous">previous</string>
```
frett: let's name these something more unique. when phrases are translated to other languages they can be different based on their context. In order to allow translators to do this we need to make sure we allow multiple "next" phrases based on where we use them in the app. the easiest way is to use some context in phrase names. so the phrase should probably be `tract_card_next` and `tract_card_previous`

PR #747, opened 2019-11-11
Release Build

gyasistory:


frett:
aside from moving the changes into master, this doesn't trigger any release.

wrandall22:
@frett curious about process in releases:  let's say a bug fix and a feature request are in the same test build, the bug fix is needed to go out, but the feature request needs tweaks; do we hold the bug fix back until all the feature request tweaks are finished?

frett:
@wrandall22 with the process we use in MPDX it's not possible to push the bug fix separate from the feature request. with GodTools assuming we don't merge develop into master until we are ready to actually release it is possible to do a small bugfix release without pushing out the feature request

frett:
Ann should have access to a test build with whatever the current state of develop is, so there is no need to merge to master for QA in GodTools

frett:
GodTools process is more manual, which yields greater dexterity in how we do releases vs MPDX's automated process using fastlane

wrandall22:
That seems to put a lot of pressure on the QA process to be an extremely quick turnaround, since QA can't start until we merge to master and deploy a test build.

frett:
for MPDX yes because we opted to utilize the play store internal test channel for QA. that coupled with the need for versions to always increase means that we can't release a bug fix build without building on top of the version currently in internal test

frett:
MH would live under the same constraints

frett:
There is nothing immediately preventing you from publishing older code under a new version code, but that could introduce issues for people that had the QA app with newer code upgrading to the bugfix release with older code.

Database migrations are written in one direction only and there is potential to downgrade the schema in that upgrade

frett:
@gyasistory what are your reasons for creating this PR? is it so Ann can QA, or is it because she is ready to release these changes to users?

gyasistory:
> @gyasistory what are your reasons for creating this PR? is it so Ann can QA, or is it because she is ready to release these changes to users?

It is for QA

frett:
This isn't necessary for Ann to QA changes

PR #753, opened 2019-11-12
GT-642 : tool home video integration

gyasistory:
https://jira.cru.org/browse/GT-642


app/src/main/res/layout-v17/fragment_tool_details.xml
```diff
@@ -0,0 +1,133 @@
...
+
+            <org.ccci.gto.android.common.picasso.view.SimplePicassoImageView
+                android:id="@+id/banner"
+                android:adjustViewBounds="true"
+                android:background="@color/gray_E6"
+                android:scaleType="centerCrop"
+                app:layout_aspectRatio="210%"
+                app:layout_widthPercent="100%" />
+
+            <com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.YouTubePlayerView
```
frett: instead of duplicating the entire layout I believe you should be able to use an `<import />` tag with a layout containing `<merge />` for v16 or a layout with the player view for v17+

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -308,6 +331,24 @@ private void updateViews() {
         }
     }
 
+    private void updateOverviewVideo() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+            if (mVideoBanner != null) {
```
wrandall22: Is this guaranteed from the above call? Or perhaps its not because its Android and is a class attribute?

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -280,6 +289,20 @@ private void updateViews() {
...
         bindShares(mShares, mTool);
+        if (mBanner != null) {
+            mBanner.asImageView().setVisibility(
+                    (mTool != null && TextUtils.isEmpty(mTool.getOverviewVideo())) ||
+                            Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR1 ?
+                            View.VISIBLE : View.GONE);
+        }
+        if (mVideoBanner != null) {
+            mVideoBanner.setVisibility(
+                    mTool != null && !TextUtils.isEmpty(mTool.getOverviewVideo()) ? View.VISIBLE :
```
wrandall22: This conditional logic is here twice, would it be better to stick the `setVisibility` call inside the if block below, then have an else setting it to `View.GONE`?

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -280,6 +289,20 @@ private void updateViews() {
             mTitle.setText(ModelUtils.getTranslationName(getContext(), mLatestTranslation, mTool));
         }
         bindShares(mShares, mTool);
+        if (mBanner != null) {
+            mBanner.asImageView().setVisibility(
```
wrandall22: Since this is being added, how was the banner being set to visible before this?

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
wrandall22: Also, would we want to check the SDK version before setting visibility?

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -308,6 +331,24 @@ private void updateViews() {
...
 
+    private void updateOverviewVideo() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+            if (mVideoBanner != null) {
+                mVideoBanner.setEnableAutomaticInitialization(false);
+                getLifecycle().addObserver(mVideoBanner);
+                mVideoBanner.addYouTubePlayerListener(new AbstractYouTubePlayerListener() {
+                    @Override
+                    public void onReady(@NotNull final YouTubePlayer youTubePlayer) {
+                        if (mTool != null && mTool.getOverviewVideo() != null) {
```
wrandall22: Is this guaranteed based on how this is being called? Or perhaps its not because its Android and these are class attributes?

app/src/main/res/layout-v17/fragment_tool_details.xml
gyasistory: @frett can we talk in person I am not 100% clear on what you are referring to.

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
gyasistory: @wrandall22 I can't guarantee when `onReady` is called.  So to prevent null pointer errors I am checking the values before use.

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
gyasistory: @wrandall22 The banner was always visible before. Now that we are displaying it only when there is no video present I wanted to remove it from the view.

app/src/main/res/layout-v17/tool_detail_banner_view.xml
```diff
@@ -0,0 +1,39 @@
...
+            app:layout_widthPercent="100%"
+            app:showFullScreenButton="false"
+            tools:visibility="gone" />
+
+        <ProgressBar
+            android:id="@+id/download_progress"
+            style="@style/Widget.GodTools.Tools.Download.ProgressBar"
+            android:layout_width="match_parent"
+            android:layout_height="10dp"
+            android:layout_alignBottom="@id/banner"
```
wrandall22: Does this need to be dynamic based on which banner is showing?

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
gyasistory: > Also, would we want to check the SDK version before setting visibility?

We would want to check the version because there may be an instance where we have a video but the player is not able to be played. (The current video implementation doesn't work on lower versions of Android

app/src/main/res/layout/tool_detail_banner_view.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
```
wrandall22: This appears to be exactly the same as the one in `layout-v17`. Is one of the supposed to not include the video player? Or be different in some way?

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
wrandall22: Ok, so that should probably be added to this conditional then.

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
gyasistory: @wrandall22 I placed this null check for reusability

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
gyasistory: @wrandall22 `mVideoBanner` will be null on Older devices.

wrandall22: I still think the conditional could be simplified to something like:
```
if (mVideoBanner != null) {
  if (mTool != null && !TextUtils.isEmpty(mTool.getOverviewVideo())) {
    mVideoBanner.setVisibility(View.VISIBLE);
    updateOverviewVideo();
  } else {
    mVideoBanner.setVisibility(View.GONE);
  }
}
```

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -288,6 +297,20 @@ private void updateViews() {
...
+                            View.VISIBLE : View.GONE);
+        }
+        if (mVideoBanner != null) {
+            if (mTool != null && !TextUtils.isEmpty(mTool.getOverviewVideo())) {
+                mVideoBanner.setVisibility(View.VISIBLE);
+                updateOverviewVideo();
+            } else {
+                mVideoBanner.setVisibility(View.GONE);
+            }
+        }
```
frett: you could probably simplify a lot of this logic by doing something like this:
```java
final boolean hasOverviewVideo = mVideoBanner != null && mTool != null && mTool.getOverviewVideo() != null
if (mBanner != null) {
    mBanner.asImageView().setVisibility(hasOverviewVideo ? View.GONE : View.VISIBLE);
}
if (mVideoBanner != null) {
    mVideoBanner.setVisibilty(hasOverviewVideo ? View.VISIBLE : View.GONE);
    if (hasOverviewVideo) {
        updateOverviewVideo();
    }
}
```

frett: I left off the version check here because the video banner view is only present on versions of android that support it.

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -306,6 +329,24 @@ private void updateViews() {
...
+    private void updateOverviewVideo() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+            if (mVideoBanner != null) {
+                mVideoBanner.setEnableAutomaticInitialization(false);
+                getLifecycle().addObserver(mVideoBanner);
+                mVideoBanner.addYouTubePlayerListener(new AbstractYouTubePlayerListener() {
+                    @Override
+                    public void onReady(@NotNull final YouTubePlayer youTubePlayer) {
+                        if (mTool != null && mTool.getOverviewVideo() != null) {
+                            String videoId = mTool.getOverviewVideo().split("=")[1];
```
frett: this is a really brittle way to handle this

app/src/main/res/layout-v17/tool_detail_banner_video_view.xml
```diff
@@ -0,0 +1,15 @@
...
+        android:id="@+id/video_banner"
+        android:adjustViewBounds="true"
+        android:background="@color/gray_E6"
+        android:scaleType="centerCrop"
+        app:autoPlay="false"
+        app:layout_aspectRatio="210%"
+        app:layout_widthPercent="100%"
+        app:showFullScreenButton="false" />
+
+</merge>
```
frett: in this specific layout you don't need the `<merge>` tags, this can just be the `YouTubePlayerView`

app/src/main/res/layout/fragment_tool_details.xml
```diff
@@ -23,6 +23,9 @@
                 app:layout_aspectRatio="210%"
                 app:layout_widthPercent="100%" />
 
+
```
frett: excess whitespace

app/src/main/res/layout/tool_detail_banner_video_view.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<merge >
+
+</merge>
```
frett: you can collapse this down to `<merge />`

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
frett: we also need to do something to try and allow this to be forward compatible if we change where we host these overview videos in the future.

frett: there are 2 issues here.
1. youtube video urls come in many forms, splitting on `=` to get the video id is incredibly brittle. (e.g. these video link will fail: https://youtu.be/dQw4w9WgXcQ https://www.youtube.com/watch?v=dQw4w9WgXcQ&feature=youtu.be)
2. how do we handle changing from youtube to another provider in the future without breaking existing apps.

issue 2 can be addressed by tweaking the attribute name in the API and you would just have to update the JsonApi annotation. Issue 1 needs a bit more robust youtube link parsing, along with treating an unhandled link format as having no overview video.

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -34,6 +39,7 @@
 import org.cru.godtools.shortcuts.GodToolsShortcutManager;
 import org.cru.godtools.shortcuts.GodToolsShortcutManager.PendingShortcut;
 import org.cru.godtools.util.ActivityUtilsKt;
+import org.jetbrains.annotations.NotNull;
```
frett: you should use the AndroidX `@NonNull` annotation instead of the jetbrains `@NotNull` annotation

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -306,6 +329,24 @@ private void updateViews() {
         }
     }
 
+    private void updateOverviewVideo() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
```
frett: you don't need this version check, `mVideoBanner` will only be non-null if this version condition is met

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
frett: I added youtube urls to 3 more tools on stage, 2 of them are valid but different forms, and 1 is not a video link and should be invalid.

frett: I updated the attribute name to be `attr-about-overview-video-youtube`

gyasistory: @frett I think it would be best just to place the youtube VideoId in the API.  Since we can easily recreate a URL if we need to also this will eliminate the first issue.  To resolve issue number two we can set the JSON attribute to `attr-about-overview-video-youtube`. This way we can change the trailing text to match the provider.

frett: I was thinking this as well and I'm a bit torn. we can discuss briefly this afternoon

frett: talked with Ann, she is fine with just using the youtube video id, so I updated the attribute to return the video id only

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -288,6 +295,17 @@ private void updateViews() {
...
 
+        boolean hasOverviewVideo = mVideoBanner != null && mTool != null &&
+                !TextUtils.isEmpty(mTool.getOverviewVideo());
+        if (mBanner != null) {
+            mBanner.asImageView().setVisibility(hasOverviewVideo ? View.GONE : View.VISIBLE);
+        }
+        if (mVideoBanner != null) {
+            mVideoBanner.setVisibility(hasOverviewVideo ? View.VISIBLE : View.GONE);
+            if (hasOverviewVideo) {
+                updateOverviewVideo();
```
frett: what happens if this function is triggered while a video is already playing?

the `updateViews()` method can be triggered multiple times as data is loaded/reloaded.

app/src/main/res/layout/fragment_tool_details.xml
```diff
@@ -31,6 +33,7 @@
                 android:layout_alignBottom="@id/banner"
```
wrandall22: How does this work if the `@id/banner` element (`SimplePicassoImageView`) is not visible and the `@id/video_banner` element (`YouTubePlayerView`) is visible?

PR #755, opened 2019-11-14
update org.jetbrain:annotations

gyasistory:
This seems like it wasn't caught by dependabot.


frett:
The version number is repeated between this location and a force resolution block. We should make it a variable

gyasistory:
> The version number is repeated between this location and a force resolution block. We should make it a variable

@frett OK resolved

PR #759, opened 2019-11-19
GT-624 :  provide mini tutorial for home screen

gyasistory:
https://jira.cru.org/browse/GT-624


gyasistory:
@wrandall22 and @frett  waiting on images from Justin.

app/src/main/java/org/cru/godtools/activity/OnBoardingActivity.kt
```diff
@@ -7,12 +7,13 @@ import org.cru.godtools.R
 import org.cru.godtools.adapter.OnboardingPagerAdapter
 
 class OnBoardingActivity : AppCompatActivity(), OnBoardingCallbacks {
-    val viewPager by lazy { findViewById<ViewPager>(R.id.onboarding_viewpager) }
+    private lateinit var viewPager: ViewPager
```
wrandall22: It would be very helpful, if in the commit message, you'd put the why.

app/src/main/res/layout/onboarding_explore_slide.xml
```diff
@@ -0,0 +1,85 @@
...
+            <com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.YouTubePlayerView
+                android:id="@+id/youTubePlayerView"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                bind:showFullScreenButton="false"
+                bind:showSeekBar="false"
+                bind:showVideoCurrentTime="false"
+                bind:showVideoDuration="false"
+                bind:showYouTubeButton="false"
+                bind:videoId="9HRYIFXbT_s" />
```
wrandall22: Are we meaning for this to remain hard-coded?

app/src/main/java/org/cru/godtools/adapter/OnboardingPagerAdapter.kt
```diff
@@ -0,0 +1,21 @@
...
+    FragmentPagerAdapter(fm) {
+
+    override fun getItem(position: Int): Fragment {
+        return OnboardingSlideFragment.newInstance(position).apply {
+            setCallback(callbacks)
+        }
+    }
+
+    override fun getCount(): Int {
+        return 4
```
wrandall22: Is there no way to determine this dynamically?

app/src/main/res/layout/onboarding_explore_slide.xml
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:bind="http://schemas.android.com/apk/res-auto">
```
frett: for consistency I would prefer we use the `app` prefix instead of `bind`

app/src/main/res/layout/onboarding_explore_slide.xml
```diff
@@ -0,0 +1,85 @@
...
+            android:layout_marginRight="32dp"
+            android:layout_marginBottom="32dp"
+            android:gravity="top|center_horizontal"
+            android:orientation="vertical"
+            bind:layout_constraintBottom_toTopOf="@+id/onboarding_next_button"
+            bind:layout_constraintEnd_toEndOf="parent"
+            bind:layout_constraintStart_toStartOf="parent"
+            bind:layout_constraintTop_toBottomOf="@+id/onboarding_video_slide_text">
+
+            <!-- The YouTube player is in a linear layout because you can't set gravity on it. -->
```
frett: why not?

app/src/main/res/layout/onboarding_menu_slide.xml
```diff
@@ -0,0 +1,59 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:bind="http://schemas.android.com/apk/res-auto"
```
frett: again, would prefer we use `app` instead of `bind` for the xml namespace prefix

library/base/src/main/java/org/cru/godtools/base/Settings.java
```diff
@@ -27,6 +27,7 @@
     public static final String FEATURE_LOGIN = "login";
     public static final String FEATURE_TRACT_CARD_SWIPED = "tractCardSwiped";
     public static final String FEATURE_TRACT_CARD_CLICKED = "tractCardClicked";
+    public static final String FEATURE_TUTORIAL_VIEWED = "tutorialViewed";
```
frett: I would name this feature `FEATURE_TUTORIAL = "tutorial"`

app/src/main/java/org/cru/godtools/adapter/OnboardingPagerAdapter.kt
frett: a better pattern for this is to create an enum with what the pages are and use that to drive the pager adapter. This makes it much easier to add/remove pages and reorder them in the future.

frett: for example: https://github.com/CruGlobal/mpdx-android/blob/develop/app/src/main/java/org/mpdx/features/appeals/details/AppealDetailsPagerAdapter.kt

because you are just loading a layout, you may even want to include the layout id within the enum

app/src/main/java/org/cru/godtools/activity/OnBoardingActivity.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.cru.godtools.R
+import org.cru.godtools.adapter.OnboardingPagerAdapter
+
+class OnBoardingActivity : AppCompatActivity(), OnBoardingCallbacks {
+    private lateinit var viewPager: ViewPager
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_onboarding)
+        val pagerAdapter = OnboardingPagerAdapter(this, supportFragmentManager)
```
wrandall22: Is `pagerAdapter` useful for anything? Why not put this after the `viewPager` instantiation and do `viewPager.adapter = OnboardingPagerAdapter(this, supportFragmentManager)`?

app/src/main/res/layout/onboarding_explore_slide.xml
gyasistory: yes

app/build.gradle
```diff
@@ -146,10 +146,12 @@ dependencies {
     implementation project(':ui:base-app')
     implementation project(':ui:shortcuts')
     implementation project(':ui:tract-renderer')
+    implementation project(':ui:tutorial-renderer')
```
frett: `ui:tutorial`, `renderer` is specific to tool renderers

app/build.gradle
```diff
@@ -190,6 +193,7 @@ dependencies {
 
     kapt "com.jakewharton:butterknife-compiler:${deps.butterknife}"
     kapt "org.greenrobot:eventbus-annotation-processor:${deps.eventbus}"
+    compile project(path: ':ui:tutorial-renderer')
```
frett: you shouldn't need this since you are already including it as an `implementation` dependency

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.kt
```diff
@@ -379,5 +384,13 @@ abstract class BasePlatformActivity : BaseDesignActivity(), NavigationView.OnNav
         }
     }
 
+    private fun launchOptInTutorial() {
+        startActivity(Intent(this, org.cru.godtools.tutorial.activity.OptInOnBoardingActivity::class.java))
+    }
+
+    fun launchBakedInTutorial() {
+        startActivity(Intent(this, BakedInOnBoardingActivity::class.java))
+    }
```
frett: what is the difference between these 2?

ui/tutorial-renderer/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+    implementation "androidx.constraintlayout:constraintlayout:${deps.androidX.constraintLayout}"
+    implementation "androidx.recyclerview:recyclerview:${deps.androidX.recyclerView}"
+
+    implementation "org.ccci.gto.android:gto-support-core:${deps.gtoSupport}"
+    implementation "org.ccci.gto.android:gto-support-picasso:${deps.gtoSupport}"
+    implementation "org.ccci.gto.android:gto-support-recyclerview:${deps.gtoSupport}"
+    implementation "org.ccci.gto.android:gto-support-util:${deps.gtoSupport}"
+    implementation "me.relex:circleindicator:${deps.circleindicator}"
+
+    implementation "com.google.guava:guava:${deps.guava}"
```
frett: what are you using guava for in this module? is it something that can be replaced by kotlin?

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/adapter/BakedInOnBoardingPagerAdapter.kt
```diff
@@ -0,0 +1,30 @@
...
+) : FragmentPagerAdapter(fragmentManager, BEHAVIOR_RESUME_ONLY_CURRENT_FRAGMENT) {
+
+    override fun getItem(position: Int): Fragment =
+        BakedInOnBoardingFragment.newInstance(OnBoardingPages.values()[position].layout).also {
+            it.setCallback(callbacks)
+        }
+
+    override fun getCount(): Int = OnBoardingPages.values().size
+
+    private enum class OnBoardingPages(@StringRes val layout: Int) {
```
frett: `layout` isn't a `@StringRes`, it is a `@LayoutRes`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/fragment/BakedInOnBoardingFragment.kt
```diff
@@ -0,0 +1,70 @@
...
+                inflater,
+                container,
+                false
+            ).also { it.callback = callback }.root
+            else -> BakedInOnboardingWelcomeBinding.inflate(
+                inflater,
+                container,
+                false
+            ).also { it.callback = callback }.root
+        }
```
frett: This can be greatly simplified, look at what `DataBindingUtil` provides. (I may be wrong with the class name)

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/adapter/OptInOnBoardingPagerAdapter.kt
```diff
@@ -0,0 +1,30 @@
...
+import androidx.annotation.StringRes
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentManager
+import androidx.fragment.app.FragmentPagerAdapter
+import org.cru.godtools.tutorial.R
+import org.cru.godtools.tutorial.fragment.OptInOnBoardingSlideFragment
+import org.cru.godtools.tutorial.util.OnBoardingCallbacks
+
+class OptInOnBoardingPagerAdapter(val callbacks: OnBoardingCallbacks, fm: FragmentManager) :
+    FragmentPagerAdapter(fm, BEHAVIOR_RESUME_ONLY_CURRENT_FRAGMENT) {
```
frett: There is no need to make these fragments since you are just using data binding with them, you should utilize `DataBindingPagerAdapter` instead.

https://github.com/CruGlobal/android-gto-support/blob/master/gto-support-viewpager/src/main/java/org/ccci/gto/android/common/viewpager/adapter/DataBindingPagerAdapter.kt

ui/tutorial-renderer/src/main/res/layout/activity_baked_in_onboarding.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
```
frett: what's the point of the `ConstraintLayout` here?

ui/tutorial-renderer/src/main/res/layout/activity_optin_onboarding.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
```
frett: again, what's the point of the `ConstraintLayout` here?

ui/tutorial-renderer/src/main/res/layout/baked_in_onboarding_final.xml
```diff
@@ -0,0 +1,15 @@
...
+        <variable
+            name="callback"
+            type="org.cru.godtools.tutorial.util.OnBoardingCallbacks" />
+    </data>
+
+    <androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent">
+
+    </androidx.constraintlayout.widget.ConstraintLayout>
```
frett: why use an empty `ConstraintLayout` here?

ui/tutorial-renderer/src/main/res/layout/optin_onboarding_explore_slide.xml
```diff
@@ -0,0 +1,88 @@
...
+            android:layout_marginRight="32dp"
+            android:layout_marginBottom="32dp"
+            android:gravity="top|center_horizontal"
+            android:orientation="vertical"
+            app:layout_constraintBottom_toTopOf="@+id/onboarding_next_button"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@+id/onboarding_video_slide_text">
+
+            <!-- The YouTube player is in a linear layout because you can't set gravity on it. -->
```
frett: gravity is probably the wrong attribute for whatever you were trying to do. There shouldn't be need to nest this is another layout

app/build.gradle
gyasistory: @frett I am still working on this PR and I am going to clean up the Gradle files on my final pass.

gyasistory:
@frett and @wrandall22 sorry this is going to be a very large PR.

app/build.gradle
frett: although since it sounds like there are 2 different sets of tutorial screens, this could be fine for the module name

ui/tutorial-renderer/src/main/res/layout/activity_optin_onboarding.xml
gyasistory: I will need it later when I add the indicator.

ui/tutorial-renderer/src/main/res/layout/activity_baked_in_onboarding.xml
gyasistory: I will need it later when I add the indicator.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/adapter/OptInOnBoardingPagerAdapter.kt
gyasistory: @frett I am using different ViewDataBinding so I don't see how this will work for me.  I am going to try to remove the fragment and just use a basic PageAdapter (to simplify things).

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.kt
gyasistory: They are two different Tutorials

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/adapter/BakedInOnBoardingPagerAdapter.kt
```diff
@@ -0,0 +1,55 @@
...
+            container,
+            false
+        ).also { binding ->
+            when (binding) {
+                is BakedInOnboardingWelcomeBinding -> binding.callback = callbacks
+                is BakedInOnboardingReadyBinding -> binding.callback = callbacks
+                is BakedInOnboardingToolsBinding -> binding.callback = callbacks
+                is BakedInOnboardingOthersBinding -> binding.callback = callbacks
+                is BakedInOnboardingFinalBinding -> binding.callback = callbacks
+            }
```
frett: `binding.setVariable(BR.callback, callbacks)`

app/build.gradle
frett: this should be unnecessary since you already specify the dependency on line 149

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/OptInOnBoardingActivity.kt
```diff
@@ -0,0 +1,57 @@
...
+        setContentView(R.layout.activity_optin_onboarding)
+        viewPager = findViewById(R.id.onboarding_viewpager)
+        viewPager.adapter = OnBoardingPagerAdapter(this).also {
+            it.onBoardingPagesLayout = listOf(
+                R.layout.optin_onboarding_explore_slide,
+                R.layout.optin_onboarding_prepare_slide,
+                R.layout.optin_onboarding_try_slide,
+                R.layout.optin_onboarding_menu_slide
+            )
+            it.notifyDataSetChanged()
```
frett: any `notifyDataSetChanged()` calls should be internal to the adapter, not from an external source

frett: triggering it from an external source has 1 of 2 issues. it's either unnecessary, or prone to be forgotten. the adapter should be the component that is in charge of knowing when it's data was updated.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/OptInOnBoardingActivity.kt
```diff
@@ -0,0 +1,57 @@
...
+        }
+        val indicator: CircleIndicator = findViewById(R.id.on_boarding_indicator)
+        indicator.setViewPager(viewPager)
```
frett: there is no point in declaring an `indicator` variable here

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/adapter/OnBoardingPagerAdapter.kt
```diff
@@ -0,0 +1,50 @@
...
+import androidx.viewpager.widget.PagerAdapter
+import org.cru.godtools.tutorial.BR
+import org.cru.godtools.tutorial.R
+import org.cru.godtools.tutorial.animation.animateToNextText
+import org.cru.godtools.tutorial.databinding.BakedInOnboardingWelcomeBinding
+import org.cru.godtools.tutorial.util.OnBoardingCallbacks
+
+class OnBoardingPagerAdapter(val callbacks: OnBoardingCallbacks) : PagerAdapter() {
+
+    lateinit var onBoardingPagesLayout: List<Int>
```
frett: this should probably just be `pages`

ui/tutorial-renderer/src/main/res/drawable/img_baked_in_others.xml
```diff
@@ -0,0 +1,107 @@
...
+          android:startX="108.369"
+          android:endY="79.8484"
+          android:endX="97.3104"
+          android:type="linear">
+        <item android:offset="0" android:color="#BF3BA4DB"/>
+        <item android:offset="0.84375" android:color="#FFFFFFFF"/>
+      </gradient>
+    </aapt:attr>
+  </path>
+</vector>
```
frett: this is a massive vector image, is it something that would make more sense as a jpg or png?

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/animation/TextViewAnimation.kt
```diff
@@ -0,0 +1,22 @@
...
+            fadeInToNewText(animateNextText)
+        }
+    }).start()
+}
+
+private fun AppCompatTextView.fadeInToNewText(@StringRes newText: Int) {
+    alpha = 0f
+    setText(newText)
+    animate().alpha(1f).setDuration(3000).setListener(null).start()
+}
```
frett: I'm pretty sure using the animation APIs you can do a sequence of animations without needing to trigger 2 separate animations

frett: doing 2 separate animations like this makes it harder to maintain

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/adapter/OnBoardingPagerAdapter.kt
```diff
@@ -0,0 +1,50 @@
...
+import androidx.databinding.DataBindingUtil
+import androidx.databinding.ViewDataBinding
+import androidx.viewpager.widget.PagerAdapter
+import org.cru.godtools.tutorial.BR
+import org.cru.godtools.tutorial.R
+import org.cru.godtools.tutorial.animation.animateToNextText
+import org.cru.godtools.tutorial.databinding.BakedInOnboardingWelcomeBinding
+import org.cru.godtools.tutorial.util.OnBoardingCallbacks
+
+class OnBoardingPagerAdapter(val callbacks: OnBoardingCallbacks) : PagerAdapter() {
```
frett: I'm glad you have only 1 PagerAdapter. name it something more appropriate, such as `TutorialPagerAdapter`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/adapter/OnBoardingPagerAdapter.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    lateinit var onBoardingPagesLayout: List<Int>
+
+    override fun instantiateItem(container: ViewGroup, position: Int): Any {
+        val layoutID = onBoardingPagesLayout[position]
```
frett: there is no need to create a `layoutID` variable

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/adapter/OnBoardingPagerAdapter.kt
```diff
@@ -0,0 +1,50 @@
...
+    private fun ViewDataBinding.setDataBindingAnimation() {
+        when (this) {
+            is BakedInOnboardingWelcomeBinding -> {
+                welcomeTextView.postDelayed(
+                    { welcomeTextView.animateToNextText(R.string.baked_in_welcome_helping) },
+                    3000
+                )
+            }
+        }
+    }
```
frett: all of this animation can be cleaned up to be more straight forward. and based on what you are actually doing, `postDelayed` doesn't make sense, you should add an initial delay when you create the animation set that covers the entire animation

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/animation/TextViewAnimation.kt
```diff
@@ -0,0 +1,22 @@
...
+import androidx.annotation.StringRes
+import androidx.appcompat.widget.AppCompatTextView
+
+fun AppCompatTextView.animateToNextText(@StringRes animateNextText: Int) {
+    alpha = 1f
+    animate().setDuration(3000).alpha(0f).setListener(object : AnimatorListenerAdapter() {
+        override fun onAnimationEnd(animation: Animator?) {
+            super.onAnimationEnd(animation)
+            fadeInToNewText(animateNextText)
+        }
```
frett: a 3 second fade out is pretty slow, you will probably want it to be on the order of 500-1000 ms

frett: same thing with the fade in

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/adapter/OnBoardingPagerAdapter.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    override fun isViewFromObject(view: View, `object`: Any): Boolean = view == `object`
+
+    override fun getCount(): Int = onBoardingPagesLayout.size
+
+    private fun ViewDataBinding.setDataBindingAnimation() {
+        when (this) {
+            is BakedInOnboardingWelcomeBinding -> {
+                welcomeTextView.postDelayed(
```
frett: `(this as? BakedInOnboardingWelcomeBinding)?.welcomeTextView?.triggerWelcomeTextAnimation()`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/adapter/OnBoardingPagerAdapter.kt
```diff
@@ -0,0 +1,50 @@
...
+        return binding.root
+    }
+
+    override fun destroyItem(container: ViewGroup, position: Int, `object`: Any) {
+        container.removeView(`object` as View)
+    }
+
+    override fun isViewFromObject(view: View, `object`: Any): Boolean = view == `object`
+
+    override fun getCount(): Int = onBoardingPagesLayout.size
```
frett: `: Int` is unnecessary

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/BakedInOnBoardingActivity.kt
```diff
@@ -0,0 +1,89 @@
...
+            finish()
+        }
+    }
+
+    override fun onCloseClicked() {
+        finish()
+    }
+
+    override fun onOptInClicked() {
+        startActivity(Intent(this, OptInOnBoardingActivity::class.java))
```
frett: I don't like the pattern of creating an `Intent` manually anywhere you need one. it makes it hard to adjust any expected parameters over time. right now the best pattern I've come up with is either a `Context.startOnBoardingActivity()` or `Context.buildOnBoardingActivityIntent()` extension method

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/BakedInOnBoardingActivity.kt
```diff
@@ -0,0 +1,89 @@
...
+        Settings.getInstance(this).setFeatureDiscovered(Settings.FEATURE_OPT_IN_TUTORIAL)
```
frett: I would log the feature discovered in the activity itself when it starts

ui/tutorial-renderer/src/main/res/drawable/img_baked_in_others.xml
gyasistory: godtools____Documents_GitHub_godtools-android__-_____ui_tutorial-renderer_src_main_res_drawable_img_baked_in_others_xml__ui-tutorial-renderer_

This is the image let me know what you would like me to do?

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/animation/TextViewAnimation.kt
```diff
@@ -1,22 +1,40 @@
...
-    alpha = 0f
-    setText(newText)
-    animate().alpha(1f).setDuration(3000).setListener(null).start()
+        override fun onAnimationCancel(animation: Animator?) {}
+
+        override fun onAnimationStart(animation: Animator?) {}
+    })
 }
+
+private const val delayDuration: Long = 3000
```
wrandall22: Normally constants follow an upper underscore convention (e.g. `DELAY_DURATION`)

wrandall22: Also, could this just be `3000L` instead of `: Long = 3000`?

wrandall22: Third, isn't the convention to be at the top of the file?

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/animation/TextViewAnimation.kt
```diff
@@ -1,22 +1,40 @@
...
-import android.animation.AnimatorListenerAdapter
+import android.animation.AnimatorSet
+import android.animation.ObjectAnimator
 import androidx.annotation.StringRes
 import androidx.appcompat.widget.AppCompatTextView
 
 fun AppCompatTextView.animateToNextText(@StringRes animateNextText: Int) {
     alpha = 1f
-    animate().setDuration(3000).alpha(0f).setListener(object : AnimatorListenerAdapter() {
+    val fadOutAnimator = ObjectAnimator.ofFloat(this, "alpha", 1f, 0f).apply {
```
wrandall22: Typo here, should be `fadeOutAnimator`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
```diff
@@ -21,7 +21,7 @@ class TutorialActivity : AppCompatActivity(), TutorialCallbacks {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        setContentView(R.layout.activity_baked_in_onboarding)
+        setContentView(R.layout.activity_tutorial)
         viewPager = findViewById(R.id.baked_in_viewpager)
```
wrandall22: Is this view pager specific to the baked in state, or is it general for both states?

gyasistory: for both

wrandall22: If that's the case, we should probably rename it to reflect that fact.

ui/tutorial-renderer/build.gradle
```diff
@@ -0,0 +1,36 @@
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-kapt'
+
+android {
+    defaultConfig {
+        vectorDrawables.useSupportLibrary = true
+
+        javaCompileOptions {
+            annotationProcessorOptions {
+                arguments = [eventBusIndex: 'org.cru.godtools.tract.TractEventBusIndex']
```
frett: you aren't using eventbus, so this isn't necessary. plus it would cause a conflict with the tract renderer if you were using eventbus

ui/tutorial-renderer/build.gradle
```diff
@@ -0,0 +1,36 @@
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-kapt'
```
frett: I don't think you will need `kotlin-kapt` for this module since you aren't doing any annotation processing

gyasistory: I needed it for the dataBinding

app/build.gradle
frett: This still needs to be removed

frett: starting a new thread about this since the previous comment seems to have been lost.

this dependency should be unnecessary because you already define it earlier

app/src/main/AndroidManifest.xml
```diff
@@ -24,6 +25,8 @@
         android:name="android.hardware.location"
         android:required="false" />
 
+    <uses-sdk tools:overrideLibrary="com.pierfrancescosoffritti.androidyoutubeplayer"/>
```
frett: this line doesn't fit in this PR now since you are defining it in the tutorial module. But this will end up showing back up with the tool details UI, so it can stay

app/src/main/java/org/cru/godtools/fragment/ToolsFragment.java
```diff
@@ -117,19 +138,30 @@ public void onCreate(@Nullable final Bundle savedInstanceState) {
             mMode = args.getInt(EXTRA_MODE, mMode);
         }
 
+        mSettings = Settings.getInstance(requireContext());
```
frett: you should get the Settings singleton in the `onAttach` method instead.

app/src/main/java/org/cru/godtools/fragment/ToolsFragment.java
```diff
@@ -117,19 +138,30 @@ public void onCreate(@Nullable final Bundle savedInstanceState) {
...
         super.onViewCreated(view, savedInstanceState);
         setupToolsList();
+        mToolsBinding.setCallback(mTutorialCallbacks);
+    }
+
+    @Override
+    public void onResume() {
+        super.onResume();
+        mToolsBinding.setIsTutorialViewable(
+                !mSettings.isFeatureDiscovered(Settings.FEATURE_OPT_IN_TUTORIAL));
```
frett: you should be updating the visibility of the tool when the preference changes, not in `onResume()`

app/src/main/java/org/cru/godtools/fragment/ToolsFragment.java
```diff
@@ -210,6 +242,17 @@ private boolean showDownloading() {
         return mMode == MODE_AVAILABLE;
     }
 
+    private void closeTutorial() {
+        mSettings.setFeatureDiscovered(Settings.FEATURE_OPT_IN_TUTORIAL);
+        mToolsBinding.setIsTutorialViewable(false);
```
frett: this should be unnecessary once you correctly monitor for changes to the feature flag

app/src/main/java/org/cru/godtools/fragment/ToolsFragment.java
```diff
@@ -226,9 +269,9 @@ private void restartToolsLoader() {
 
     @SuppressWarnings("unchecked")
     private void setupToolsList() {
-        if (mToolsView != null) {
-            mToolsView.setLayoutManager(new LinearLayoutManager(requireActivity()));
-            mToolsView.setHasFixedSize(false);
+        if (mToolsBinding.resources != null) {
+            mToolsBinding.resources.setLayoutManager(new LinearLayoutManager(requireActivity()));
```
frett: the layout manager should be defined in the XML

app/src/main/java/org/cru/godtools/fragment/ToolsFragment.java
```diff
@@ -117,19 +138,30 @@ public void onCreate(@Nullable final Bundle savedInstanceState) {
...
+
         startLoaders();
     }
 
     @Override
     public View onCreateView(@NonNull final LayoutInflater inflater, @NonNull final ViewGroup container,
                              @Nullable final Bundle savedInstanceState) {
-        return inflater.inflate(R.layout.fragment_tools, container, false);
+        mToolsBinding = FragmentToolsBinding.inflate(inflater, container, false);
+        return mToolsBinding.getRoot();
```
frett: now that you are using data binding you need to clean up and unset the data binding in `onViewDestroyed()` to prevent it from leaking views

ui/tutorial-renderer/src/main/res/drawable/ic_close_tutorial.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportHeight="24.0"
+    android:viewportWidth="24.0">
+    <path
+        android:fillColor="@color/tintable"
+        android:pathData="M19,6.41L17.59,5 12,10.59 6.41,5 5,6.41 10.59,12 5,17.59 6.41,19 12,13.41 17.59,19 19,17.59 13.41,12z" />
+</vector>
```
frett: did you copy this drawable from elsewhere? instead you could just move the `ic_close` drawable to the `ui-base` module so it's available in both places

app/src/main/res/layout/fragment_tools.xml
```diff
@@ -1,14 +1,96 @@
...
+                app:srcCompat="@drawable/ic_close"
+                app:tint="@color/design_default_color_primary_dark" />
+
+            <androidx.appcompat.widget.AppCompatTextView
+                android:id="@+id/tutorialQuestionTextView"
+                android:layout_width="0dp"
+                android:layout_height="wrap_content"
+                android:layout_marginTop="30dp"
+                android:gravity="center"
+                android:text="Not sure where to start?"
```
frett: this needs to be translatable

app/src/main/res/layout/fragment_tools.xml
```diff
@@ -1,14 +1,96 @@
...
+            <TextView
+                android:id="@+id/openTutorialTextView"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:onClick="@{() -> callback.onTutorialOpen()}"
+                android:paddingTop="10dp"
+                android:paddingBottom="20dp"
+                app:drawableEndCompat="@drawable/ic_nav_next_arrow"
+                android:gravity="center|center_horizontal|center_vertical"
+                android:text="Open Tutorial"
```
frett: This needs to be translatable

ui/tutorial-renderer/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+    }
+    dataBinding {
+        enabled = true
+    }
+}
+
+dependencies {
+    implementation project(':library:base')
+
+    implementation project(':ui:base-tool')
```
frett: `base-tool` is probably the wrong module to depend on here. `base-tool` has the framework for rendering tools being returned from the API.

frett:
ugg, the up navigation button doesn't operate how I would want it to operate....

app/src/main/java/org/cru/godtools/fragment/BasePlatformFragment.kt
```diff
@@ -135,6 +138,9 @@ abstract class BasePlatformFragment : BaseFragment() {
             when (key) {
                 PREF_PRIMARY_LANGUAGE, PREF_PARALLEL_LANGUAGE -> loadLanguages(false)
             }
+            if (key?.startsWith(PREF_FEATURE_DISCOVERED) == true) {
```
wrandall22: Do we really have to specify `== true`?

wrandall22:
There is a mix of `on_boarding` and `onboarding` IDs in the strings. We should probably normalize.

wrandall22:
@frett when you renamed `BAKED_IN` to `ONBOARDING` and `OPT_IN` to `TRAINING` it seems like there were other things that maybe should have been renamed in the process such as strings IDs and `Settings.FEATURE_BAKED_IN_TUTORIAL`, etc.

app/src/main/java/org/cru/godtools/fragment/BasePlatformFragment.kt
frett: yes because `key?.startsWith(PREF_FEATURE_DISCOVERED)` could be null. this is actually the recommended kotlin idiom when dealing with a nullable boolean

frett: https://kotlinlang.org/docs/reference/idioms.html#consuming-a-nullable-boolean

frett:
@wrandall22 yup, was doing it in smaller self-contained changes instead of doing all the renaming at once :)

app/src/main/java/org/cru/godtools/fragment/BasePlatformFragment.kt
wrandall22: Ah yes, nullable.

frett: matt asked a similar question on a different PR 2 days ago :)

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
```diff
@@ -52,7 +52,7 @@ class TutorialActivity : AppCompatActivity(), TutorialCallbacks {
     }
 
     private fun ViewPager.setupIndicator() {
-        findViewById<CircleIndicator>(R.id.on_boarding_indicator)?.let { indicator ->
+        this@TutorialActivity.findViewById<CircleIndicator>(R.id.on_boarding_indicator)?.let { indicator ->
```
frett: good catch

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
```diff
@@ -57,17 +82,47 @@ class TutorialActivity : AppCompatActivity(), TutorialCallbacks {
...
+    private fun setMenuVisibility(isToolbarVisible: Boolean) {
+        var i = 0
+        val size = tutorialMenu?.size() ?: 0
+        while (i < size) {
+            tutorialMenu?.getItem(i)?.isVisible = isToolbarVisible
+            i++
+        }
+    }
+
+    private fun setHomLinkVisibility(isHomeLinkVisible: Boolean) {
```
wrandall22: Should be `setHomeLinkVisibility`

app/src/main/java/org/cru/godtools/fragment/BasePlatformFragment.kt
```diff
@@ -135,6 +136,9 @@ abstract class BasePlatformFragment : BaseFragment() {
...
+                onUpdateFeatureDiscovery(key.removePrefix(PREF_FEATURE_DISCOVERED))
```
frett: @gyasistory handy function I just found

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
```diff
@@ -0,0 +1,156 @@
...
+    override fun onStart() {
+        super.onStart()
+        pageSet.feature?.let { Settings.getInstance(this).setFeatureDiscovered(it) }
+    }
+
+    private var tutorialMenu: Menu? = null
+
+    override fun onCreateOptionsMenu(menu: Menu?): Boolean {
+        if (pageSet == PageSet.ONBOARDING) {
+            menuInflater.inflate(R.menu.tutorial_menu, menu)
```
frett: instead of hardcoding a specific menu here, you should add a property to the `PageSet` that defines the menu res to inflate. And then you only inflate a menu if one is defined for the pageSet

frett: it simplifies some of your menu customization, unfortantly you will still need to manually hide the menu for the first page of the onboarding tutorial

ui/tutorial-renderer/src/main/res/layout/baked_in_tutorial_tools.xml
```diff
@@ -70,7 +70,11 @@
             android:id="@+id/onboarding_next_button"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
-            android:layout_margin="32dp"
+            android:layout_marginStart="32dp"
+            android:layout_marginLeft="32dp"
+            android:layout_marginEnd="32dp"
+            android:layout_marginRight="32dp"
```
frett: in the situation that the left/right margin is identical there is no need to use both left/right and start/end. it makes layouts easier to digest if you leave off one of those sets when not necessary.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
```diff
@@ -47,13 +47,13 @@ class TutorialActivity : AppCompatActivity(), TutorialCallbacks {
...
     override fun onCreateOptionsMenu(menu: Menu?): Boolean {
-        if (pageSet == PageSet.ONBOARDING) {
-            menuInflater.inflate(R.menu.tutorial_menu, menu)
+        pageSet.menu?.let {
+            menuInflater.inflate(it, menu)
             tutorialMenu = menu
             setMenuVisibility(false)
-        } else {
-            setHomeLinkVisibility(true)
+            return super.onCreateOptionsMenu(menu)
```
gyasistory: @wrandall22 `setHomeLinkVisibility(true)` is only being called if there is no menu in the enum.  The `return` will end the call.

wrandall22: Yeah, sorry about that, didn't see the return statement at first.

frett:
the up navigation in the tutorial activity doesn't work

ui/tutorial-renderer/src/main/res/drawable/img_tutorial_selected.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
frett: let's convert these circles to be shape drawables

ui/tutorial-renderer/src/main/res/drawable/img_tutorial_unselected.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
frett: convert this circle to a simple shape drawable

ui/tutorial-renderer/src/main/res/drawable/img_tutorial_selected.xml
```diff
@@ -0,0 +1,10 @@
...
+    android:width="8dp"
+    android:height="8dp"
+    android:viewportWidth="8"
+    android:viewportHeight="8">
+  <path
+      android:pathData="M4,8.00024C6.20914,8.00024 8,6.20938 8,4.00024C8,1.7911 6.20914,0.00024 4,0.00024C1.79086,0.00024 0,1.79111 0,4.00024C0,6.20938 1.79086,8.00024 4,8.00024Z"
+      android:fillColor="#3BA4DB"
```
frett: this color should be a resource and not hard-coded

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/animation/TextViewAnimation.kt
```diff
@@ -0,0 +1,40 @@
...
+    }
+    AnimatorSet().apply {
+        startDelay = DELAY_DURATION
+        play(fadeOutAnimator).before(fadeInAnimator)
+        start()
+    }
+}
+
+private fun ObjectAnimator.setTextAfterAnimation(@StringRes animateNextText: Int, textView: AppCompatTextView) {
+    addListener(object : Animator.AnimatorListener {
```
frett: since you are only implementing a single method you should extend `AnimatorListenerAdapter` instead to simplify this code

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
```diff
@@ -0,0 +1,156 @@
...
+        supportActionBar?.setDisplayShowTitleEnabled(false)
+    }
+
+    private fun setMenuVisibility(isToolbarVisible: Boolean) {
+        var i = 0
+        val size = tutorialMenu?.size() ?: 0
+        while (i < size) {
+            tutorialMenu?.getItem(i)?.isVisible = isToolbarVisible
+            i++
+        }
```
frett: there should be a much simpler way to do this in Kotlin

frett: `implementation "androidx.core:core-ktx:${deps.androidX.core}"`
`tutorialMenu?.forEach { it.isVisible = isToolbarVisible }`

frett:
the up navigation icon should be the close X instead of the up arrow

ui/tutorial-renderer/src/main/res/values/tutorial_style.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <style name="Theme.GodTools.AppBar.Tutorial" parent="Theme.AppCompat.Light">
+        <item name="android:background">@android:color/transparent</item>
+        <item name="android:tint">@color/gt_blue</item>
+        <item name="actionMenuTextColor">@color/gt_blue</item>
+    </style>
```
frett: you are combining a style and a theme together which are 2 distinct things

gyasistory: @frett it looks like you took care of this?

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
gyasistory: @frett it looks like you did this one as well?

app/src/main/java/org/cru/godtools/adapter/ToolsHeaderAdapter.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    private lateinit var adapterToolHeader: ToolHeader
+
+    var isHeaderVisible: Boolean = true
+        set(value) {
+            field = value
+            headerAdapter?.notifyDataSetChanged()
+        }
+
+    lateinit var headerBannerCallbacks: HeaderBannerCallbacks
```
frett: you are overusing `lateinit`, this shouldn't be a lateinit var

app/src/main/java/org/cru/godtools/adapter/ToolsHeaderAdapter.kt
```diff
@@ -0,0 +1,61 @@
...
+            this.toolHeader = header
+            return this
+        }
+
+        lateinit var toolHeader: ToolHeader
+    }
+
+    // region Header
+
+    private lateinit var adapterToolHeader: ToolHeader
```
frett: again this shouldn't be lateinit, it should be nullable and visibility is controlled by if there is a banner defined or not.

also, this shouldn't be private, and you don't need to use a builder to set it

app/src/main/java/org/cru/godtools/adapter/ToolsHeaderAdapter.kt
```diff
@@ -0,0 +1,61 @@
...
+                this.adapterToolHeader = toolHeader
+            }
+        }
+
+        fun toolHeader(header: ToolHeader): Builder {
+            this.toolHeader = header
+            return this
+        }
+
+        lateinit var toolHeader: ToolHeader
```
frett: I actually don't care for the builder pattern here. you shouldn't be setting what banner to display in the builder/constructor because it should dynamically change after the adapter was created.

app/src/main/java/org/cru/godtools/adapter/ToolsHeaderAdapter.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    // region Header
+
+    private lateinit var adapterToolHeader: ToolHeader
+
+    var isHeaderVisible: Boolean = true
+        set(value) {
+            field = value
+            headerAdapter?.notifyDataSetChanged()
+        }
```
frett: this is unnecessary. you should determine banner visibility based on if there is a banner defined or the banner is null

app/src/main/java/org/cru/godtools/adapter/ToolsHeaderAdapter.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    lateinit var headerBannerCallbacks: HeaderBannerCallbacks
+
+    override fun getHeaderItemCount(): Int {
```
frett: you are missing implementations of `getHeaderItemViewType()` and `getHeaderItemId()`

app/src/main/java/org/cru/godtools/adapter/ToolsHeaderAdapter.kt
```diff
@@ -0,0 +1,61 @@
...
+        return if (isHeaderVisible) 1 else 0
+    }
+
+    override fun onBindHeaderItemViewHolder(holder: RecyclerView.ViewHolder, localPosition: Int) =
+        (holder as DataBindingViewHolder<ToolsTutorialBannerBinding>).let {
+            it.binding.callback = headerBannerCallbacks
```
frett: the callback should be set when the DataBinding is inflated

app/src/main/java/org/cru/godtools/content/ToolHeader.kt
```diff
@@ -0,0 +1,12 @@
+package org.cru.godtools.content
+
+import androidx.annotation.StringRes
+import org.cru.godtools.R
+
+enum class ToolHeader(
+    @StringRes val descriptionText: Int,
+    @StringRes val openText: Int,
+    @StringRes val dismissText: Int
```
frett: I don't want to follow the same pattern as the empty list UI follows. I would have fully self-contained layouts for each potential banner we display.

This enum should then just reference which Banner layout to inflate

app/src/main/res/layout/tools_tutorial_banner.xml
```diff
@@ -0,0 +1,78 @@
...
+        android:layout_marginLeft="@dimen/tools_card_margin_horiz"
+        android:layout_marginRight="@dimen/tools_card_margin_horiz">
+
+        <androidx.appcompat.widget.AppCompatTextView
+            android:id="@+id/banner_open_description_text"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_marginLeft="16dp"
+            android:layout_marginRight="16dp"
+            android:firstBaselineToTopHeight="36dp"
```
frett: this attribute was added in API 28. it will be a long time until we can utilize it :(

app/src/main/java/org/cru/godtools/adapter/ToolsHeaderAdapter.kt
gyasistory: @frett it looks like you completed most of these changes, am I correct?

frett: yeah, pending feedback from justin on the banner we are done

PR #760, opened 2019-11-20
GT-601 : Requested Changes from QA

gyasistory:
https://jira.cru.org/browse/GT-601

Two issues: 

1. Can you increase the buffer space between the footer of the card and the text on the card? The words Previous and Next are sometimes too close to the text on the card and I think it might be confusing. Is this easy to do?

2. Is there any way on the prayer screens to say 1/2 cards? Right now, it only shows that there is one card on that screen and I think a lot of people will miss the prayer screen sign up for follow up

### DIFF
--- a/ui/tract-renderer/src/main/java/org/cru/godtools/tract/viewmodel/CardViewHolder.java
+++ b/ui/tract-renderer/src/main/java/org/cru/godtools/tract/viewmodel/CardViewHolder.java
@@ -191,7 +191,7 @@ public final class CardViewHolder extends ParentViewHolder<Card> {
         int count = 0;
         if (mModel != null) {
             for (Card card : mModel.getPage().getCards()) {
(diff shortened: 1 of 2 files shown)


ui/tract-renderer/src/main/java/org/cru/godtools/tract/viewmodel/CardViewHolder.java
```diff
@@ -176,22 +176,11 @@ private boolean isPrayerForm(Card card) {
...
-                }
-            }
-        }
-        return false;
-    }
-
     private int getCardCount() {
         int count = 0;
         if (mModel != null) {
             for (Card card : mModel.getPage().getCards()) {
```
frett: shouldn't the card count be the total number of visible cards instead of filtering out the arbitrary `PrayerForm` card?

visible cards is a much less brittle number to calculate

PR #763, opened 2019-11-21
GT-641 : jira requested changes

gyasistory:
https://jira.cru.org/browse/GT-641

God_Tools_-_Agile_Board_-_Digital_Products_and_Strategies_Projects_for_Cru

### DIFF
--- a/app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
+++ b/app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
@@ -22,7 +22,6 @@ import org.ccci.gto.android.common.support.v4.app.SimpleLoaderCallbacks;
 import org.ccci.gto.android.common.support.v4.util.FragmentUtils;
 import org.ccci.gto.android.common.viewpager.view.ChildHeightAwareViewPager;
 import org.cru.godtools.R;
-import org.cru.godtools.base.Settings;
 import org.cru.godtools.base.ui.util.ModelUtils;
 import org.cru.godtools.base.util.LocaleUtils;
 import org.cru.godtools.content.AttachmentLoader;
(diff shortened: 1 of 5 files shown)


frett:
you should leverage the latest translation loader to load the latest primary and parallel language translation. the primary one is already configured in ToolDetailsFragment, you will just need to create a parallel language latest language loader. you can update the `TranslationLoaderCallbacks` inner class to support both the primary and parallel language loaders and don't need to duplicate that code.

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -470,11 +493,16 @@ public void onLoadFinished(@NonNull final Loader<Attachment> loader, @Nullable f
...
-                case LOADER_LATEST_TRANSLATION:
+                case LOADER_LATEST_PRIMARY_TRANSLATION:
                     if (mToolCode != null) {
                         return new LatestTranslationLoader(requireContext(), mToolCode, getPrimaryLanguage());
                     }
                     break;
+                case LOADER_LATEST_PARALLEL_TRANSLATION:
+                    if (mToolCode != null && getParallelLanguage() != null) {
+                        return new LatestTranslationLoader(requireContext(), mToolCode, getParallelLanguage());
+                    }
```
frett: you should always create the loader even if the parallel language isn't currently set (I believe)

frett: hmm, you can't at the moment because locale is marked as needing to be non-null

app/src/main/java/org/cru/godtools/fragment/ToolDetailsFragment.java
```diff
@@ -204,13 +207,21 @@ public boolean onOptionsItemSelected(final MenuItem item) {
...
         stopProgressListener();
         startProgressListener();
     }
 
+    @Override
+    protected void onUpdateParallelLanguage() {
+        updateLatestParallelTranslationLoader();
+
+        stopProgressListener();
+        startProgressListener();
```
frett: we don't need to restart the progress listener on changing the parallel language. the progress listener only monitors the primary language download

PR #785, opened 2019-12-17
GT-631 : Add more space between Tab and text

gyasistory:
https://jira.cru.org/browse/GT-631


wrandall22:
What's the difference between padding and margin as handled by Android? Is it the same as how CSS handles it?

frett:
yeah, I believe it is the same.

padding makes the view dimensions larger by filling the sides with empty space. (backgrounds, touch area, and similar things extend into the padding)
margin just offsets layout of the view

frett:
so, this PR effectively makes the view `16dp` larger instead of just shifting the view down `16dp`

PR #803, opened 2020-01-03
GT-660 : update menu to include account section

gyasistory:


app/src/main/res/menu/activity_generic_left_menu.xml
```diff
@@ -24,19 +24,25 @@
                 android:id="@+id/action_contact_us"
                 android:orderInCategory="400"
                 android:title="@string/menu_contact_us" />
+        </menu>
+    </item>
 
+    <item
+        android:orderInCategory="100"
```
frett: if we are actually utilizing `orderInCategory`, then it shouldn't have the same value as other items at the same level

frett: https://github.com/CruGlobal/godtools-android/pull/803/files#diff-c09f752f83501a05f88e2ed5ca34a244R4

PR #807, opened 2020-01-03
GT-634 : Include Global Actvity

gyasistory:


app/src/main/java/org/cru/godtools/activity/GlobalDashboardActivity.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        setContentView(R.layout.activity_generic_fragment_with_nav_drawer)
+    }
+
+    override fun onSetupActionBar() {
+        super.onSetupActionBar()
+        title = resources.getString(R.string.title_my_profile)
```
frett: you shouldn't need to set the title here

app/src/main/java/org/cru/godtools/activity/GlobalDashboardActivity.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    override fun onStart() {
+        super.onStart()
+    }
+
+    private fun loadInitialFragment() {
+        supportFragmentManager?.apply {
+           beginTransaction().replace(R.id.frame, GlobalDashboardFragment(), TAG_GLOBAL_DASHBOARD_FRAGMENT)
+        }
```
frett:
```kotlin
supportFragmentManager.commit {
    replace(R.id.frame, GlobalDashboardFragment(), TAG_GLOBAL_DASHBOARD_FRAGMENT)
}
```

app/src/main/java/org/cru/godtools/fragment/GlobalDashboardFragment.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import org.cru.godtools.databinding.FragmentGlobalDashboardBinding
+
+class GlobalDashboardFragment : BasePlatformFragment() {
+
+    private lateinit var binding: FragmentGlobalDashboardBinding
```
frett: `lateinit` for the binding in a fragment is bad because it will leak the view hierarchy after `onDestroyView()`.

I know we utilize it in MPDX, it's something that will need to change there as well, but is more involved because of how much we are currently using it

app/src/main/res/values/strings.xml
```diff
@@ -54,6 +54,7 @@
     <string name="feature_discovery_desc_language_settings" translatable="false">@string/text_language_parallel_description</string>
     <string name="menu_language_settings">Language Settings</string>
     <string name="title_language_settings">Language Settings</string>
+    <string name="title_my_profile">My Profile</string>
```
frett: this string isn't part of the "Language Selection UI"

app/src/main/java/org/cru/godtools/activity/GlobalDashboardActivity.kt
```diff
@@ -0,0 +1,38 @@
...
+import android.content.Intent
+import android.os.Bundle
+import org.cru.godtools.R
+import org.cru.godtools.fragment.GlobalDashboardFragment
+
+private const val TAG_GLOBAL_DASHBOARD_FRAGMENT = "globalDashboardFragment"
+
+fun Activity.startGlobalDashboardActvity() {
+    Intent(this, GlobalDashboardActivity::class.java)
+        .also { startActivity(it) }
```
frett: let's simplify this to: `startActivity(Intent(this, GlobalDashboardActivity::class.java))`

frett:
we need a strategy for the database, I don't want to just put this in Room without having a plan forward to migrate the other database

PR #816, opened 2020-01-13
GT-664   english only tutorials 

gyasistory:
https://jira.cru.org/browse/GT-664

### DIFF
--- a/app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.kt
+++ b/app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.kt
@@ -80,6 +80,7 @@ abstract class BasePlatformActivity : BaseDesignActivity(), NavigationView.OnNav
     private var loginItem: MenuItem? = null
     private var signupItem: MenuItem? = null
     private var logoutItem: MenuItem? = null
+    private var tutorialItem: MenuItem? = null
 
     private var primaryLanguage = Settings.getDefaultLanguage()
     private var parallelLanguage: Locale? = null
@@ -250,6 +251,7 @@ abstract class BasePlatformActivity : BaseDesignActivity(), NavigationView.OnNav
                 loginItem = findItem(R.id.action_login)
                 signupItem = findItem(R.id.action_signup)
                 logoutItem = findItem(R.id.action_logout)
+                tutorialItem = findItem(R.id.action_tutorial)
             }
             updateNavigationDrawerMenu()
         }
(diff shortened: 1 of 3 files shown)


library/base/src/main/java/org/cru/godtools/base/Settings.java
```diff
@@ -65,7 +65,8 @@ public boolean isFeatureDiscovered(@NonNull final String feature) {
             boolean changed = false;
             switch (feature) {
                 case FEATURE_TUTORIAL_ONBOARDING:
-                    if (getFirstLaunchVersion() <= VERSION_5_1_4) {
+                    if (getFirstLaunchVersion() <= VERSION_5_1_4 ||
+                            Locale.getDefault().getLanguage().startsWith("en")) {
```
frett: this test is backwards, it should be `!Locale.getDefault().getLanguage().startsWith("en")`

gyasistory: Great Catch I missed it in the change.

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.kt
```diff
@@ -250,6 +251,7 @@ abstract class BasePlatformActivity : BaseDesignActivity(), NavigationView.OnNav
                 loginItem = findItem(R.id.action_login)
                 signupItem = findItem(R.id.action_signup)
                 logoutItem = findItem(R.id.action_logout)
+                tutorialItem = findItem(R.id.action_tutorial)
```
frett:
```suggestion

                findItem(R.id.action_tutorial)?.isVisible = Locale.getDefault().language.startsWith("en")
```

frett: visibility can be set when creating the menu because the default language won't change without restarting the activity

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.kt
```diff
@@ -80,6 +80,7 @@ abstract class BasePlatformActivity : BaseDesignActivity(), NavigationView.OnNav
     private var loginItem: MenuItem? = null
     private var signupItem: MenuItem? = null
     private var logoutItem: MenuItem? = null
+    private var tutorialItem: MenuItem? = null
```
frett: not necessary if we just set the visibility inline when setting up the nav drawer

app/src/main/java/org/cru/godtools/activity/BasePlatformActivity.kt
```diff
@@ -260,6 +262,7 @@ abstract class BasePlatformActivity : BaseDesignActivity(), NavigationView.OnNav
         loginItem?.isVisible = guid == null
         signupItem?.isVisible = guid == null
         logoutItem?.isVisible = guid != null
+        tutorialItem?.isVisible = Locale.getDefault().language.startsWith("en")
```
frett: only needs to be set when we initially load the menu

PR #817, opened 2020-01-13
implement new LiveDataDao

frett:


library/db/src/main/java/org/keynote/godtools/android/db/GodToolsDao.java
```diff
@@ -54,6 +64,29 @@ private GodToolsDao(@NonNull final Context context) {
                      LocalFileTable.SQL_WHERE_PRIMARY_KEY);
         registerType(TranslationFile.class, TranslationFileTable.TABLE_NAME, TranslationFileTable.PROJECTION_ALL,
                      new TranslationFileMapper(), TranslationFileTable.SQL_WHERE_PRIMARY_KEY);
+
+        // TODO: this may be temporary until a better solution is available
+        EventBus.getDefault().register(new Object() {
+            @Subscribe(threadMode = ThreadMode.MAIN)
```
gyasistory: @frett can you explain why you need to to this.

frett: we need a way to know when data has changed in the database table. the quick solution was to tie into the EventBus broadcasts that we currently use to indicate that data has changed.

Long term I want to look into directly invalidating for classes that we update through the DAO, but there is some complications between triggering this taking transactions into account

PR #818, opened 2020-01-13
GT-661   add analytics to onboarding and tutorial

gyasistory:


gyasistory:
@frett I have no idea what is causing this not to compile.  From what I can tell everything looks good.

frett:
@gyasistory do you have a reference to what the data being sent to Adobe should look like?

frett:
Also, are you looking to send the `onboarding_start` event for both the onboarding and training tutorials?

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/TutorialAnalyticsActionEvent.kt
```diff
@@ -0,0 +1,16 @@
...
+internal const val ACTION_TUTORIAL_ONBOARDING_TRAINING = "onboarding_more"
+internal const val ACTION_TUTORIAL_FINISH = "onboarding_start"
+
+class TutorialAnalyticsActionEvent(action: String) : AnalyticsActionEvent(null, action) {
+    override fun isForSystem(system: AnalyticsSystem): Boolean {
+        return system == AnalyticsSystem.ADOBE || system == AnalyticsSystem.FACEBOOK
+    }
+
+    override val adobeSiteSection get() = ADOBE_SITE_SECTION_TUTORIAL
+    override fun getAttributes() = mapOf(action to 1)
```
frett: I'm mostly curious about what the action attributes should look like, this is the first instance where we use the same string for the action and the attribute. So either they have changed how they are doing it or you are doing something unnecessry or incorrect

frett:
@gyasistory do you have an example of what you are trying to send to the adobe analytics API for events?

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
```diff
@@ -67,13 +67,21 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
     }
 
     override fun launchTraining() {
-        EventBus.getDefault().post(TutorialAnalyticsActionEvent(ACTION_TUTORIAL_ONBOARDING_TRAINING))
+        if (shouldSendOnBoardingEvent()) {
+            EventBus.getDefault().post(TutorialAnalyticsActionEvent(ACTION_TUTORIAL_ONBOARDING_TRAINING))
+        }
```
frett: this is a really hacky way to accomplish this

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/TutorialAnalyticsActionEvent.kt
```diff
@@ -0,0 +1,18 @@
...
+internal const val ACTION_TUTORIAL_FINISH = "onboarding_start"
+internal const val ACTION_NAME_TUTORIAL_FINISH = "On-Boarding Start"
+
+class TutorialAnalyticsActionEvent(private val actionName: String, private val actionTitle: String) : AnalyticsActionEvent(null, actionTitle) {
+    override fun isForSystem(system: AnalyticsSystem): Boolean {
+        return system == AnalyticsSystem.ADOBE || system == AnalyticsSystem.FACEBOOK
+    }
+
+    override val adobeSiteSection get() = ADOBE_SITE_SECTION_TUTORIAL
+    override val adobeAttributes: Map<String?, Any> = mapOf(actionName to 1)
```
frett: did you follow up with Lauren about this? currently for the tutorial analytics you are sending the variable `onboarding_start`, all other variables we have implemented for ADOBE have been prefixed with `cru.`

gyasistory:
> Just the one outstanding question about confirming with Lauren about the attribute name being different than the rest of the attributes we send to adobe

Lauren didn't write the cru prefix in the documentation but I will double-check with her.

frett:
This is approved pending Lauren's response

PR #826, opened 2020-01-16
GT-631 : Text being cut off

gyasistory:
https://jira.cru.org/browse/GT-631

### DIFF
--- a/app/src/main/res/layout/fragment_tool_details.xml
+++ b/app/src/main/res/layout/fragment_tool_details.xml
@@ -101,19 +101,7 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:elevation="7dp"
-            app:tabBackground="@color/action_bar_action_tint">
-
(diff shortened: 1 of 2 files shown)


frett:
This isn't even technically correct, it still is cropping off a couple languages

app/src/main/res/layout/fragment_tool_details.xml
```diff
@@ -96,32 +96,31 @@
...
+                tools:text="@string/label_tools_about" />
 
-                <com.google.android.material.tabs.TabItem
-                    android:layout_width="wrap_content"
-                    android:layout_height="wrap_content"
-                    tools:text="@string/label_tools_about" />
+            <com.google.android.material.tabs.TabItem
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                tools:text="40 Languages" />
```
frett: I don't think you actually need to have `TabItems` defined here. they should automatically be created when you link the TabLayout to the ViewPager

PR #835, opened 2020-01-23
Gt 634   database configuration

gyasistory:
https://jira.cru.org/browse/GT-634

The first part of the Jira Task


library/api/src/main/java/org/cru/godtools/api/GlobalActivityAnalyticsApi.java
```diff
@@ -0,0 +1,16 @@
...
+import org.cru.godtools.model.GlobalActivityAnalytics;
+
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Streaming;
+
+public interface GlobalActivityAnalyticsApi {
+    String PATH_ANALYTICS_GLOBAL = "analytics/global";
+
+    @Streaming
```
frett: why are you using `@Streaming`?

library/db/src/main/java/org/keynote/godtools/android/db/Contract.java
```diff
@@ -314,4 +315,32 @@
...
+        private static final String SQL_COLUMN_LAUNCHES = COLUMN_LAUNCHES + " INTEGER";
+        private static final String SQL_COLUMN_GOSPEL_PRESENTATIONS = COLUMN_GOSPEL_PRESENTATIONS + " INTEGER";
+
+        static final Expression SQL_WHERE_PRIMARY_KEY = FIELD_ID.eq(bind());
+
+        static final String SQL_CREATE_TABLE =
+                create(TABLE_NAME, SQL_COLUMN_ID, SQL_COLUMN_USERS, SQL_COLUMN_COUNTRIES, SQL_COLUMN_LAUNCHES,
+                       SQL_COLUMN_GOSPEL_PRESENTATIONS);
+        static final String SQL_DELETE_TABLE = drop(TABLE_NAME);
+        static final String SQL_V41_CREATE_GLOBAL_ANALYTICS = SQL_CREATE_TABLE;
```
frett: What are some potential problems with how this constant is currently implemented?

frett: I'm using this as an opportunity to help develop some long term thinking about the code you write

library/db/src/main/java/org/keynote/godtools/android/db/GlobalActivityAnalyticsMapper.java
```diff
@@ -0,0 +1,57 @@
...
+import org.cru.godtools.model.GlobalActivityAnalytics;
+
+import androidx.annotation.NonNull;
+
+import static org.keynote.godtools.android.db.Contract.GlobalActivityAnalyticsTable.COLUMN_COUNTRIES;
+import static org.keynote.godtools.android.db.Contract.GlobalActivityAnalyticsTable.COLUMN_GOSPEL_PRESENTATIONS;
+import static org.keynote.godtools.android.db.Contract.GlobalActivityAnalyticsTable.COLUMN_LAUNCHES;
+import static org.keynote.godtools.android.db.Contract.GlobalActivityAnalyticsTable.COLUMN_USERS;
+
+public class GlobalActivityAnalyticsMapper extends BaseMapper<GlobalActivityAnalytics> {
```
frett: Implement this in Kotlin, it will be much simpler. There should be a few Mappers already converted to Kotlin if you need examples

library/db/src/main/java/org/keynote/godtools/android/db/GodToolsDatabase.java
```diff
@@ -98,6 +100,8 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
                     case 40:
                         db.execSQL(ToolTable.SQL_V40_ALTER_OVERVIEW_VIDEO);
                         break;
+                    case 41:
+                        db.execSQL(GlobalActivityAnalyticsTable.SQL_V41_CREATE_GLOBAL_ANALYTICS);
```
frett: this will cause a crash when a user upgrades from v40 to v41, why?

frett: this is separate from my other comment about long term issues with the `SQL_V41_CREATE_GLOBAL_ANALYTICS` constant

library/model/src/main/java/org/cru/godtools/model/GlobalActivityAnalytics.java
```diff
@@ -0,0 +1,59 @@
+package org.cru.godtools.model;
+
+import org.ccci.gto.android.common.jsonapi.annotation.JsonApiAttribute;
+import org.ccci.gto.android.common.jsonapi.annotation.JsonApiType;
+
+import static org.cru.godtools.model.GlobalActivityAnalytics.JSON_API_TYPE;
+
+@JsonApiType(JSON_API_TYPE)
+public class GlobalActivityAnalytics extends Base {
```
frett: make this a Kotlin class, it will be much simpler

library/db/src/main/java/org/keynote/godtools/android/db/GodToolsDatabase.java
gyasistory: I forgot the break right?

frett: yeah

library/db/src/main/java/org/keynote/godtools/android/db/Contract.java
gyasistory: Would it be better to make it equal to the create method, to separate it from the `SQL_CREATE_TABLE`.

PR #838, opened 2020-01-27
GT-634 : Create UI element for global dashboard

gyasistory:
This PR is only the UI implementation of GT-634


app/src/main/java/org/cru/godtools/activity/GlobalDashboardActivity.kt
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    override fun onResume() {
+        super.onResume()
+        mEventBus.post(AnalyticsScreenEvent(SCREEN_GLOBAL_DASHBOARD))
+    }
+
+    override fun onSetupActionBar() {
+        super.onSetupActionBar()
+        setTitle(R.string.gt_gd_my_profile_text)
```
frett: you should be setting this in the `AndroidManifest.xml`

app/src/main/java/org/cru/godtools/activity/GlobalDashboardActivity.kt
```diff
@@ -0,0 +1,44 @@
...
+class GlobalDashboardActivity : BasePlatformActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_generic_fragment)
+    }
+
+    override fun onStart() {
+        super.onStart()
+        loadGlobalDashboardFragment()
```
frett: can be triggered in `onCreate()`

app/src/main/java/org/cru/godtools/activity/GlobalDashboardActivity.kt
```diff
@@ -0,0 +1,44 @@
...
+        mEventBus.post(AnalyticsScreenEvent(SCREEN_GLOBAL_DASHBOARD))
+    }
+
+    private fun loadGlobalDashboardFragment() {
+        with(supportFragmentManager) {
+            commit {
+                val fragment = GlobalDashboardFragment()
+                replace(R.id.frame, fragment)
+            }
+        }
```
frett: you will want to make the fragment the primary navigation fragment. you will also want to check to see if the fragment has already been loaded to avoid creating a new fragment every time the activity is re-created

app/src/main/java/org/cru/godtools/activity/GlobalDashboardActivity.kt
```diff
@@ -0,0 +1,41 @@
...
+        super.onResume()
+        mEventBus.post(AnalyticsScreenEvent(SCREEN_GLOBAL_DASHBOARD))
+    }
+
+    private fun loadGlobalDashboardFragment() {
+        val tag = "GlobalDashboardFragment"
+        with(supportFragmentManager) {
+            commit {
+                val fragment = findFragmentByTag(tag) ?: GlobalDashboardFragment()
+                replace(R.id.frame, fragment, tag).setPrimaryNavigationFragment(fragment)
```
frett: 👎 
You don't want to do anything if the fragment is already attached, you should check for the fragments presence and not execute any transaction if the fragment exists.

also, use the fact that it's the primaryNavigationFragment to find it, there is no need to set a tag on the fragment

app/src/main/java/org/cru/godtools/fragment/GlobalDashboardFragment.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        binding = FragmentGlobalDashboardBinding.inflate(inflater, container, false)
+        binding?.viewModel = GlobalDashboardDataModel()
+        return binding?.root
+    }
+
+    override fun onActivityCreated(savedInstanceState: Bundle?) {
+        super.onActivityCreated(savedInstanceState)
+        binding?.viewModel?.theKey?.value = TheKey.getInstance(requireContext())
```
frett: this is not how you should put TheKey singleton in the view model

app/src/main/java/org/cru/godtools/activity/GlobalDashboardActivity.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    override fun onResume() {
+        super.onResume()
+        mEventBus.post(AnalyticsScreenEvent(SCREEN_GLOBAL_DASHBOARD))
+    }
+
+    private fun loadGlobalDashboardFragment() {
+        with(supportFragmentManager) {
+            commit {
+                if (primaryNavigationFragment != null) return@with
```
frett: do this before you start a fragment transaction

app/src/main/java/org/cru/godtools/activity/GlobalDashboardActivity.kt
```diff
@@ -0,0 +1,41 @@
...
+                val fragment = GlobalDashboardFragment()
+                replace(R.id.frame, fragment).setPrimaryNavigationFragment(fragment)
```
frett: separate this into 2 lines, there isn't a need to chain these calls together

app/src/main/java/org/cru/godtools/fragment/GlobalDashboardFragment.kt
```diff
@@ -0,0 +1,27 @@
...
+        binding = FragmentGlobalDashboardBinding.inflate(inflater, container, false)
+        binding?.viewModel = ViewModelProviders.of(this).get(GlobalDashboardDataModel::class.java)
+        return binding?.root
+    }
+
+    override fun onActivityCreated(savedInstanceState: Bundle?) {
+        super.onActivityCreated(savedInstanceState)
+        val key = TheKey.getInstance(requireContext())
+        binding?.viewModel?.firstName?.value = key.cachedAttributes.firstName
+        binding?.viewModel?.lastName?.value = key.cachedAttributes.lastName
```
frett: you can get `TheKey` singleton directly from within the ViewModel

app/src/main/java/org/cru/godtools/fragment/GlobalDashboardFragment.kt
```diff
@@ -0,0 +1,22 @@
...
+import android.view.View
+import android.view.ViewGroup
+import androidx.fragment.app.viewModels
+import org.cru.godtools.databinding.FragmentGlobalDashboardBinding
+import org.cru.godtools.viewmodel.GlobalDashboardDataModel
+import java.util.Calendar
+
+class GlobalDashboardFragment : BasePlatformFragment() {
+    private var binding: FragmentGlobalDashboardBinding? = null
+    private val viewModel: GlobalDashboardDataModel by viewModels { defaultViewModelProviderFactory }
```
frett: you should be able to do this to simplify this a bit more
```suggestion
    private val viewModel: GlobalDashboardDataModel by viewModels()
```

frett:
the Activity crashes when it launches

app/src/main/java/org/cru/godtools/ui/profile/MyProfileActivity.kt
```diff
@@ -25,15 +27,14 @@ class MyProfileActivity : BasePlatformActivity() {
     // region lifeCycle Calls
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        binding = ActivityMyProfileBinding.inflate(layoutInflater).also {
-            setContentView(it.root.id)
-        }
+        binding = DataBindingUtil.setContentView(this, R.layout.activity_my_profile)
```
frett: `DataBindingUtil.setContentView` can cause crashes. we already had to rework logic in mpdx to deal with crashes from this method.

app/src/main/java/org/cru/godtools/ui/profile/GlobalDashboardFragment.kt
```diff
@@ -4,13 +4,13 @@ import android.os.Bundle
...
 import android.view.ViewGroup
+import androidx.fragment.app.Fragment
 import androidx.fragment.app.viewModels
 import org.cru.godtools.databinding.FragmentGlobalDashboardBinding
-import org.cru.godtools.fragment.BasePlatformFragment
 import org.cru.godtools.viewmodel.GlobalDashboardDataModel
 import java.util.Calendar
 
-class GlobalDashboardFragment : BasePlatformFragment() {
+class GlobalDashboardFragment : Fragment() {
```
frett: why are you no longer extending `BasePlatformFragment`?

app/src/main/java/org/cru/godtools/ui/profile/MyProfileActivity.kt
frett: https://github.com/CruGlobal/mpdx-android/pull/891

app/src/main/res/layout/activity_my_profile.xml
```diff
@@ -0,0 +1,71 @@
...
+    </com.google.android.material.tabs.TabLayout>
+
+    <androidx.viewpager.widget.ViewPager
+        android:id="@+id/my_profile_viewpager"
+        android:layout_width="0dp"
+        android:layout_height="0dp"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/my_profile_tab_layout" />
```
frett: for new code you should use `ViewPager2`

app/src/main/res/layout/activity_my_profile.xml
```diff
@@ -0,0 +1,71 @@
...
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools">
+    <data>
+        <variable name="accountName" type="String" />
+        <variable name="accountDate" type="String" />
+    </data>
+<androidx.constraintlayout.widget.ConstraintLayout
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
```
frett: this indention is messed up, if android studio is auto-formatting like this then you should check your settings

app/src/main/java/org/cru/godtools/ui/profile/GlobalDashboardFragment.kt
gyasistory: I wasn't using anything from the BasePlatformFragment so I didn't use it.

frett: If you are planning on using data binding there is a `BaseBindingPlatformFragment` available to extend now.

Currently you only have access to the binding object when the binding is created or destroyed, I want to see if that provides everything we need when using databinding in a fragment to avoid needing to use a nullable property everywhere

app/build.gradle
```diff
@@ -149,6 +149,7 @@ dependencies {
     implementation "androidx.lifecycle:lifecycle-livedata-ktx:${deps.androidX.fragment}"
     implementation "androidx.percentlayout:percentlayout:${deps.androidX.percentLayout}"
     implementation "androidx.swiperefreshlayout:swiperefreshlayout:${deps.androidX.swipeRefreshLayout}"
+    implementation "androidx.lifecycle:lifecycle-extensions:${deps.androidX.lifecycle}"
```
frett: this duplicates the dependency defined on line 148

app/build.gradle
```diff
@@ -159,6 +160,7 @@ dependencies {
     implementation "org.ccci.gto.android:gto-support-lifecycle:${deps.gtoSupport}"
     implementation "org.ccci.gto.android:gto-support-okhttp3:${deps.gtoSupport}"
     implementation "org.ccci.gto.android:gto-support-picasso:${deps.gtoSupport}"
+    implementation "org.ccci.gto.android:gto-support-lifecycle:${deps.gtoSupport}"
```
frett: this duplicates the dependency defined on line 160

app/src/main/java/org/cru/godtools/viewmodel/GlobalDashboardDataModel.kt
```diff
@@ -3,12 +3,17 @@ package org.cru.godtools.viewmodel
...
 
-class GlobalDashboardDataModel(application: Application) : AndroidViewModel(application) {
+class GlobalDashboardDataModel(app: Application) : AndroidViewModel(app) {
 
-    // TODO:  Create temporary Data to be updated in later PR
-    val uniqueUsers = MutableLiveData<String>()
-    val gospelPresentation = MutableLiveData<String>()
-    val sessions = MutableLiveData<String>()
-    val countries = MutableLiveData<String>()
+    private val dao = GodToolsDao.getInstance(getApplication()).getLiveData(Query.select(GlobalActivityAnalytics::class.java).where("_id", 1))
```
frett: this isn't a `dao`, it's a specific live data query handle.

frett: also, there are better options for fetching a single record.

`dao.findLiveData(GlobalActivityAnalytics::class.java, 1)`

app/src/main/java/org/cru/godtools/viewmodel/GlobalDashboardDataModel.kt
```diff
@@ -3,12 +3,17 @@ package org.cru.godtools.viewmodel
...
+
+    val uniqueUsers = dao.map { it.first().users.toString() }
+    val gospelPresentation = dao.map { it.first().gospelPresentation.toString() }
+    val sessions = dao.map { it.first().launches.toString() }
+    val countries = dao.map { it.first().countries.toString() }
```
frett: why are you forcing them to be strings?

gyasistory: This is changed

app/src/main/java/org/cru/godtools/viewmodel/GlobalDashboardDataModel.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.ccci.gto.android.common.db.getAsLiveData
+import org.cru.godtools.model.GlobalActivityAnalytics
+import org.keynote.godtools.android.db.GodToolsDao
+
+class GlobalDashboardDataModel(app: Application) : AndroidViewModel(app) {
+
+    private val globalAnalyticsQuery: LiveData<List<GlobalActivityAnalytics>>? by lazy {
+        Query.select(GlobalActivityAnalytics::class.java)
+            .getAsLiveData(GodToolsDao.getInstance(getApplication()))
+    }
```
frett: did you see my comment about using `dao.findLiveData()` instead of using a query?

gyasistory: Sorry about that I will adjust that now.

app/src/main/java/org/cru/godtools/ui/profile/GlobalDashboardFragment.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.cru.godtools.viewmodel.GlobalDashboardDataModel
+import java.util.Calendar
+
+class GlobalDashboardFragment :
+    BaseBindingPlatformFragment<FragmentGlobalDashboardBinding>(R.layout.fragment_global_dashboard) {
+    private val viewModel: GlobalDashboardDataModel by viewModels()
+
+    override fun onBindingCreated(binding: FragmentGlobalDashboardBinding, savedInstanceState: Bundle?) {
+        binding.viewModel = viewModel
+        binding.lifecycleOwner = this
```
frett: this is the wrong lifecycleOwner to assign. plus this is being assigned in `BaseBindingPlatformFragment` already

app/src/main/java/org/cru/godtools/ui/profile/MyProfileActivity.kt
```diff
@@ -0,0 +1,66 @@
...
+                }.attach()
+            }
+        }
+    }
+
+    private class MyProfilePageAdapter(fm: FragmentActivity) :
+        FragmentStateAdapter(fm) {
+        override fun getItemCount() = 1
+        override fun createFragment(position: Int) = GlobalDashboardFragment()
+    }
```
frett: this doesn't need to be an inner class

library/sync-tasks/src/main/java/org/cru/godtools/sync/task/GlobalActivityAnalyticsSyncTask.java
```diff
@@ -0,0 +1,78 @@
...
+        return true;
+    }
+
+    private void storeGlobalAnalytics(List<GlobalActivityAnalytics> data,
+                                      LongSparseArray<GlobalActivityAnalytics> existing) {
+        for (final GlobalActivityAnalytics globalActivityAnalytics : data) {
+            if (existing != null) {
+                existing.remove(globalActivityAnalytics.getId());
+            }
+            mDao.updateOrInsert(globalActivityAnalytics, CONFLICT_REPLACE, API_FIELDS_GLOBAL_ANALYTICS);
```
frett: use `mDao.replace()` instead

library/sync-tasks/src/main/java/org/cru/godtools/sync/task/GlobalActivityAnalyticsSyncTask.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+        return true;
+    }
+
+    private void storeGlobalAnalytics(List<GlobalActivityAnalytics> data,
+                                      LongSparseArray<GlobalActivityAnalytics> existing) {
+        for (final GlobalActivityAnalytics globalActivityAnalytics : data) {
+            if (existing != null) {
+                existing.remove(globalActivityAnalytics.getId());
+            }
```
frett: there is no need to remove the existing value from the database

app/src/main/java/org/cru/godtools/viewmodel/GlobalDashboardDataModel.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import android.app.Application
+import androidx.lifecycle.AndroidViewModel
+import androidx.lifecycle.map
+import org.cru.godtools.model.GlobalActivityAnalytics
+import org.keynote.godtools.android.db.GodToolsDao
+
+class GlobalDashboardDataModel(app: Application) : AndroidViewModel(app) {
+
+    private val globalActivityAnalytics by lazy {
```
frett: why do you need lazy here? is it actually doing anything?

app/src/main/java/org/cru/godtools/ui/profile/MyProfileActivity.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.cru.godtools.activity.BasePlatformActivity
+import org.cru.godtools.analytics.model.AnalyticsScreenEvent
+import org.cru.godtools.analytics.model.AnalyticsScreenEvent.Companion.SCREEN_GLOBAL_DASHBOARD
+import org.cru.godtools.base.ui.activity.BaseActivity
+import org.cru.godtools.databinding.ActivityMyProfileBinding
+
+fun Activity.startMyProfileActivity() {
+    Intent(this, MyProfileActivity::class.java)
+        .putExtras(BaseActivity.buildExtras(this))
+        .also { startActivity(it) }
```
frett: the `.also` scope method here doesn't really gain anything and just makes this logic slightly harder to understand

app/src/main/java/org/cru/godtools/ui/profile/MyProfileActivity.kt
```diff
@@ -0,0 +1,66 @@
...
+        super.onCreate(savedInstanceState)
+        binding = ActivityMyProfileBinding.inflate(layoutInflater)
+        setContentView(binding?.root)
+        setBindingData()
+    }
+
+    override fun onResume() {
+        super.onResume()
+        mEventBus.post(AnalyticsScreenEvent(SCREEN_GLOBAL_DASHBOARD))
+        binding?.myProfileTabLayout?.getTabAt(0)?.setText(R.string.gt_gd_activity_text)
```
frett: why are you updating the tab name both here and on line 55?

app/src/main/java/org/cru/godtools/ui/profile/MyProfileActivity.kt
```diff
@@ -0,0 +1,66 @@
...
+
+fun Activity.startMyProfileActivity() {
+    Intent(this, MyProfileActivity::class.java)
+        .putExtras(BaseActivity.buildExtras(this))
+        .also { startActivity(it) }
+}
+
+class MyProfileActivity : BasePlatformActivity() {
+
+    private var binding: ActivityMyProfileBinding? = null
```
frett: in an Activity the view is only created once (unless you trigger `setContentView` other times during an Activity's lifespan).

because of that you don't need to make the binding nullable in an activity unless you have use cases that won't create the binding.

app/src/main/java/org/cru/godtools/ui/profile/MyProfileActivity.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    override fun onResume() {
+        super.onResume()
+        mEventBus.post(AnalyticsScreenEvent(SCREEN_GLOBAL_DASHBOARD))
+        binding?.myProfileTabLayout?.getTabAt(0)?.setText(R.string.gt_gd_activity_text)
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        binding = null
```
frett: there should be no need to clear out the binding when destroying the activity

app/src/main/java/org/cru/godtools/ui/profile/MyProfileActivity.kt
gyasistory: @frett do you mean moving it to its own file?

frett: no, you can just put the class definition in this file, but after the current class

library/sync-tasks/src/main/java/org/cru/godtools/sync/task/GlobalActivityAnalyticsSyncTask.java
```diff
@@ -0,0 +1,74 @@
...
+                    return null;
+                });
+            }
+            mDao.updateLastSyncTime(SYNC_TIME_GLOBAL_ANALYTICS);
+        }
+        return true;
+    }
+
+    private void storeGlobalAnalytics(List<GlobalActivityAnalytics> data,
+                                      LongSparseArray<GlobalActivityAnalytics> existing) {
```
frett: what's the purpose of passing in existing?

app/src/main/java/org/cru/godtools/ui/profile/MyProfileActivity.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.cru.godtools.activity.BasePlatformActivity
+import org.cru.godtools.analytics.model.AnalyticsScreenEvent
+import org.cru.godtools.analytics.model.AnalyticsScreenEvent.Companion.SCREEN_GLOBAL_DASHBOARD
+import org.cru.godtools.base.ui.activity.BaseActivity
+import org.cru.godtools.databinding.ActivityMyProfileBinding
+
+fun Activity.startMyProfileActivity() {
+    val intent = Intent(this, MyProfileActivity::class.java)
+        .putExtras(BaseActivity.buildExtras(this))
+    startActivity(intent)
```
frett: just wrap the intent with `startActivity`.
```kotlin
startActivity(Intent(this, class)
    .putExtras())
```

app/src/main/java/org/cru/godtools/viewmodel/GlobalDashboardDataModel.kt
```diff
@@ -0,0 +1,27 @@
...
+    }
+    val gospelPresentation = globalActivityAnalytics.map {
+        NumberFormat.getInstance().format(it?.gospelPresentation ?: 0)
+    }
+    val sessions = globalActivityAnalytics.map {
+        NumberFormat.getInstance().format(it?.launches ?: 0)
+    }
+    val countries = globalActivityAnalytics.map {
+        NumberFormat.getInstance().format(it?.countries ?: 0)
+    }
```
frett: this is very repetitive, and probably should be handled with a databinding adapter along with binding expressions right in the layout xml

app/src/main/java/org/cru/godtools/ui/profile/MyProfileActivity.kt
```diff
@@ -0,0 +1,57 @@
...
+                TabLayoutMediator(tabLayout, viewPager) { tab, _ ->
+                    tab.text = getString(R.string.gt_gd_activity_text)
+                    viewPager.setCurrentItem(tab.position, true)
+                }.attach()
+            }
+        }
+    }
+}
+
+private class MyProfilePageAdapter(fm: FragmentActivity) : FragmentStateAdapter(fm) {
```
frett: minor nit, what does the `m` in `fm` stand for?

app/src/main/java/org/cru/godtools/ui/profile/ProfileActivity.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    private fun setBindingData(binding: ActivityMyProfileBinding) {
+        val key = TheKey.getInstance(this)
+        binding.accountName = "${key.cachedAttributes.firstName} ${key.cachedAttributes.lastName}"
+        binding.myProfileViewpager?.adapter = ProfilePageAdapter(this)
+        binding.myProfileTabLayout?.let { tabLayout ->
+            binding.myProfileViewpager?.let { viewPager ->
+                TabLayoutMediator(tabLayout, viewPager) { tab, _ ->
+                    tab.text = getString(R.string.gt_gd_activity_text)
+                    viewPager.setCurrentItem(tab.position, true)
```
frett: what is the point of this statement?

library/analytics/src/main/java/org/cru/godtools/analytics/model/AnalyticsScreenEvent.kt
```diff
@@ -31,6 +31,7 @@ open class AnalyticsScreenEvent @JvmOverloads constructor(val screen: String, lo
 
         /* Adobe Site Sections */
         const val ADOBE_SITE_SECTION_TOOLS = "tools"
+        const val SCREEN_GLOBAL_DASHBOARD = "Global Dashboard"
```
frett: this isn't an adobe site section, it shouldn't be under the `Adobe Site Sections` constants

app/src/main/java/org/cru/godtools/ui/profile/ProfileActivity.kt
```diff
@@ -0,0 +1,58 @@
...
+        mEventBus.post(AnalyticsScreenEvent(SCREEN_GLOBAL_DASHBOARD))
+    }
+    // endregion Lifecycle
+
+    // region Data Binding
+    private lateinit var binding: ProfileActivityBinding
+
+    private fun setBindingData() {
+        val key = TheKey.getInstance(this)
+        binding.accountName = "${key.cachedAttributes.firstName} ${key.cachedAttributes.lastName}"
```
frett: you can now get a LiveData handle around the Attributes object using `key.getAttributesLiveData()`. you should utilize this in data binding so that the name is updated when the attributes change.

PR #840, opened 2020-01-27
GT-670 : EditText to dark to view

gyasistory:
https://jira.cru.org/browse/GT-670

I set the EditText to Transparent to allow visibility.

### DIFF
--- a/ui/tract-renderer/src/main/res/layout/tract_content_input.xml
+++ b/ui/tract-renderer/src/main/res/layout/tract_content_input.xml
@@ -9,5 +9,6 @@
         android:id="@+id/input"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
+        android:background="@android:color/transparent"
         android:textCursorDrawable="@null" />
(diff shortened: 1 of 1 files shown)


ui/tract-renderer/src/main/res/layout/tract_content_input.xml
```diff
@@ -9,5 +9,6 @@
         android:id="@+id/input"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
+        android:background="@android:color/transparent"
```
frett: I want to take a look at the material components theming, this may not be the correct way to do this with the material components

frett:
this isn't how I want to address this, but I'm not sure exactly how I want to handle it yet

frett:
#842 addresses this by actually leveraging the TextInputLayout styling

frett:
https://github.com/material-components/material-components-android/blob/master/docs/components/TextInputLayout.md

PR #843, opened 2020-01-28
Performance fixes

frett:


frett:
these are a couple things that could help a little bit with launch performance

PR #846, opened 2020-01-29
GT-661   fix site sub section

gyasistory:
https://jira.cru.org/browse/GT-634

### DIFF
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/Constants.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/Constants.kt
@@ -1,3 +1,4 @@
 package org.cru.godtools.tutorial.analytics.model
 
+internal const val ADOBE_SITE_SECTION_ONBOARDING = "onboarding"
 internal const val ADOBE_SITE_SECTION_TUTORIAL = "tutorial"
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/Constants.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/Constants.kt
@@ -1,6 +1,3 @@
 package org.cru.godtools.tutorial.analytics.model
 
 internal const val ADOBE_SITE_SECTION_TUTORIAL = "tutorial"
-internal const val ADOBE_SITE_SECTION_ON_BOARDING = "on_boarding"
-internal const val BASE_SCREEN_NAME_ON_BOARDING = "onboarding"
-internal const val BASE_SCREEN_NAME_TUTOIRAL = "tutorial"
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/Constants.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/Constants.kt
@@ -1,3 +1,6 @@
 package org.cru.godtools.tutorial.analytics.model
 
 internal const val ADOBE_SITE_SECTION_TUTORIAL = "tutorial"
+internal const val ADOBE_SITE_SECTION_ON_BOARDING = "on_boarding"
+internal const val BASE_SCREEN_NAME_ON_BOARDING = "onboarding"
+internal const val BASE_SCREEN_NAME_TUTOIRAL = "tutorial"
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/TutorialAnalyticsScreenEvent.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/TutorialAnalyticsScreenEvent.kt
@@ -4,7 +4,10 @@ import org.cru.godtools.analytics.model.AnalyticsScreenEvent
 import org.cru.godtools.tutorial.PageSet
 import java.util.Locale
 
-class TutorialAnalyticsScreenEvent(tutorial: PageSet, page: Int) :
+class TutorialAnalyticsScreenEvent(val tutorial: PageSet, page: Int) :
     AnalyticsScreenEvent("${tutorial.analyticsBaseScreenName}-${page + 1}", Locale.getDefault()) {
-    override val adobeSiteSection get() = ADOBE_SITE_SECTION_TUTORIAL
+    override val adobeSiteSection get() = when(tutorial.analyticsBaseScreenName) {
+        BASE_SCREEN_NAME_ON_BOARDING -> ADOBE_SITE_SECTION_ON_BOARDING
+        else -> ADOBE_SITE_SECTION_TUTORIAL
+    }
 }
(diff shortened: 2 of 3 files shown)


ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/TutorialAnalyticsScreenEvent.kt
```diff
@@ -4,7 +4,11 @@ import org.cru.godtools.analytics.model.AnalyticsScreenEvent
...
 
-class TutorialAnalyticsScreenEvent(tutorial: PageSet, page: Int) :
+class TutorialAnalyticsScreenEvent(val tutorial: PageSet, page: Int) :
     AnalyticsScreenEvent("${tutorial.analyticsBaseScreenName}-${page + 1}", Locale.getDefault()) {
-    override val adobeSiteSection get() = ADOBE_SITE_SECTION_TUTORIAL
+    override val adobeSiteSection
+        get() = when (tutorial.analyticsBaseScreenName) {
+            BASE_SCREEN_NAME_ON_BOARDING -> ADOBE_SITE_SECTION_ON_BOARDING
+            else -> ADOBE_SITE_SECTION_TUTORIAL
+        }
```
frett: you know that `tutorial` is an enum, there isn't a need to extract string constants to switch between the different `PageSets`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/TutorialAnalyticsScreenEvent.kt
```diff
@@ -4,7 +4,11 @@ import org.cru.godtools.analytics.model.AnalyticsScreenEvent
 import org.cru.godtools.tutorial.PageSet
 import java.util.Locale
 
-class TutorialAnalyticsScreenEvent(tutorial: PageSet, page: Int) :
+class TutorialAnalyticsScreenEvent(val tutorial: PageSet, page: Int) :
```
frett: make it a private val

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/Constants.kt
```diff
@@ -1,3 +1,6 @@
 package org.cru.godtools.tutorial.analytics.model
 
 internal const val ADOBE_SITE_SECTION_TUTORIAL = "tutorial"
+internal const val ADOBE_SITE_SECTION_ON_BOARDING = "on_boarding"
+internal const val BASE_SCREEN_NAME_ON_BOARDING = "onboarding"
```
frett: I'm guessing the difference between the text of these 2 constants is intentional?

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/Constants.kt
```diff
@@ -1,3 +1,6 @@
...
+internal const val BASE_SCREEN_NAME_TUTOIRAL = "tutorial"
```
frett: the `BASE_SCREEN_NAME_*` constants are unncessary

frett:
these changes don't address the adobe site section for action events

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/TutorialAnalyticsScreenEvent.kt
gyasistory: @frett I am not sure what you mean by this comment.

frett:
```kotlin
get() = when (tutorial) {
    ONBOARDING -> ADOBE_SITE_SECTION_ON_BOARDING
    TUTORIAL -> ADOBE_SITE_SECTION_TUTORIAL
    else -> null
}
```

gyasistory: AH I see

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/Constants.kt
gyasistory: @frett so are you saying the constants should only be named `ON_BOARDING` and  `TUTORIAL`.

frett: I'm saying it was unnecessary to make these constants.

frett: the existence of these constants I don't feel strongly one way or another, but in the context of the current usage it was unnecessary to move these values to constants

frett: we really should try to exclude unnecessary changes from PRs, it will reduce the amount of code that someone needs to look at when reviewing the PR

gyasistory: @frett Removing the constants will mean that I will have to evaluate the `when` on a string (example below).  Not sure if that is the best solution.

```
 override val adobeSiteSection
        get() = when (tutorial.analyticsBaseScreenName) {
            "onboarding" -> ADOBE_SITE_SECTION_ON_BOARDING
            "tutorial" -> ADOBE_SITE_SECTION_TUTORIAL
            else -> null
        }
```

frett: Not if you change the when to switching on the enums like the other comment suggested

gyasistory: @frett Are you saying just set the `adobeSiteSection = tutorial.analyticsBaseScreenName`

frett: No, I'm saying you shouldn't need to use `analyticsBaseScreenName` at all for the adobe site section

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/analytics/model/TutorialAnalyticsScreenEvent.kt
frett: I don't think you understood what I was saying here at all

frett: this suggested change was the reason that the `BASE_SCREEN_NAME_*` constants weren't needed

frett:
@gyasistory this was what I was trying to push you towards

PR #850, opened 2020-01-30
Setup AppsFlyer

frett:


library/analytics/build.gradle
```diff
@@ -6,6 +6,7 @@ android {
             }
         }
 
+        buildConfigField "String", "APPSFLYER_DEV_KEY", "\"QdbVaVHi9bHRchUTWtoaij\""
         buildConfigField "String", "GOOGLE_ANALYTICS_CLIENT_ID", "\"UA-325725-50\""
```
gyasistory: @frett should this be encrypted?

frett: that actually should be removed, Google Analytics is no longer in use.

I'll take care of that in a different PR

frett: and it makes no sense to encrypt these constants because it doesn't gain us any security

PR #866, opened 2020-02-05
LanguagesFragment fixes

frett:


app/src/main/res/layout/list_item_language.xml
```diff
@@ -18,7 +19,7 @@
         android:background="@drawable/bkg_list_item_language"
         android:minHeight="48dp"
         android:onClick="@{() -> listener.onLanguageSelected(language)}"
-        app:selected="@{language.code == selected}">
+        app:selected="@{ObjectsCompat.equals(language.code, selected)}">
```
frett: I had thought that Data Binding would correctly write logic utilizing the `equals` method, but it doesn't :(

PR #874, opened 2020-02-10
GT-682 : Tool info analytic screen name

gyasistory:
https://jira.cru.org/browse/GT-682

### DIFF
--- a/app/src/main/java/org/cru/godtools/analytics/model/ToolDetailsScreenEvent.kt
+++ b/app/src/main/java/org/cru/godtools/analytics/model/ToolDetailsScreenEvent.kt
@@ -2,7 +2,6 @@ package org.cru.godtools.analytics.model
 
 import java.util.Locale
 
-private const val SITE_SECTION_TOOLS = "tools"
 private const val SITE_SUB_SECTION_ADD_TOOLS = "add tools"
 
 class ToolDetailsScreenEvent(toolCode: String, locale: Locale? = null) :
--- a/app/src/main/java/org/cru/godtools/analytics/model/ToolDetailsScreenEvent.kt
+++ b/app/src/main/java/org/cru/godtools/analytics/model/ToolDetailsScreenEvent.kt
@@ -3,8 +3,13 @@ package org.cru.godtools.analytics.model
 import java.util.Locale
 
 private const val SITE_SECTION_TOOLS = "tools"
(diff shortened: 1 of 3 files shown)


library/analytics/src/main/java/org/cru/godtools/analytics/model/AnalyticsScreenEvent.kt
```diff
@@ -36,7 +32,6 @@ open class AnalyticsScreenEvent @JvmOverloads constructor(val screen: String, lo
     override val adobeSiteSection
         get() = when (screen) {
             SCREEN_FIND_TOOLS,
-            SCREEN_TOOL_DETAILS -> SITE_SECTION_TOOLS
```
frett: this breaks `adobeSiteSection` for `SCREEN_FIND_TOOLS`

library/analytics/src/main/java/org/cru/godtools/analytics/model/AnalyticsScreenEvent.kt
```diff
@@ -6,10 +6,7 @@ import javax.annotation.concurrent.Immutable
 
 private const val SNOWPLOW_CONTENT_SCORING_URI_PATH_SCREEN = "screenview"
 
-/* Adobe Site Sections */
```
frett: this comment shouldn't be removed, the remaining constants are still for adobe site section

app/src/main/java/org/cru/godtools/analytics/model/ToolDetailsScreenEvent.kt
```diff
@@ -0,0 +1,15 @@
...
+private const val SITE_SECTION_TOOLS = "tools"
+private const val SITE_SUB_SECTION_ADD_TOOLS = "add tools"
+
+class ToolDetailsScreenEvent(toolCode: String, locale: Locale? = null) :
+    AnalyticsScreenEvent("$toolCode-tool-info", locale) {
+    override val adobeSiteSection: String?
+        get() = SITE_SECTION_TOOLS
+
+    override val adobeSiteSubSection: String?
+        get() = SITE_SUB_SECTION_ADD_TOOLS
```
frett: I know this was the previous behavior, but this seems wrong. can we double check to see if tool info should have this sub section?

PR #907, opened 2020-03-02
Gt-682 : adding analytic to exit links in tool detail

gyasistory:
https://jira.cru.org/browse/GT-682


app/src/main/java/org/cru/godtools/widget/AnalyticsMovementMethod.kt
```diff
@@ -0,0 +1,52 @@
...
+        val links = buffer?.getSpans(off, off, ClickableSpan::class.java)
+        links?.forEach {
+            if (it is URLSpan) {
+                it.url.let { url ->
+                    eventBus.post(ExitLinkActionEvent(Uri.parse(url)))
+                }
+            }
+        }
+        return super.onTouchEvent(widget, buffer, event)
+    }
```
frett: this is a very hacky way to trigger the analytics, I would rather see us use a custom `URLSpan` class if possible.

app/src/main/java/org/cru/godtools/util/ViewUtils.kt
```diff
@@ -16,3 +26,26 @@ fun PicassoImageView?.bindLocalImage(attachment: Attachment?) =
...
+) {
+        val spannable = SpannableString(changedText)
+        val matcher = Patterns.WEB_URL.matcher(spannable)
+        while (matcher.find()) {
+            val url = spannable.toString().substring(matcher.start(), matcher.end())
+            val urlSpan = object : URLSpan(changedText.toString()) {
+                override fun onClick(widget: View) {
+                    val uri = Uri.parse(url)
+                    EventBus.getDefault().post(ExitLinkActionEvent(uri))
+                    context.startActivity(Intent(Intent.ACTION_VIEW, uri))
```
frett: use `super.onClick(widget)` here to defer to the framework behavior for launching the external link

app/src/main/java/org/cru/godtools/util/ViewUtils.kt
```diff
@@ -16,3 +26,26 @@ fun PicassoImageView?.bindLocalImage(attachment: Attachment?) =
     bindLocalImage(attachment?.takeIf { it.isDownloaded }?.localFileName)
 
 fun PicassoImageView?.bindLocalImage(filename: String?) = this?.setPicassoFile(context.getGodToolsFile(filename))
+
+fun TextView.setTextViewWebLinksForAnalytics(
```
frett: I like this better, I want to dig into this logic a bit before I have a more in depth review

gyasistory: @frett OK I am cleaning up the conflict right now.

app/src/main/res/layout/tool_details_page_description.xml
```diff
@@ -14,11 +15,10 @@
         style="@style/Widget.GodTools.Tools.Details.Page.Content"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:autoLink="web"
         android:paddingLeft="@dimen/tool_details_margin_horizontal"
         android:paddingTop="@dimen/tool_details_page_padding_vertical"
         android:paddingRight="@dimen/tool_details_margin_horizontal"
         android:paddingBottom="@dimen/tool_details_page_padding_vertical"
-        android:text="@{ModelUtils.getTranslationDescription(translation, tool, context)}"
+        app:webLinkText="@{ModelUtils.getTranslationDescription(translation, tool, context).toString()}"
```
frett: using `toString()` here will break fonts for Sinhala and Tibetan on older devices

app/src/main/java/org/cru/godtools/adapter/TextViewAdapter.kt
```diff
@@ -0,0 +1,37 @@
...
+                val uri = Uri.parse(url)
+                EventBus.getDefault().post(ExitLinkActionEvent(uri))
+                super.onClick(widget)
+            }
+        }
+        spannable.setSpan(urlSpan, matcher.start(), matcher.end(), Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+    }
+    text = spannable
+    movementMethod = LinkMovementMethod.getInstance() // Make link clickable
+}
```
frett: what are your thoughts on this logic for the binding adapter: https://github.com/CruGlobal/android-gto-support/pull/436/files

gyasistory: The only issue I see is what I ran into a link that doesn't start with HTTP will not open when you click it.  Other than that it looks great.

frett: hmm, I would push back that links that don't start with http:// should be updated to be proper URLs since we do control that content.

frett: although if that is a blocker I believe there is an option for LinkifyCompat to automatically add a uri scheme.

frett: I'll research shortly

gyasistory: Android_Emulator_-_Pixel_3a_API_29_5554
These are the two that gave me the issue.

frett: I just tested on those links and they worked fine with the LinkifyCompat logic

PR #932, opened 2020-03-13
GT-734 : expanding onboarding to include Spanish and Chinese

gyasistory:
https://jira.cru.org/browse/GT-734


gyasistory:
We still need to add the translations from one sky.

PR #1100, opened 2020-06-01
GT-784 GT-785 Live Sharing

frett:
Dependent on:
- [x] https://github.com/CruGlobal/mobile-content-api/pull/391
- [x] https://github.com/CruGlobal/android-gto-support/pull/508
- [x] https://github.com/CruGlobal/mobile-content-api/pull/413

Until the mobile-content-api PR is merged to master this needs to be tested against staging


frett:
After 10 seconds we now dismiss the dialog and report we were unable to connect to the live sharing service.

PR #1104, opened 2020-06-02
GT-809 add remove favorite confirmation dialog to tool details and all tools list

frett:


frett:
This will no longer be necessary after we merge #1107

PR #1105, opened 2020-06-02
GT-813 Update share message for tools

frett:


frett:
This has a high line count due to moving a string to a different module so it was available to be used in tools

PR #1107, opened 2020-06-03
GT-793 don't remove translations that aren't favorited

frett:
The actual disk space usage for GodTools data is very minimal, so removing translations isn't really necessary at this time.

We still remove translations if a newer version of the translation has been downloaded

### DIFF
--- a/library/download-manager/src/main/java/org/cru/godtools/download/manager/GodToolsDownloadManager.java
+++ b/library/download-manager/src/main/java/org/cru/godtools/download/manager/GodToolsDownloadManager.java
@@ -54,7 +54,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


library/download-manager/src/main/java/org/cru/godtools/download/manager/GodToolsDownloadManager.java
```diff
@@ -250,18 +247,18 @@ void pruneStaleTranslations() {
...
-                    .and(TranslationTable.FIELD_LAST_ACCESSED.lt(new Date(System.currentTimeMillis() - WEEK_IN_MS)))
-                    .and(TranslationTable.SQL_WHERE_DOWNLOADED), TranslationTable.COLUMN_DOWNLOADED);
+//            // remove any translation that is no longer added to this device and hasn't been accessed within the past
+//            // week
+//            final Translation translation = new Translation();
+//            translation.setDownloaded(false);
+//            int changes = mDao.update(translation, TranslationTable.FIELD_TOOL.notIn(constants(tools))
+//                    .or(TranslationTable.FIELD_LANGUAGE.notIn(constants(languages)))
+//                    .and(TranslationTable.FIELD_LAST_ACCESSED.lt(new Date(System.currentTimeMillis() - WEEK_IN_MS)))
+//                    .and(TranslationTable.SQL_WHERE_DOWNLOADED), TranslationTable.COLUMN_DOWNLOADED);
```
frett, line 257: I left the code in but commented it out because long term I want to rework this to prune unused/stale translations when too much disk space is being used.

PR #1118, opened 2020-06-05
GT-792 Tool categories

frett:


ui/base/src/main/java/org/cru/godtools/base/ui/util/ModelUtils.kt
```diff
@@ -19,3 +21,17 @@ fun Translation?.getTagline(tool: Tool?, context: Context?) =
     this?.let { (tagline ?: description)?.applyTypefaceSpan(getTypeface(context)) } ?: tool?.description ?: ""
 
 private fun Translation.getTypeface(context: Context?) = context?.getTypeface(languageCode)
+
+// region Tool Category
+@JvmName("getToolCategory")
+fun Tool?.getCategory(context: Context, locale: Locale? = null) =
+    this?.category?.let { c -> (locale?.let { context.localize(it) } ?: context).getToolCategoryStringRes(c) ?: c }
```
wrandall22, line 28: Is this saying that `this` can be null?

ui/base/src/test/java/org/cru/godtools/base/ui/util/ModelUtilsRobolectricTest.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    @Test
+    fun testToolGetCategoryValidCategory() {
+        assertEquals("Gospel Invitation", tool.getCategory(context, null))
+        assertEquals("Invitation Evangile", tool.getCategory(context, Locale.FRENCH))
+    }
+
+    @Test
+    fun testToolGetCategoryUnknownCategory() {
+        tool.category = "unknown"
```
wrandall22, line 41: I think in the Java web space, sometimes tests get run out of order. If this test happened to get run before the one above it, wouldn't the one above it fail?

ui/base/src/main/java/org/cru/godtools/base/ui/util/ModelUtils.kt
frett: yeah, when I define the extension function I specified it can be called on a null `Tool` object.

ui/base/src/test/java/org/cru/godtools/base/ui/util/ModelUtilsRobolectricTest.kt
frett: JUnit actually runs each test it a fresh instance of the test class

wrandall22: Ah, interesting.

ui/base/src/main/java/org/cru/godtools/base/ui/util/ModelUtils.kt
frett: so something like this would be valid and execute this function:
```kotlin
val tool: Tool? = null
tool.getCategory(context, locale)
```

the nullability here really matters because it is being called from data binding, which can call this method with a `null` Tool object

PR #1119, opened 2020-06-08
GT-808 utilize LiveData for activeManifest and activeToolState

frett:
This shifts some logic around the active tool and tool state to utilize LiveData for the observer pattern instead of having to manually trigger re-evaluation whenever the underlying data changes.

### DIFF
--- a/ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/BaseSingleToolActivity.kt
+++ b/ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/BaseSingleToolActivity.kt
@@ -20,7 +20,7 @@ abstract class BaseSingleToolActivity(
     @LayoutRes contentLayoutId: Int = INVALID_LAYOUT_RES,
     private val requireTool: Boolean = true
 ) : BaseToolActivity(immersive, contentLayoutId) {
-    override var activeManifest: Manifest? = null
+    override val activeManifestLiveData get() = dataModel.manifest
     private var translationLoaded = false
     private var translation: Translation? = null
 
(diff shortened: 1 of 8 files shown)


ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/BaseToolActivity.kt
```diff
@@ -192,16 +190,23 @@ abstract class BaseToolActivity @JvmOverloads constructor(
...
-        val state = determineActiveToolState()
-        loadingContent?.visibility = if (state == STATE_LOADING) View.VISIBLE else View.GONE
-        mainContent?.visibility = if (state == STATE_LOADED) View.VISIBLE else View.GONE
-        missingContent?.visibility =
-            if (state == STATE_NOT_FOUND || state == STATE_INVALID_TYPE) View.VISIBLE else View.GONE
+    enum class ToolState {
+        LOADING, LOADED, NOT_FOUND, INVALID_TYPE;
+
+        companion object {
+            // TODO: this should be an actual state
```
frett, line 197: this will be addressed in a future PR when I also implement an `OFFLINE` state. I just wanted to start utilizing `UNKNOWN` for now with the same behavior as `LOADING`

ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/BaseSingleToolActivity.kt
```diff
@@ -86,32 +81,23 @@ abstract class BaseSingleToolActivity(
         downloadManager.cacheTranslation(toolCode, locale)
     }
 
-    override fun determineActiveToolState() = when {
-        !hasTool() -> STATE_LOADED
```
wrandall22, line 90: Is this case still being accounted for somewhere?

frett: I don't think this case was originally necessary. the only time `hasTool` would return false would be in `AemArticleActivity`, but that class had it's own implementation of `determineActiveToolState()`.

This logic was probably written before it got overridden in `AemArticleActivity`

PR #1133, opened 2020-06-16
GT-843 Article share message

frett:
This PR updates the share message for all article activities. We still only show the share action when viewing an article, so it doesn't change any behavior for the categories or articles list UI


ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/BaseToolActivity.kt
```diff
@@ -162,6 +162,8 @@ abstract class BaseToolActivity<B : ViewDataBinding>(
 
     protected open fun hasShareLinkUri() = shareLinkUri != null
     protected open val shareLinkTitle get() = activeManifest?.title
+    @get:StringRes
```
wrandall22, line 165: I see you did this here, but not in `BasePlatformActivity`. Is there a reason to use 2 different styles?

frett: the `@StringRes` annotation provides a little static analysis of code to inform if you are passing the correct type of resource id into functions. Beyond the static analysis aspect there is no performance impact or behavior changes based on the annotation.

I moved the string resource to an overridable val here because I wanted to be able to override the value in subclasses of `BaseToolActivity`. `BasePlatformActivity` is in a different hierarchy and has no need for this level of abstraction currently.

I could have included the `@StringRes` annotation to the overridden method in `BaseArticleActivity`, was just an oversight not including it there.

frett: The `get:` prefix defines where the annotation actually gets applied, in this case I want the annotation applied to the getter.

PR #1158, opened 2020-06-24
Simplify nested content parsing

frett:
Reduce code duplication by creating a centralized content parsing extension function

### DIFF
--- a/library/xml-model/src/main/java/org/cru/godtools/xml/model/Parent.kt
+++ b/library/xml-model/src/main/java/org/cru/godtools/xml/model/Parent.kt
@@ -6,6 +6,9 @@ interface Parent : BaseModel {
     val content: List<Content>
 }
 
+/**
+ * @param block Custom parsing logic, if the block processes the current tag, it should advance the parser to the END_TAG event.
+ */
 @OptIn(ExperimentalStdlibApi::class)
 internal inline fun Parent.parseContent(
     parser: XmlPullParser,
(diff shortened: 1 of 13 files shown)


library/xml-model/src/main/java/org/cru/godtools/xml/model/Parent.kt
```diff
@@ -1,5 +1,26 @@
...
+    block: () -> Unit = { }
+) = buildList {
+    parser.require(XmlPullParser.START_TAG, null, null)
+
+    while (parser.next() != XmlPullParser.END_TAG) {
+        if (parser.eventType != XmlPullParser.START_TAG) continue
+
+        // execute any custom parsing logic from the call-site
+        block()
+        if (parser.eventType == XmlPullParser.END_TAG) continue
```
wrandall22, line 21: I don't see this line in the old code. Is it something that was missing? Or is it equated to something else (like `continue@parsingChildren`)?

frett: it's the equivalent because the parser should be on an `END_TAG` if it consumed the current tag.

But thinking about this a bit last night I'm not totally happy with this approach, so I'm going to refactor how I determine if the block processed the node before merging this.

frett: The 2 options I see are:
1. return a boolean if the block consumed the tag
   * This makes all of the blocks being used for parsing more complex and harder to quickly understand
2. state the contract that the block needs to consume the tag if it has processed it and doesn't want the tag to be parsed as content.

The pattern I've established for all of the XML pull processing in GodTools is that any nested parsing behavior starts on a `START_TAG` and returns on an `END_TAG`. With this in mind the `eventType == END_TAG` test is a valid mechanism to determine if nested parsing actually happened.

I'm going with option 2 because all of the parsing Unit tests will catch any invalid behavior, and if some invalid behavior makes it through we can fix that invalid behavior and write a parsing test to ensure it doesn't recur.

PR #1165, opened 2020-06-29
Page position fix

frett:


library/xml-model/src/test/java/org/cru/godtools/xml/model/ManifestTest.kt
```diff
@@ -17,4 +18,14 @@ class ManifestTest {
         assertThat(manifest.pages, empty())
         assertThat(manifest.resources.size, equalTo(0))
     }
+
+    @Test
+    fun verifyParseManifestPages() {
```
gyasistory, line 23: This test is failing

frett:
There was an unused import in `TestParserUtils`

PR #1168, opened 2020-06-30
GT-846 Create a sharing submenu for Tracts

frett:


frett:
yup, that's what they wanted

frett:
hmm, although we do need to address notches in general

frett:
https://medium.com/@hiteshdhamshaniya.wvmagic/notch-android-emulator-with-p-e9b3556d9d5e

PR #1192, opened 2020-07-14
GT-860 Dagger & Coroutines

frett:
- Utilize Coroutines to delay execution of analytics events when requested
- Use Dagger to inject appropriate dependencies from the Dagger object graph


ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.kt
```diff
@@ -352,13 +353,15 @@ class TractActivity : BaseToolActivity<TractActivityBinding>(true, R.layout.trac
...
-        ManifestPagerAdapter().also {
+        ManifestPagerAdapter(pageControllerFactory).also {
             it.setCallbacks(this)
             lifecycle.addObserver(it)
             dataModel.activeManifest.observe(this, it)
         }
     }
     private var initialPage = 0
+    @Inject
+    internal lateinit var pageControllerFactory: PageController.Factory
```
wrandall22, line 364: Stylistically, it seems this should be defined before its usage.

frett: Updated

PR #1220, opened 2020-08-04
GT-877 : Add missing Analytics

gyasistory:
https://docs.google.com/spreadsheets/d/1YLkGQ8ptXWn6xmB82kzUYIkXHsI-BsnA2iQanQdPcnw/edit?usp=sharing


app/src/main/java/org/cru/godtools/ui/tools/analytics/model/ToolsAnalyticsActionEvent.kt
```diff
@@ -0,0 +1,14 @@
+package org.cru.godtools.ui.tools.analytics.model
+
+import org.cru.godtools.analytics.model.AnalyticsActionEvent
+import org.cru.godtools.analytics.model.AnalyticsSystem
+
+sealed class ToolsAnalyticsActionEvent(action: String) : AnalyticsActionEvent(action) {
+    override fun isForSystem(system: AnalyticsSystem): Boolean {
+        return system == AnalyticsSystem.ADOBE
+    }
```
frett, line 9: this is a very simple function and is a good candidate to follow the single expression function idiom. https://kotlinlang.org/docs/reference/idioms.html#single-expression-functions

PR #1223, opened 2020-08-05
GT-897 : Tool Detail Training UI Changes

gyasistory:


app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -57,6 +57,7 @@ class ToolDetailsFragment() : BasePlatformFragment<ToolDetailsFragmentBinding>(R
         super.onBindingCreated(binding, savedInstanceState)
         binding.fragment = this
         binding.tool = dataModel.tool
+        binding.hasTraining = dataModel.hasTraining
```
gyasistory, line 60: @frett I just curious why we are adding things one by one instead of just adding the dataModel.

frett: it actually makes testing of the data binding much easier.

If I wanted to test the behavior of `hasTraining` I could provide a mock instance of that without needing to create a full data model object.

app/src/main/res/values/strings_tools.xml
```diff
@@ -18,6 +18,7 @@
     <string name="action_tools_open_tool">Open Tool</string>
     <string name="action_tools_add_favorite">Add to Favorites</string>
     <string name="action_tools_remove_favorite">Remove from Favorites</string>
+    <string name="action_tool_training">Learn to share this tool</string>
```
frett, line 21: rename the string to `action_tools_open_training`

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
frett: take a look at https://github.com/CruGlobal/godtools-android/blob/develop/app/src/test/java/org/cru/godtools/databinding/ToolsListItemToolBindingTest.kt to see how setting individual LiveData objects on the binding allows us to more easily test behavior without having to mock the entire data model

frett: also, the data binding layouts are a bit more portable if it depends on 4 LiveData objects vs a data model that is more tightly coupled to a single fragment or activity.

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -57,6 +58,7 @@ class ToolDetailsFragment() : BasePlatformFragment<ToolDetailsFragmentBinding>(R
         super.onBindingCreated(binding, savedInstanceState)
         binding.fragment = this
         binding.tool = dataModel.tool
+        binding.hasTraining = dataModel.primaryManifest.map { !it?.tips.isNullOrEmpty() }
```
frett, line 61: For now only enable training for debug builds
```suggestion
        binding.hasTraining = dataModel.primaryManifest.map { !it?.tips.isNullOrEmpty() && BuildConfig.DEBUG }
```

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -121,6 +123,16 @@ class ToolDetailsFragment() : BasePlatformFragment<ToolDetailsFragmentBinding>(R
             }
         }
     }
+
+    fun openTraining(tool: Tool?, primaryTranslation: Translation?) {
+        // TODO: Open training activity
+        tool?.code?.let { code ->
+            val primaryLanguage = primaryTranslation?.languageCode ?: Locale.ENGLISH
+
+            manifestManager.get().getLatestPublishedManifestLiveData(code, primaryLanguage)
```
frett, line 132: this won't be necessary for training, the manifest will have already been loaded

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -121,6 +123,16 @@ class ToolDetailsFragment() : BasePlatformFragment<ToolDetailsFragmentBinding>(R
             }
         }
     }
+
+    fun openTraining(tool: Tool?, primaryTranslation: Translation?) {
```
frett, line 127: hmm, I'm thinking passing in the manifest you want to open would make more sense than to use tool and primaryTranslation

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
frett: we actually can probably remove this preloading functionality for all of the tool details activity because we are always loading the manifest to determine if there are tips available.

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
frett: That would mean instead of doing a `hasTraining` variable on data binding, just put the `manifest` in data binding and evaluate if there are tips directly in data binding. You will also need to import `BuildConfig` and test the `DEBUG` constant to enable training for debug builds only

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -121,6 +126,12 @@ class ToolDetailsFragment() : BasePlatformFragment<ToolDetailsFragmentBinding>(R
             }
         }
     }
+
+    fun openTraining(manifest: Manifest) {
+        if (manifest.type == Manifest.Type.TRACT) {
```
frett, line 131: hmm, I hadn't realized we had a different manifest & tool type

app/src/main/res/layout/tool_details_fragment.xml
```diff
@@ -141,12 +143,23 @@
...
 
+                <com.google.android.material.button.MaterialButton
+                    android:id="@+id/action_tool_training"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:layout_marginLeft="@dimen/tool_details_margin_horizontal"
+                    android:layout_marginRight="@dimen/tool_details_margin_horizontal"
+                    android:onClick="@{() -> fragment.openTraining(manifest)}"
+                    android:text="@string/action_tools_open_training"
+                    app:goneIf="@{!hasTraining}"
```
frett, line 154: now that the manifest is present in data binding, this can be changed to `app:visibleIf="@{!manifest.tips.isEmpty &amp;&amp; BuildConfig.DEBUG}"`

frett: up to you whether you use `app:visibleIf` or `app:goneIf`

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
gyasistory: Yeah, just Article and Tract.  Is it ok to leave it like this or should I handle the else condition

frett: this is fine, at some point I'll re-evaluate the types to see if there is a better option

PR #1227, opened 2020-08-07
GT-904 Disable immersive mode

frett:
Disable immersive mode and color the status bar when viewing a tool.

Dependent on:
- [x] https://github.com/CruGlobal/android-gto-support/pull/601

### DIFF
--- a/ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/ImmersiveActivity.kt
+++ b/ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/ImmersiveActivity.kt
@@ -1,18 +1,12 @@
 package org.cru.godtools.base.tool.activity
 
-import android.os.Build
 import android.os.Bundle
-import android.view.View
 import androidx.annotation.LayoutRes
-import androidx.annotation.RequiresApi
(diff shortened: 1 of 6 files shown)


ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/ImmersiveActivity.kt
```diff
@@ -1,36 +1,19 @@
...
-import androidx.annotation.RequiresApi
 import androidx.databinding.DataBindingUtil
 import androidx.databinding.ViewDataBinding
 import org.cru.godtools.base.ui.activity.BaseActivity
 
-abstract class ImmersiveActivity<B : ViewDataBinding>(
-    private val enableImmersive: Boolean,
-    @LayoutRes private val contentLayoutId: Int
-) : BaseActivity() {
+abstract class ImmersiveActivity<B : ViewDataBinding>(@LayoutRes private val contentLayoutId: Int) : BaseActivity() {
```
frett, line 9: eventually this class will go away, but first the ViewDataBinding management needs to be moved to `BaseActivity`

PR #1244, opened 2020-08-14
GT-915 : Add null check for manifest

gyasistory:


frett:
take a look at these tests as an example: https://github.com/CruGlobal/godtools-android/blob/develop/app/src/test/java/org/cru/godtools/databinding/ToolsListItemToolBindingTest.kt#L133-L169

app/src/test/java/org/cru/godtools/databinding/ToolDetailsFragmentBindingTest.kt
```diff
@@ -0,0 +1,45 @@
...
+    fun createBinding(){
+        val activityController = Robolectric.buildActivity(ToolDetailsActivity::class.java)
+        binding = ToolDetailsFragmentBinding.inflate(LayoutInflater.from(activityController.get()), null, false)
+    }
+
+    @Test
+    fun verifyWithTips() {
+        val manifest = Manifest(TOOL_CODE, Locale.ENGLISH, getXmlParserForResource("manifest_tips_invalid.xml")){
+            getXmlParserForResource(it)
+        }
```
frett, line 37: instead of parsing xml to get the manifest, just create it using something like this:
```kotlin
val manifest = Manifest(tips = { listOf(Tip(it, "tip1")) })
```

gyasistory: @frett sorry I didn't see this I used mock values.  let me know if you are OK with that approach

frett: use these testing constructors, it easier than having to mock everything for tool models

app/src/test/java/org/cru/godtools/databinding/ToolDetailsFragmentBindingTest.kt
```diff
@@ -0,0 +1,54 @@
...
+        val manifest = Manifest()
+        binding.manifest = MutableLiveData<Manifest>(manifest)
+        binding.executePendingBindings()
+
+        assertEquals(View.GONE, binding.actionToolTraining.visibility)
+    }
+
+    @Test
+    fun verifyEmptyManifest() {
+        binding.manifest = MutableLiveData<Manifest>()
```
frett, line 49:
```suggestion
        binding.manifest = MutableLiveData<Manifest?>(null)
```

frett: I don't think this would actually matter for the actual binding, but it is a bit more correct in the context of how LiveData works. Without a value the LiveData hasn't really been initialized yet.

app/src/test/java/org/cru/godtools/databinding/ToolDetailsFragmentBindingTest.kt
```diff
@@ -32,6 +34,7 @@ class ToolDetailsFragmentBindingTest {
         binding.manifest = MutableLiveData<Manifest>(manifest)
         binding.executePendingBindings()
 
+        assumeTrue(BuildConfig.DEBUG)
```
frett, line 37: generally I put the `assume` calls at the beginning of the test because it prevents the entire test from running when the `assume` isn't satisfied

PR #1249, opened 2020-08-17
GT-865 support compat shadow for inline tips on older android SDKs

frett:
Dependent on:
- [x] https://github.com/CruGlobal/android-gto-support/pull/609


frett:
it was an unrelated change in gto support that I fixed in a separate branch and merged last night

PR #1272, opened 2020-08-26
GT-928 Tips share menu item

frett:
Don't show the share menu icon when the user is in training mode


frett:
I tried to write some tests for this behavior change, but was having trouble getting Dagger and Robolectric to work together for testing the activity interactions.

PR #1329, opened 2020-09-29
GT-937 Lottie animations for tips tutorial

frett:


ui/tutorial-renderer/src/main/res/layout/tutorial_tips_light.xml
```diff
@@ -17,16 +17,17 @@
...
+            android:id="@+id/animation"
             android:layout_width="0dp"
             android:layout_height="0dp"
-            android:scaleType="fitCenter"
-            android:src="@drawable/img_tutorial_tips_tool"
             app:layout_constraintBottom_toBottomOf="@id/imagePosition"
             app:layout_constraintEnd_toEndOf="@id/imagePosition"
             app:layout_constraintStart_toStartOf="@id/imagePosition"
-            app:layout_constraintTop_toTopOf="@id/imagePosition" />
+            app:layout_constraintTop_toTopOf="@id/imagePosition"
```
TheNoodleMoose, line 27: I see that you have it constrained to ```@id/imagePosition```, which after some digging around I figured out was coming from https://github.com/CruGlobal/godtools-android/blob/61a525b7df6d07d1f36d168060d2d360a752f6ba/ui/tutorial-renderer/src/main/res/layout/tutorial_live_share_include_positioning.xml#L44
Is this just a helper layout that we can include in this file specifically for positioning?

TheNoodleMoose (approved):
Should I delete the related images as well once I add the animations to the screens? Also, could you explain why we have multiple ```.png``` and folders for images?

ui/tutorial-renderer/src/main/res/layout/tutorial_tips_light.xml
frett: I created a positioning helper to provide uniform layout between all the pages of a tutorial. I can talk more in detail about this when we talk

PR #1337, opened 2020-10-01
GT-845 Support Deferred Deep Linking for AppsFlyer

frett:
AppsFlyer triggers one of two callback methods whenever it receives a deep link, `onAppOpenAttribution()` or `onConversionDataSuccess()`. This PR adds logic to process these callbacks looking for supported deep links and triggering those deep links appropriately.


library/analytics/src/main/java/org/cru/godtools/analytics/appsflyer/AppsFlyerAnalyticsService.kt
```diff
@@ -58,26 +75,43 @@ class AppsFlyerAnalyticsService @Inject internal constructor(private val app: Ap
...
-        Timber.tag(TAG).e("AppsFlyer onAttributionFailure: %s", error)
+    override fun onActivityResumed(activity: Activity) {
+        activeActivity = activity
     }
 
-    override fun onConversionDataSuccess(data: Map<String, Any>?) {
-        data?.map { (key, value) ->
-            Timber.tag(TAG).d("AppsFlyer onConversionDataSuccess Attribute: %s = %s", key, value)
+    override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) = Unit
+    override fun onActivityStarted(activity: Activity) = Unit
```
TheNoodleMoose, line 87: I've seen this ```= Unit``` in a couple of different places. What is ```Unit```?

library/analytics/src/main/java/org/cru/godtools/analytics/appsflyer/AppsFlyerAnalyticsService.kt
```diff
@@ -23,23 +30,33 @@ import timber.log.Timber
...
+        app.registerActivityLifecycleCallbacks(this)
+        eventBus.register(this)
 
-        GlobalScope.launch(Dispatchers.Default) {
+        GlobalScope.launch(Dispatchers.IO) {
             val mcId = adobeMarketingCloudId
-            withContext(Dispatchers.Main) {
-                appsFlyer.setAdditionalData(hashMapOf("marketingCloudID" to mcId))
-                appsFlyer.startTracking(app)
-                eventBus.register(this@AppsFlyerAnalyticsService)
```
TheNoodleMoose, line 41: Why don't you need these anymore? Is it something that changed with the upgrade?

library/analytics/src/debug/AndroidManifest.xml
```diff
@@ -0,0 +1,26 @@
...
+            <!-- AppsFlyer deep links -->
+            <intent-filter android:autoVerify="true">
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data android:scheme="http" />
+                <data android:scheme="https" />
+                <data android:host="godtools.onelink.me" />
```
TheNoodleMoose, line 18: Does AppsFlyer provide like a default AndroidManifest that we then add our values to?

library/analytics/src/main/java/org/cru/godtools/analytics/appsflyer/AppsFlyerModule.kt
```diff
@@ -0,0 +1,21 @@
...
+@Module
+@InstallIn(SingletonComponent::class)
+abstract class AppsFlyerModule {
+    @Binds
+    @IntoSet
+    @EagerSingleton(threadMode = EagerSingleton.ThreadMode.MAIN)
+    abstract fun AppsFlyerAnalyticsService.eagerSingleton(): Any
+
+    @Multibinds
+    abstract fun deepLinkResolvers(): Set<AppsFlyerDeepLinkResolver>
```
TheNoodleMoose, line 20: Is this binding your deepLinResolvers to AppsFlyer so we can use it?

library/analytics/src/test/java/org/cru/godtools/analytics/appsflyer/AppsFlyerAnalyticsServiceTest.kt
```diff
@@ -0,0 +1,80 @@
...
+        application = mock()
+        activity = mock()
+        eventBus = mock()
+        deepLinkResolver = mock()
+
+        analyticsService = AppsFlyerAnalyticsService(application, eventBus, setOf(deepLinkResolver))
+    }
+
+    // region onAppOpenAttribution()
+    @Test
```
TheNoodleMoose, line 44: Can you test any func as long as you add this ```@Test``` above it, or is there other requirements as well?

ui/base-tool/build.gradle
```diff
@@ -45,5 +45,6 @@ dependencies {
     kapt "com.google.dagger:dagger-compiler:${deps.dagger}"
     kapt "com.google.dagger:hilt-compiler:${deps.hilt}"
 
+    testImplementation project(':ui:tract-renderer')
```
TheNoodleMoose, line 48: If you add tests to a certain part of the app do you need to add this to the build.gradle(if it's not already there) so it will correctly run the tests?

ui/tract-renderer/src/test/java/org/cru/godtools/tract/util/DeepLinkUtilsTest.kt
```diff
@@ -0,0 +1,34 @@
...
+    fun setup() {
+        context = Robolectric.buildActivity(Activity::class.java).get()
+    }
+
+    @Test
+    fun verifyIsTractDeepLink() {
+        assertFalse(Uri.parse("wss://knowgod.com/en/kgp").isTractDeepLink(context))
+        assertFalse(Uri.parse("https://example.com/en/kgp").isTractDeepLink(context))
+        assertFalse(Uri.parse("https://knowgod.com/en").isTractDeepLink(context))
+        assertTrue(Uri.parse("http://knowgod.com/en/kgp").isTractDeepLink(context))
```
TheNoodleMoose, line 28: We use ```assertTrue``` when we expect the value to be true and vice-versa, right?

TheNoodleMoose (commented):
Based on looking at the test, I think I have an idea of what deferred deep linking is referring to. But could you explain it to me just so I make sure I understand correctly?

library/analytics/src/main/java/org/cru/godtools/analytics/appsflyer/AppsFlyerAnalyticsService.kt
frett: in Kotlin it's the equivalent of `void`

library/analytics/src/main/java/org/cru/godtools/analytics/appsflyer/AppsFlyerAnalyticsService.kt
frett: they got moved a little bit higher in the initialization logic.

basically in order to capture AppsFlyer deep links tracking has to be started, so we can't defer startup until we have the marketing cloud id anymore.

All of this will probably end up changing as we move from Adobe to Google Analytics

library/analytics/src/debug/AndroidManifest.xml
frett: AppsFlyer itself doesn't, we are adding these values to the `MainActivity` that is defined in the `app` module. there is an AndroidManifest merge process that combines the manifest from all included libraries to generate the final manifest.

the reason I put the filters in the `analytics` module instead of directly in the `app` module was to try and keep all AppsFlyer related logic within the `analytics` module if possible.

library/analytics/src/main/java/org/cru/godtools/analytics/appsflyer/AppsFlyerModule.kt
frett: yes, it's our DI (Dependency Injection) framework.

https://dagger.dev/dev-guide/multibindings.html

frett: DI & Dagger are a pretty deep topic, for now you can skip over Dagger stuff in you want.

library/analytics/src/test/java/org/cru/godtools/analytics/appsflyer/AppsFlyerAnalyticsServiceTest.kt
frett: the `@Test` annotation indicates that a function is a test entry point for JUnit 4. so every function/method that is annotated with `@Test` will be run by the unit test framework and will report success or any failures

frett: all tests live in a separate source set

ui/base-tool/build.gradle
frett: this is actually an unusual dependency and the only time I've done it (so far).

Normally `tract-renderer` depends on `base-tool`, and defining a normal dependency in the opposite way will introduce a dependency cycle that breaks the build.

For my unit tests in `base-tool` I wanted access to a class in `tract-renderer`, so because `test` is a separate build target/source set, I could introduce a test only dependency on the regular `tract-renderer` artifact.

so the dependency chain looks like:
`base-tool (test)` -> `tract-renderer` -> `base-tool`

ui/tract-renderer/src/test/java/org/cru/godtools/tract/util/DeepLinkUtilsTest.kt
frett: yeah, there is a whole set of `assert*()` methods that allow you to test various things. It's part of the JUnit 4 framework

frett:
deferred deep linking refers to a user opening up a GodTools deep link that first takes them to the app store to install GodTools, once they install GodTools and launch it for the first time AppsFlyer kicks in, recognizes that the install was trigger by a specific deep link, then triggers the deep link internally within the app.

library/analytics/src/main/java/org/cru/godtools/analytics/appsflyer/AppsFlyerAnalyticsService.kt
frett: another way to define this function would be:
```kotlin
override fun onActivityStarted(activity: Activity) {
}
```

using the `= Unit` here is basically saying there is no function body for this override.

library/analytics/src/main/java/org/cru/godtools/analytics/appsflyer/AppsFlyerModule.kt
frett: we talked about Dagger & DI, so I'm going to resolve this

ui/base-tool/build.gradle
frett: we talked about source sets and how test has it's own source set, so I'm going to resolve this

ui/tract-renderer/src/test/java/org/cru/godtools/tract/util/DeepLinkUtilsTest.kt
frett: we talked about testing frameworks, once you work on a task that requires some unit tests you'll get experience with this

library/analytics/src/test/java/org/cru/godtools/analytics/appsflyer/AppsFlyerAnalyticsServiceTest.kt
frett: we talked about how source sets and unit testing work on android, so resolving this

gyasistory (approved):

PR #1347, opened 2020-10-03
GT-833 Start converting DownloadManager to Kotlin and adding tests

frett:
This converts the logic that writes attachments and translations to disk after downloading them. This is only partially converted to Kotlin because I want to keep PR size reasonable.

Dependent on:
- [x] #1344 
- [x] #1356


library/initial-content/src/main/java/org/cru/godtools/init/content/task/Tasks.kt
```diff
@@ -64,21 +66,25 @@ internal class Tasks @Inject constructor(
         }
     }
 
-    fun initSystemLanguages() {
+    suspend fun initSystemLanguages() {
```
TheNoodleMoose, line 69: What does ```suspend``` add to this?

library/initial-content/src/main/java/org/cru/godtools/init/content/task/Tasks.kt
```diff
@@ -64,21 +66,25 @@ internal class Tasks @Inject constructor(
...
-                // add all device languages and fallbacks
-                .onEach { downloadManager.addLanguage(it) }
-                .asSequence()
-                // set the first available language as the primary language
-                .firstOrNull { dao.find<Language>(it) != null }?.let { settings.primaryLanguage = it }
-
-            // always add english and bundled languages
-            downloadManager.addLanguage(Locale.ENGLISH)
-            BuildConfig.BUNDLED_LANGUAGES.forEach { downloadManager.addLanguage(LocaleCompat.forLanguageTag(it)) }
+            coroutineScope {
```
TheNoodleMoose, line 74: What is the scope of coroutines?

library/model/src/main/java/org/cru/godtools/model/TranslationFile.kt
```diff
@@ -2,9 +2,9 @@ package org.cru.godtools.model
 
 import org.cru.godtools.model.Base.INVALID_ID
 
-class TranslationFile constructor(
+data class TranslationFile(
```
TheNoodleMoose, line 5: What does ```data``` add to this?

library/model/src/main/java/org/cru/godtools/model/TranslationFile.kt
```diff
@@ -2,9 +2,9 @@ package org.cru.godtools.model
...
     val translationId: Long = INVALID_ID,
-    val filename: String? = null
+    val filename: String
```
TheNoodleMoose, line 7: Can filename no longer default to ```null``` because we always default to English?

library/download-manager/src/main/java/org/cru/godtools/download/manager/GodToolsDownloadManager.kt
```diff
@@ -0,0 +1,363 @@
...
+import org.keynote.godtools.android.db.GodToolsDao
+
+private const val TAG = "GodToolsDownloadManager"
+
+@VisibleForTesting
+internal const val CLEANUP_DELAY = HOUR_IN_MS
+@VisibleForTesting
+internal const val CLEANUP_DELAY_INITIAL = MIN_IN_MS
+
+open class KotlinGodToolsDownloadManager @VisibleForTesting internal constructor(
```
TheNoodleMoose, line 71: I've looked over this and I think I have an idea of what's going on, just not sure how it's all working if that makes sense haha.

library/download-manager/src/main/java/org/cru/godtools/download/manager/GodToolsDownloadManager.java
```diff
@@ -42,96 +31,65 @@
...
 
 import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
-import static org.ccci.gto.android.common.base.TimeConstants.HOUR_IN_MS;
 import static org.ccci.gto.android.common.db.Expression.NULL;
 import static org.ccci.gto.android.common.util.ThreadUtils.getLock;
 
 @Singleton
-public final class GodToolsDownloadManager {
+@SuppressLint("VisibleForTests")
+public final class GodToolsDownloadManager extends KotlinGodToolsDownloadManager {
```
TheNoodleMoose, line 63: Is there a reason this has to stick around since we converted most of the functionality over to KotlinGodToolsDownloadManager?

library/download-manager/src/main/java/org/cru/godtools/download/manager/DownloadProgressLiveData.kt
```diff
@@ -3,7 +3,7 @@ package org.cru.godtools.download.manager
 import androidx.annotation.MainThread
 import androidx.lifecycle.MutableLiveData
 
-class DownloadProgressLiveData : MutableLiveData<DownloadProgress?>() {
+internal class DownloadProgressLiveData : MutableLiveData<DownloadProgress?>() {
```
TheNoodleMoose, line 6: Similar question to my others, what does ```internal``` add to this?

library/db/src/main/java/org/keynote/godtools/android/db/GodToolsDao.kt
```diff
@@ -87,7 +87,7 @@ class GodToolsDao @Inject internal constructor(database: GodToolsDatabase) : Abs
 
     public override fun getPrimaryKeyWhere(obj: Any) = when (obj) {
         is LocalFile -> getPrimaryKeyWhere(LocalFile::class.java, obj.filename)
-        is TranslationFile -> getPrimaryKeyWhere(TranslationFile::class.java, obj.translationId, obj.filename!!)
+        is TranslationFile -> getPrimaryKeyWhere(TranslationFile::class.java, obj.translationId, obj.filename)
```
TheNoodleMoose, line 90: What did the ```!!``` do that we no longer need?

library/initial-content/src/main/java/org/cru/godtools/init/content/task/Tasks.kt
frett: `suspend` marks a function as being usable within a coroutine only. This gives it the capability to call other `suspend` functions which could "suspend" execution of the coroutine until a specific state is met.

library/initial-content/src/main/java/org/cru/godtools/init/content/task/Tasks.kt
frett: `coroutineScope` indicates that any coroutines launched within the scope need to complete before the scope finishes

library/model/src/main/java/org/cru/godtools/model/TranslationFile.kt
frett: data classes automatically generate an `equals`, `hashCode`, and `toString` methods for this class

library/model/src/main/java/org/cru/godtools/model/TranslationFile.kt
frett: the `TranslationFile` object tracks the fact that a specific translation depends on a file on disk. `filename` being null makes no sense in that object definition

library/download-manager/src/main/java/org/cru/godtools/download/manager/GodToolsDownloadManager.java
frett: still in process of converting the rest of the functionality. My general practice when reworking functionality is to try and keep everything functional in the process so that I can trace bugs back to where they were introduced if necessary.

library/download-manager/src/main/java/org/cru/godtools/download/manager/DownloadProgressLiveData.kt
frett: `internal` is a Kotlin visibility modifier. in this case it is restricting visibility of the `DownloadProgressLiveData` class to be visible within the `download-manager` module only.

the class it extends is still externally visible, but making this `internal` makes sure we only use this directly within the `download-manager`

library/db/src/main/java/org/keynote/godtools/android/db/GodToolsDao.kt
frett: `!!` forces a value to be non-null, or it will throw an exception if it was null. now that `TranslationFile` can no longer have a null `filename`, there is no reason to force it to be non-null here.

gyasistory (approved):

PR #1351, opened 2020-10-05
GT-954-Add-Lottie-Animations-For-General-Onboarding

TheNoodleMoose:


ui/tutorial-renderer/src/main/res/values/dimens.xml
```diff
@@ -5,4 +5,5 @@
     <dimen name="tutorial_page_inset_top">48dp</dimen>
     <dimen name="tutorial_page_inset_bottom">64dp</dimen>
     <dimen name="tutorial_page_live_share_anim_height">266dp</dimen>
+    <dimen name="tutorial_page_onboarding_anim_height">140dp</dimen>
```
TheNoodleMoose, line 8: I turned on that ```show layout bounds``` mode you suggested and used that to find what looked right for this.

ui/tutorial-renderer/src/main/res/layout/tutorial_onboarding_ready.xml
```diff
@@ -33,16 +33,18 @@
             app:layout_constraintTop_toTopOf="parent"
             app:srcCompat="@drawable/bkg_tutorial_onboarding_wave_3" />
 
-        <ImageView
+        <com.airbnb.lottie.LottieAnimationView
             android:id="@+id/image"
```
TheNoodleMoose, line 37: I left this as image because if I changed it to ```animation``` I noticed a significant styling difference that I don't believe is desired.

ui/tutorial-renderer/src/main/res/values/dimens.xml
frett: before it was relying on the implicit size of the images, which had a height of 140. so you found the right height.

frett: the scales from `dp` -> `px` are:
`mdpi`: 1x
`hdpi`: 1.5x
`xhdpi`: 2x
`xxhdpi`: 3x

frett: so, looking at the dimensions of a mdpi image will reveal what size it is in DP

ui/tutorial-renderer/src/main/res/layout/tutorial_onboarding_ready.xml
frett: yup, the positioning layout references the image view directly.

frett (approved):

ui/tutorial-renderer/src/main/res/values/dimens.xml
frett: only thing I would have done slightly different with this constant, is put it in the `styles_tutorial_onboarding.xml` file where the other onboarding styles are located

frett:
taking a look at the layouts with layout bounds enabled I actually see that the current constraints aren't providing the exact positioning we were looking for, but that's something I'll address in a different PR, and it isn't related to the animations at all.

ui/tutorial-renderer/src/main/res/values/dimens.xml
TheNoodleMoose: Would it be worth moving there?

frett: yeah, just move it to the beginning of this file: https://github.com/CruGlobal/godtools-android/blob/develop/ui/tutorial-renderer/src/main/res/values/styles_tutorial_onboarding.xml

frett: it provides a bit of organization long term to prevent having to search the code base for things :)

TheNoodleMoose: Maybe a dumb question, but should it be inside the style tag or outside?

frett: outside the `<style>` tag, but still inside the `<resources>` tag, you're still learning, so not dumb.

all the xml files within the `values` folder are treated the same, so you can mix and match resources in any file if you want to. It's really just conventions that dictate where things end up

TheNoodleMoose: Makes sense thanks, Daniel! 🙂

PR #1360, opened 2020-10-08
GT-960 Tips tutorial frequency

frett:
Show the tips tutorial 3 times per tool

### DIFF
--- a/app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
+++ b/app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
@@ -133,7 +133,8 @@ class ToolDetailsFragment() : BasePlatformFragment<ToolDetailsFragmentBinding>(R
 
         activity?.run {
             openToolActivity(code, tool.type, locale, showTips = true)
-            if (!settings.isFeatureDiscovered(FEATURE_TUTORIAL_TIPS)) {
+            if (settings.getFeatureDiscoveredCount("$FEATURE_TUTORIAL_TIPS$code") < 3) {
+                settings.setFeatureDiscovered("$FEATURE_TUTORIAL_TIPS$code")
                 startTutorialActivity(
                     PageSet.TIPS,
                     Bundle().apply { putString(FORMAT_ARG_TOOL_NAME, translation.getName(tool).toString()) }
--- a/library/base/src/main/java/org/cru/godtools/base/Settings.kt
+++ b/library/base/src/main/java/org/cru/godtools/base/Settings.kt
@@ -51,7 +51,7 @@ class Settings @Inject internal constructor(
         const val FEATURE_TUTORIAL_ONBOARDING = "tutorialOnboarding"
         const val FEATURE_TUTORIAL_TRAINING = "tutorialTraining"
         const val FEATURE_TUTORIAL_LIVE_SHARE = "tutorialLiveShare."
-        const val FEATURE_TUTORIAL_TIPS = "tutorialTips"
+        const val FEATURE_TUTORIAL_TIPS = "tutorialTips."
 
         @JvmStatic
         val defaultLanguage: Locale get() = Locale.ENGLISH
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
@@ -79,7 +79,7 @@ class TutorialActivity : BaseActivity<TutorialActivityBinding>(), TutorialCallba
     }
 
     override fun onOptionsItemSelected(item: MenuItem) = when (item.itemId) {
-        R.id.onboarding_action_skip -> {
+        R.id.onboarding_action_skip, R.id.action_tips_skip -> {
             finish()
             true
         }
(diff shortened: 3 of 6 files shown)


library/base/src/main/java/org/cru/godtools/base/Settings.kt
```diff
@@ -51,7 +51,7 @@ class Settings @Inject internal constructor(
         const val FEATURE_TUTORIAL_ONBOARDING = "tutorialOnboarding"
         const val FEATURE_TUTORIAL_TRAINING = "tutorialTraining"
         const val FEATURE_TUTORIAL_LIVE_SHARE = "tutorialLiveShare."
-        const val FEATURE_TUTORIAL_TIPS = "tutorialTips"
+        const val FEATURE_TUTORIAL_TIPS = "tutorialTips."
```
TheNoodleMoose, line 54: Why did you add a period here?

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -133,7 +133,8 @@ class ToolDetailsFragment() : BasePlatformFragment<ToolDetailsFragmentBinding>(R
 
         activity?.run {
             openToolActivity(code, tool.type, locale, showTips = true)
-            if (!settings.isFeatureDiscovered(FEATURE_TUTORIAL_TIPS)) {
+            if (settings.getFeatureDiscoveredCount("$FEATURE_TUTORIAL_TIPS$code") < 3) {
+                settings.setFeatureDiscovered("$FEATURE_TUTORIAL_TIPS$code")
```
TheNoodleMoose, line 137: I assume that ```settings.getFeatureDiscoveredCount``` returns an Int value for how many times the tutorial tips have been shown for a specific tool. And we decide which tool to check by passing ```$code``` in the function as an argument. Where are those values stored?

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
```diff
@@ -79,7 +79,7 @@ class TutorialActivity : BaseActivity<TutorialActivityBinding>(), TutorialCallba
     }
 
     override fun onOptionsItemSelected(item: MenuItem) = when (item.itemId) {
-        R.id.onboarding_action_skip -> {
+        R.id.onboarding_action_skip, R.id.action_tips_skip -> {
```
TheNoodleMoose, line 82: I assume skip logic is already implemented, and this is actually just letting that logic actually run once the skip menu item is pressed for tutorial tips. So is it safe to assume if you didn't add this, and I tried pressing skip that it would do nothing?

ui/tutorial-renderer/src/main/res/menu/tutorial_tips_menu.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <item
+        android:id="@+id/action_tips_skip"
+        android:title="@string/tutorial_tips_action_skip"
+        app:showAsAction="always" />
```
TheNoodleMoose, line 8: Does this attribute change styling in some way?

TheNoodleMoose (approved):

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
frett: yes, the `Settings` object has all the feature discovery logic. using Flipper you can look at the underlying SharedPreferences that actually stores the raw data.
Screen Shot 2020-10-08 at 11 37 03 AM

frett: `$code` is the tool code (e.g. `kgp`, `fourlaws`)

library/base/src/main/java/org/cru/godtools/base/Settings.kt
frett: because I'm concatenating the tool code to the end of the key. so the key will look like: `tutorialTips.kgp`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
frett: yeah, I opted to provide unique ids to the menu items for different tutorials that way if the logic ever needs to diverge I'm ready to handle the unique case for tips vs onboarding

ui/tutorial-renderer/src/main/res/menu/tutorial_tips_menu.xml
frett: so, this is menu XML, it controls what menus items are present in the toolbar. by default the toolbar decides how many items to show directly as actions, and the rest get put into the action overflow menu. this attribute forces this action to always be shown directly and never be put in the action overflow menu

TheNoodleMoose: Ah okay makes sense!

library/base/src/main/java/org/cru/godtools/base/Settings.kt
TheNoodleMoose: Oh, gotcha!

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
TheNoodleMoose: I was able to get Flipper connected and find the stored raw data, thanks!

PR #1366, opened 2020-10-13
Revert "Add new lottie animation"

frett:
This reverts commit dbf8d2446e9730d9a679d031a94305036fb6c1fe.


frett:
There is an issue with one of the emojis in the animation, so Ann wants us to revert back to the static image for release, then we will fix the animation later

TheNoodleMoose (approved):

PR #1384, opened 2020-10-26
GT-853-Track-Shortcut-Launches

TheNoodleMoose:
This is a pretty basic implementation, in the future I'd imagine we'd like to know the tool that was launched maybe? Let me know if you have a better naming convention for the event as well. I went with "tool-opened-shortcut" but I was considering something like "tool-opened-via-shortcut". I think either gets the point across.

### DIFF
--- a/ui/base-tool/src/main/java/org/cru/godtools/base/tool/Activities.kt
+++ b/ui/base-tool/src/main/java/org/cru/godtools/base/tool/Activities.kt
@@ -14,7 +14,7 @@ private const val ACTIVITY_CLASS_TRACT = "org.cru.godtools.tract.activity.TractA
 
 const val EXTRA_SHOW_TIPS = "org.cru.godtools.tract.activity.TractActivity.SHOW_TIPS"
 
-const val SHORTCUT_LAUNCH = "org.cru.goodtools.tract.activity.TractActivity.SHORTCUT_LAUNCH"
+const val SHORTCUT_LAUNCH = "org.cru.godtools.tool.SHORTCUT_LAUNCH"
 
 fun Activity.startTractActivity(toolCode: String, vararg languages: Locale?, showTips: Boolean) =
     startActivity(createTractActivityIntent(toolCode, *languages, showTips = showTips))
--- a/ui/base-tool/src/main/java/org/cru/godtools/base/tool/Activities.kt
+++ b/ui/base-tool/src/main/java/org/cru/godtools/base/tool/Activities.kt
@@ -14,6 +14,8 @@ private const val ACTIVITY_CLASS_TRACT = "org.cru.godtools.tract.activity.TractA
 
 const val EXTRA_SHOW_TIPS = "org.cru.godtools.tract.activity.TractActivity.SHOW_TIPS"
 
+const val SHORTCUT_LAUNCH = "org.cru.goodtools.tract.activity.TractActivity.SHORTCUT_LAUNCH"
+
 fun Activity.startTractActivity(toolCode: String, vararg languages: Locale?, showTips: Boolean) =
     startActivity(createTractActivityIntent(toolCode, *languages, showTips = showTips))
 
--- a/ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/BaseToolActivity.kt
+++ b/ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/BaseToolActivity.kt
@@ -29,9 +29,11 @@ import org.cru.godtools.base.Settings.Companion.FEATURE_TOOL_SHARE
 import org.cru.godtools.base.tool.BR
 import org.cru.godtools.base.tool.BaseToolRendererModule.IS_CONNECTED_LIVE_DATA
 import org.cru.godtools.base.tool.R
+import org.cru.godtools.base.tool.SHORTCUT_LAUNCH
 import org.cru.godtools.base.tool.analytics.model.FirstToolOpenedAnalyticsActionEvent
 import org.cru.godtools.base.tool.analytics.model.ShareActionEvent
 import org.cru.godtools.base.tool.analytics.model.ToolOpenedAnalyticsActionEvent
+import org.cru.godtools.base.tool.analytics.model.ToolOpenedViaShortcutAnalyticsActionEvent
 import org.cru.godtools.base.tool.databinding.ToolGenericFragmentActivityBinding
 import org.cru.godtools.base.tool.model.view.getTypeface
 import org.cru.godtools.base.ui.activity.BaseActivity
@@ -54,11 +56,18 @@ abstract class BaseToolActivity<B : ViewDataBinding>(@LayoutRes contentLayoutId:
     @Inject
     protected lateinit var downloadManager: GodToolsDownloadManager
 
+    private val shortcutLaunch get() = intent?.getBooleanExtra(SHORTCUT_LAUNCH, false) ?: false
+
     // region Lifecycle
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         isConnected.observe(this) { if (it) syncTools() }
         setupStatusBar()
+
+        // track if activity was launched by a shortcut
+        if(savedInstanceState == null && shortcutLaunch) {
+            eventBus.post(ToolOpenedViaShortcutAnalyticsActionEvent)
+        }
     }
 
     @CallSuper
--- a/ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/BaseToolActivity.kt
+++ b/ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/BaseToolActivity.kt
@@ -65,7 +65,7 @@ abstract class BaseToolActivity<B : ViewDataBinding>(@LayoutRes contentLayoutId:
         setupStatusBar()
 
         // track if activity was launched by a shortcut
-        if(savedInstanceState == null && shortcutLaunch) {
+        if (savedInstanceState == null && shortcutLaunch) {
             eventBus.post(ToolOpenedViaShortcutAnalyticsActionEvent)
         }
     }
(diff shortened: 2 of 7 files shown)


ui/base-tool/src/main/java/org/cru/godtools/base/tool/analytics/model/OpenToolsAnalyticsActionEvent.kt
```diff
@@ -6,3 +6,4 @@ import org.cru.godtools.analytics.model.AnalyticsSystem
 object ToolOpenedAnalyticsActionEvent : AnalyticsActionEvent("tool-opened", system = AnalyticsSystem.APPSFLYER)
 object FirstToolOpenedAnalyticsActionEvent :
     AnalyticsActionEvent("first-tool-opened", system = AnalyticsSystem.APPSFLYER)
+object ToolOpenedViaShortcutAnalyticsActionEvent : AnalyticsActionEvent("tool-opened-shortcut", system = AnalyticsSystem.APPSFLYER)
```
frett, line 9: the `system = APPSFLYER` parameter limits this event to appsflyer only. for now let's change it to be Adobe only.

We will revisit it when we do the adobe -> google 360 migration in a couple weeks

frett: also, let's move this event object to it's own source file

TheNoodleMoose: Okay sounds good.

ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
```diff
@@ -273,7 +274,7 @@ class GodToolsShortcutManager @Inject constructor(
             Tool.Type.TRACT -> context.createTractActivityIntent(code, *locales.toTypedArray())
             Tool.Type.ARTICLE -> context.createCategoriesIntent(code, locales[0])
             else -> return@withContext null
-        }.apply { action = Intent.ACTION_VIEW }
+        }.apply { action = Intent.ACTION_VIEW; putExtra(SHORTCUT_LAUNCH, true) }
```
frett, line 277: putting multiple statements on a single line is something that in general is a Java/Kotlin anti-pattern.

Taking a second look at the surrounding logic, let's move the adjustments to the intent out of the apply block and just execute them as normal statements. e.g. `intent.action = Intent.ACTION_VIEW`

ui/base-tool/src/main/java/org/cru/godtools/base/tool/Activities.kt
```diff
@@ -14,6 +14,8 @@ private const val ACTIVITY_CLASS_TRACT = "org.cru.godtools.tract.activity.TractA
 
 const val EXTRA_SHOW_TIPS = "org.cru.godtools.tract.activity.TractActivity.SHOW_TIPS"
 
+const val SHORTCUT_LAUNCH = "org.cru.goodtools.tract.activity.TractActivity.SHORTCUT_LAUNCH"
```
frett, line 17: minor nit here, let's make the extra name a bit more generic like `org.cru.godtools.tool.SHORTCUT_LAUNCH`

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.kt
```diff
@@ -62,12 +64,7 @@ import org.cru.godtools.tract.util.ViewUtils
 import org.cru.godtools.tract.util.isTractDeepLink
 import org.cru.godtools.tutorial.PageSet
 import org.cru.godtools.tutorial.activity.buildTutorialActivityIntent
-import org.cru.godtools.xml.model.Card
-import org.cru.godtools.xml.model.Modal
-import org.cru.godtools.xml.model.Page
-import org.cru.godtools.xml.model.backgroundColor
-import org.cru.godtools.xml.model.navBarColor
-import org.cru.godtools.xml.model.navBarControlColor
+import org.cru.godtools.xml.model.*
```
frett, line 67: the wildcard import will trigger a ktlint violation, you'll have to adjust your Android Studio code style settings to prevent it from automatically happening

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.kt
```diff
@@ -104,6 +103,11 @@ class TractActivity : BaseToolActivity<TractActivityBinding>(R.layout.tract_acti
         // track this view
         if (savedInstanceState == null) dataModel.tool.value?.let { trackToolOpen(it) }
 
+        // track if activity was launched by a shortcut
+        if(shortCutLaunch) {
+            eventBus.post(ToolOpenedViaShortcutAnalyticsActionEvent)
+        }
```
frett, line 109: let's move the shortcut launch detection logic to the `BaseToolActivity`, that way it will apply to articles shortcuts as well

ui/base-tool/src/main/java/org/cru/godtools/base/tool/Activities.kt
frett: also, you misspelled godtools originally ;)

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.kt
frett: in addition, Android can recreate activities multiple times while it is being used. The way the logic is currently written this event will fire every time the activity is re-created.

the standard work around pattern we use for that is checking to see if there is `savedInstanceState` present or not

ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
TheNoodleMoose: @frett is this the correct way to do it? 
Screen Shot 2020-10-26 at 4 25 41 PM

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.kt
TheNoodleMoose: Okay I went ahead and made a change so its:
```
 if(shortCutLaunch && savedInstanceState == null) {
            eventBus.post(ToolOpenedViaShortcutAnalyticsActionEvent)
}
```
should that do the trick?

ui/base-tool/src/main/java/org/cru/godtools/base/tool/Activities.kt
TheNoodleMoose: haha that I did! Thanks for catching that 🙂

ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
frett: yup, I would get rid of the extra line return between creating the intent and updating the action/extras. 

The rationale being that all 3 of the statements deal with initializing the intent, so they can be kept together so that it's is easier to scan over the code and see a block of code that creates the appropriate intent.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.kt
frett: yeah, that works good.

A micro optimization is to reverse the order of the 2 conditionals, checking if `savedInstanceState` is null is quicker than loading a value out of the intent to see if it's true or false. boolean operations short-circuit when possible, so it would see that `savedInstanceState` isn't null, at which point it wouldn't even bother checking the intent for the flag.

TheNoodleMoose: Makes sense! I'll switch the two.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.kt
```diff
@@ -38,7 +38,9 @@ import org.cru.godtools.base.Settings.Companion.FEATURE_TUTORIAL_LIVE_SHARE
 import org.cru.godtools.base.URI_SHARE_BASE
 import org.cru.godtools.base.model.Event
 import org.cru.godtools.base.tool.EXTRA_SHOW_TIPS
+import org.cru.godtools.base.tool.SHORTCUT_LAUNCH
 import org.cru.godtools.base.tool.activity.BaseToolActivity
+import org.cru.godtools.base.tool.analytics.model.ToolOpenedViaShortcutAnalyticsActionEvent
```
frett, line 43: ktlint will complain about these unused imports. there is a code auto-format task within Android Studio that should deal with the majority of ktlint violations

ui/base-tool/src/main/java/org/cru/godtools/base/tool/activity/BaseToolActivity.kt
```diff
@@ -54,11 +56,18 @@ abstract class BaseToolActivity<B : ViewDataBinding>(@LayoutRes contentLayoutId:
...
+    private val shortcutLaunch get() = intent?.getBooleanExtra(SHORTCUT_LAUNCH, false) ?: false
+
     // region Lifecycle
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         isConnected.observe(this) { if (it) syncTools() }
         setupStatusBar()
+
+        // track if activity was launched by a shortcut
+        if(savedInstanceState == null && shortcutLaunch) {
```
frett, line 68: ktlint will complain about not having a space between `if` and `(`

frett (approved):

PR #1396, opened 2020-10-30
GT-952-Start-Of-Migrating-To-GA

TheNoodleMoose:
Still waiting on the GTM container, but figured I could have you look over what I have and see if I missed anything.


library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -80,7 +80,11 @@ class FirebaseAnalyticsService @MainThread @Inject internal constructor(
 
     @MainThread
     private fun handleScreenEvent(event: AnalyticsScreenEvent) {
-        currentActivity?.let { firebase.setCurrentScreen(it, event.screen, null) }
+        val bundle = Bundle()
+        bundle.putString(FirebaseAnalytics.Param.SCREEN_NAME, event.screen)
+        currentActivity?.let {
+            firebase.logEvent(FirebaseAnalytics.Event.SCREEN_VIEW, bundle)
+        }
```
TheNoodleMoose, line 87: I don't know if this is the best way to do this, but this is how they did it in the documentation to migrating from ```setScreenName```. [Here's the doc](https://firebase.googleblog.com/2020/08/google-analytics-manual-screen-view.html).

frett: This can be simplified a bit, you don't actually use`currentActivity` anymore.

frett: And following that up, is`currentActivity` used elsewhere?

TheNoodleMoose: Here are the only other spots where it is used: https://github.com/CruGlobal/godtools-android/blob/45d4bdb9967c77007dfb781fee7e4599edd901c0/library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt#L58-L66

frett: so those callbacks set/clear `currentActivity`, which is no longer necessary since you aren't using `currentActivity`. those calls can be removed, and then the `ActivityLifecycleCallbacks` interface can be removed from `FirebaseAnalyticsService` because it no longer does anything.

frett (approved):
approving this for what is here currently. if you are going to add GTM to this PR then I'll review again

TheNoodleMoose:
@frett Can you check my latest commit fixing the merge conflicts to make sure I merged correctly?

frett:
@TheNoodleMoose the `FirebaseAnalyticsService` looks good to me

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -18,6 +16,7 @@ import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onEach
 import org.ccci.gto.android.common.okta.oidc.OktaUserProfileProvider
 import org.ccci.gto.android.common.okta.oidc.net.response.ssoGuid
+import org.ccci.gto.android.common.compat.util.LocaleCompat
```
frett, line 19: ktlint is complaining that the import isn't properly sorted. you should be able to select the imports and use the Android Studio re-formatting to automatically fix this

TheNoodleMoose: Okay I'll go ahead and do that.

frett (approved):
Everything looks good

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -54,38 +55,21 @@ class FirebaseAnalyticsService @VisibleForTesting internal constructor(
...
-
     @MainThread
     private fun handleScreenEvent(event: AnalyticsScreenEvent) {
-        currentActivity?.let { firebase.setCurrentScreen(it, event.screen, null) }
+        val bundle = Bundle()
+        bundle.putString(FirebaseAnalytics.Param.SCREEN_NAME, event.screen)
+        bundle.putString("cru_appname", VALUE_GODTOOLS)
+        event.locale?.let { bundle.putString("cru_contentlanguage", LocaleCompat.toLanguageTag(it)) }
+        event.adobeSiteSection?.let { bundle.putString("cru_sitesection", event.adobeSiteSection) }
+        event.adobeSiteSubSection?.let { bundle.putString("cru_sitesubsection", event.adobeSiteSubSection) }
```
frett, line 71: another Kotlin idiom that would work here is something like:
```kotlin
val bundle = Bundle().apply {
    putString(FirebaseAnalytics.Param.SCREEN_NAME, event.screen)
    putString("cru_appname", VALUE_GODTOOLS)
    event.locale?.let { putString("cru_contentlanguage", LocaleCompat.toLanguageTag(it)) }
    event.adobeSiteSection?.let { putString("cru_sitesection", event.adobeSiteSection) }
    event.adobeSiteSubSection?.let { putString("cru_sitesubsection", event.adobeSiteSubSection) }
}
```

frett: this idiom makes it clear that you are configuring the bundle with everything in the `apply` block

TheNoodleMoose: Got it, will go ahead and change it.

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -54,43 +55,34 @@ class FirebaseAnalyticsService @VisibleForTesting internal constructor(
...
+            event.adobeSiteSubSection?.let { putString("cru_sitesubsection", event.adobeSiteSubSection) }
+        }
+        firebase.logEvent(FirebaseAnalytics.Event.SCREEN_VIEW, bundle)
     }
 
     @MainThread
     private fun handleActionEvent(event: AnalyticsActionEvent) {
-        firebase.logEvent(event.firebaseEventName, null)
+        val bundle = Bundle().apply {
+            event.adobeAttributes?.forEach { attribute ->
```
frett, line 79: I'd like us to move away from adobe specific properties on the event objects. So we should probably create a `firebaseAttributes` property on the event object that exposes the attributes in the correct format.

frett: hmm, `adobeSiteSection` and `adobeSiteSubSection` would also fall under the renaming need. This can be done in a followup PR if you would like

TheNoodleMoose: So should I leave the adobeAttributes as is, or put in the effort now to change it? It could also be done with all the adobe clean work that will eventually come.

frett: let's do it with all the adobe cleanup work later

frett: when we do that work I will want to get rid of the replace calls at that time

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -54,50 +56,41 @@ class FirebaseAnalyticsService @VisibleForTesting internal constructor(
...
+
+        firebase.logEvent(event.firebaseEventName, bundle)
     }
 
     init {
         oktaUserProfileProvider.userInfoFlow()
             .map { it?.ssoGuid }
             .distinctUntilChanged()
-            .onEach { firebase.setUserId(it) }
+            .onEach { firebase.setUserId(it); firebase.setUserProperty("debug", BuildConfig.DEBUG.toString()) }
```
frett, line 93: a couple things here.
1. try and avoid having multiple statements on a single line, you should have added line returns so that the 2 statements were individual lines.
2. you should just set the `debug` property directly in the init block, not every time the logged in user changes. See how `USER_PROP_APP_TYPE` is set a single time

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -54,50 +56,41 @@ class FirebaseAnalyticsService @VisibleForTesting internal constructor(
...
-    override fun onActivityDestroyed(activity: Activity) = Unit
-    // endregion ActivityLifecycleCallbacks
-    // endregion Tracking Events
-
     @MainThread
     private fun handleScreenEvent(event: AnalyticsScreenEvent) {
-        currentActivity?.let { firebase.setCurrentScreen(it, event.screen, null) }
+        val bundle = Bundle().apply {
+            putString(FirebaseAnalytics.Param.SCREEN_NAME, event.screen)
+            putString("cru_appname", VALUE_GODTOOLS)
```
frett, line 69: the new strings you introduced for attribute/user property names should be made constants.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/analytics/model/ContentAnalyticsActionEvent.kt
```diff
@@ -14,8 +14,8 @@ class ContentAnalyticsActionEvent(@VisibleForTesting internal val event: Analyti
     override val adobeAttributes: Map<String, *>? get() = event.attributes
     override val firebaseEventName = if (isForSystem(AnalyticsSystem.ADOBE)) {
         Timber.tag(TAG).e(UnsupportedOperationException("XML Adobe Analytics Usage"), "action ${event.action}")
-        event.action?.replace(Regex("[ \\-]"), "_").orEmpty()
+        event.action?.replace(Regex("[ \\-]"), "_").orEmpty().toLowerCase()
```
TheNoodleMoose, line 17: Clark requested that all the events are lowercased for consistency so I went ahead and did that here.

frett: alright, this is obviously a temporary solution, we will need to start renaming the events when we perform the analytics cleanup phase

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -86,15 +110,58 @@ class FirebaseAnalyticsService @VisibleForTesting internal constructor(
...
+    private fun updateVisitorIdIdentifiers(userProfile: UserInfo?) {
+        Visitor.syncIdentifiers(
+            mutableMapOf(
+                VISITOR_ID_GUID to userProfile?.ssoGuid,
+                VISITOR_ID_MASTER_PERSON_ID to userProfile?.grMasterPersonId
+            ), when {
+            userProfile != null -> VisitorID.VisitorIDAuthenticationState.VISITOR_ID_AUTHENTICATION_STATE_AUTHENTICATED
+            else -> VisitorID.VisitorIDAuthenticationState.VISITOR_ID_AUTHENTICATION_STATE_LOGGED_OUT
+        }
+        )
```
frett, line 140: this is an Adobe analytics API, you shouldn't be using it here

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -86,15 +110,58 @@ class FirebaseAnalyticsService @VisibleForTesting internal constructor(
...
             .map { it?.ssoGuid }
             .distinctUntilChanged()
-            .onEach { firebase.setUserId(it); firebase.setUserProperty("debug", BuildConfig.DEBUG.toString()) }
+            .onEach { it ->
+                firebase.setUserId(it)
+                analyticsExecutor.execute {
+                    updateVisitorIdIdentifiers(userProfileStateFlow.value)
+                    val data = buildMap {
+                        baseContextData(userProfileStateFlow.value)
+                    }
```
frett, line 154: there isn't a point in building out an attribute map in `baseContextData` to immediately just set all the key/values as `UserProperties`. you should just directly set the user properties

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -86,15 +110,58 @@ class FirebaseAnalyticsService @VisibleForTesting internal constructor(
...
         oktaUserProfileProvider.userInfoFlow()
             .map { it?.ssoGuid }
             .distinctUntilChanged()
-            .onEach { firebase.setUserId(it); firebase.setUserProperty("debug", BuildConfig.DEBUG.toString()) }
+            .onEach { it ->
+                firebase.setUserId(it)
+                analyticsExecutor.execute {
+                    updateVisitorIdIdentifiers(userProfileStateFlow.value)
+                    val data = buildMap {
+                        baseContextData(userProfileStateFlow.value)
```
frett, line 153: you actually don't need `userProfileStateFlow` here because you already have the current `UserInfo` object in `it`

frett: err, scratch that. you will need to rework a little bit of the previous 2 flow transforms so that you have the user profile in the `onEach`

frett: I would drop the `distinctUntilChanged` call, attributes besides the guid could change without the guid being updated. so you shouldn't be preventing updating user property settings if the guid hasn't changed but another attribute may have

TheNoodleMoose:
@frett It seems the build keeps erroring even after restarting, not sure why though.

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -54,54 +65,51 @@ class FirebaseAnalyticsService @VisibleForTesting internal constructor(
...
 
     init {
-        oktaUserProfileProvider.userInfoFlow()
-            .map { it?.ssoGuid }
-            .distinctUntilChanged()
-            .onEach { firebase.setUserId(it) }
+        oktaUserProfileProvider.userInfoFlow(refreshIfStale = false)
+            .onEach { it ->
+                firebase.setUserProperty(USER_PROP_LOGGED_IN_STATUS, "${it != null}")
+                firebase.setUserProperty(USER_PROP_APP_NAME, VALUE_GODTOOLS)
```
frett, line 102: with setting the app name here, do you also need to set it on [line 78](https://github.com/CruGlobal/godtools-android/pull/1396/files#diff-16cc0dd8554c9cc5ad5d84d969fb2c391765505cce3113ebc3537609ffe13763R78)?

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -54,54 +65,51 @@ class FirebaseAnalyticsService @VisibleForTesting internal constructor(
...
+                firebase.setUserProperty(USER_PROP_CONTENT_LANGUAGE, LocaleCompat.toLanguageTag(Locale.getDefault()))
```
frett, line 103: is this meant to be the device language or the user's selected primary language?

TheNoodleMoose: I believe the device, that's what we did for adobe.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/analytics/model/ContentAnalyticsActionEvent.kt
```diff
@@ -4,9 +4,18 @@ import androidx.annotation.VisibleForTesting
 import org.cru.godtools.analytics.model.AnalyticsActionEvent
 import org.cru.godtools.analytics.model.AnalyticsSystem
 import org.cru.godtools.xml.model.AnalyticsEvent
+import timber.log.Timber
+
+const val TAG = "ContentAnalyticsEvent"
```
frett, line 9: make this `private`

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
TheNoodleMoose: That's what they requested, but I'm confirming that they realize we are sending it twice.

TheNoodleMoose: They said just to leave it for now.

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
frett: actually we never had the device language sent for adobe, we were only sending the content language from within individual events. (that's already in place [here](https://github.com/CruGlobal/godtools-android/pull/1396/files#diff-16cc0dd8554c9cc5ad5d84d969fb2c391765505cce3113ebc3537609ffe13763R79)).

I won't block on this because it is different data, it just happened to be new data vs what we had before. I know this would be useful for MPDX/MH, but not nearly as useful in GodTools since most of the app is actually driven by the primary language and not the device language

TheNoodleMoose: Oh, you're right, this is how it's done for MPDx. Yeah, we'll see if they end up using it and if not we can remove it in the clean-up. But I agree I don't see the need either.

frett (approved):
I'm good with merging this since it's passed QA. any additional cleanup will be in subsequent PRs

library/analytics/src/test/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsServiceTest.kt
```diff
@@ -43,7 +43,7 @@ class FirebaseAnalyticsServiceTest {
         eventBus = mock()
         firebase = mock()
         oktaUserProfileProvider = mock {
-            on { userInfoFlow() } doReturn userInfoChannel.consumeAsFlow()
+            on { userInfoFlow(refreshIfStale = any()) } doReturn userInfoChannel.consumeAsFlow()
```
TheNoodleMoose, line 46: Ah, this makes sense, nice catch!

frett: yeah, would have been nice if the test had failed instead of just hung. one drawback to testing coroutines is that a lot of time you are waiting for code to execute in the background.

frett: in this case without the proper mock statement the test channel was never attached to the `userInfoFlow()` within `FirebaseAnalyticsService`. This in turn meant the test values i was pushing into the channel were never delivered and state was never updated.

PR #1415, opened 2020-11-17
GT-893 disable the login menu items before Lollipop

frett:
The Okta SDK doesn't correctly launch the Okta login flow on KitKat, which leads to a broken user interaction where the user logs in, but the app doesn't receive the correct callback.

This may be fixable, but due to our KitKat user base currently being ~1.8% we are choosing to just disable the functionality for those users.


frett:
in addition the version of the Okta SDK released a week ago dropped KitKat support

frett:
and it looks like the change to drop KitKat support was due to the same issue we are experiencing.

https://github.com/okta/okta-oidc-android/pull/196#issue-508584685

TheNoodleMoose (approved):
@frett Can you explain how changing the file name disables it for certain versions? Does adding that ```v21```define the oldest version supported for that file or something?

frett:
@TheNoodleMoose Yeah, there is a default resources [file](https://github.com/CruGlobal/godtools-android/blob/develop/app/src/main/res/values/misc.xml#L3) that disables the login menu. Then when certain resource restrictions are met (english, at least API 21) it is overridden by the [file](https://github.com/CruGlobal/godtools-android/blob/fe72ff4195460c73f5a697d185f1b2f80674652e/app/src/main/res/values-en-v21/misc.xml#L3) I moved in this PR.

Before this PR login was restricted to english only, this PR adds the additional constraint of at least API 21

PR #1416, opened 2020-11-17
GT-973 Fallback support

frett:


ui/tract-renderer/src/main/java/org/cru/godtools/tract/ui/controller/ParentController.kt
```diff
@@ -37,17 +37,19 @@ abstract class ParentController<T> protected constructor(
     // endregion Lifecycle
 
     // region Child Content
-    protected abstract val contentContainer: LinearLayout
+    protected abstract val contentContainer: ViewGroup
     private val childCache by lazy { UiControllerCache(contentContainer, this) }
     private var children: List<BaseController<Content>>? = null
 
+    protected open val contentToRender get() = model?.content
```
TheNoodleMoose, line 44: Is there value in extracting this to a variable?

ui/tract-renderer/src/main/java/org/cru/godtools/tract/ui/controller/ParentController.kt
```diff
@@ -37,17 +37,19 @@ abstract class ParentController<T> protected constructor(
     // endregion Lifecycle
 
     // region Child Content
-    protected abstract val contentContainer: LinearLayout
+    protected abstract val contentContainer: ViewGroup
```
TheNoodleMoose, line 40: What's the difference between a ViewGroup and the LinearLayout?

library/xml-model/src/test/resources/org/cru/godtools/xml/model/fallback_paragraph.xml
```diff
@@ -0,0 +1,7 @@
+<content:paragraph fallback="true" xmlns:content="https://mobile-content-api.cru.org/xmlns/content">
+    <content:image resource="image.png" restrictTo="web" />
+    <content:image resource="image.png" restrictTo="ios" />
+    <content:text>Test</content:text>
+    <content:text restrictTo="android">Test</content:text>
+    <content:text restrictTo="mobile">Test</content:text>
+</content:paragraph>
```
TheNoodleMoose, line 7: So is this loading the paragraph from the api like all the other content which is just xml?

library/xml-model/src/test/resources/org/cru/godtools/xml/model/fallback.xml
```diff
@@ -0,0 +1,8 @@
+<content:fallback xmlns:content="https://mobile-content-api.cru.org/xmlns/content">
+    <content:image resource="image.png" restrictTo="web" />
+    <content:image resource="image.png" restrictTo="ios" />
+    <content:text>Test</content:text>
+    <content:button type="unrecognized"><content:text>events</content:text></content:button>
+    <content:text restrictTo="android">Test</content:text>
+    <content:text restrictTo="mobile">Test</content:text>
+</content:fallback>
```
TheNoodleMoose, line 8: Similar thing here, or would it just say ```events``` and ```Test```?

ui/tract-renderer/src/main/java/org/cru/godtools/tract/ui/controller/FallbackController.kt
```diff
@@ -0,0 +1,17 @@
...
+class FallbackController(
+    binding: TractContentFallbackBinding,
+    parentController: BaseController<*>?
+) : ParentController<Fallback>(Fallback::class, binding.root, parentController) {
+    internal constructor(parent: ViewGroup, parentViewHolder: BaseController<*>?) :
+        this(TractContentFallbackBinding.inflate(LayoutInflater.from(parent.context), parent, false), parentViewHolder)
+
+    override val contentContainer = binding.root
+    override val contentToRender get() = model?.content?.take(1)
+}
```
TheNoodleMoose, line 17: So is this a fall back controller for all things? Like content, images, buttons, etc? That's pretty cool!

library/xml-model/src/main/java/org/cru/godtools/xml/model/Content.kt
```diff
@@ -37,11 +46,16 @@ abstract class Content : BaseModel {
 
             return when (parser.namespace) {
                 XMLNS_CONTENT -> when (parser.name) {
-                    Paragraph.XML_PARAGRAPH -> Paragraph(parent, parser)
+                    Paragraph.XML_PARAGRAPH -> when (parser.getAttributeValue(null, Paragraph.XML_FALLBACK)
+                        ?.toBoolean()) {
+                        true -> Fallback(parent, parser)
+                        else -> Paragraph(parent, parser)
```
TheNoodleMoose, line 52: So can you explain the difference where we would use Fallback vs. Paragraphe? Is it because either can be a paragraph, but one could have additional context that isn't just text?

library/xml-model/src/main/java/org/cru/godtools/xml/model/Fallback.kt
```diff
@@ -0,0 +1,33 @@
...
+            else -> parser.require(XmlPullParser.START_TAG, XMLNS_CONTENT, XML_FALLBACK)
+        }
+
+        content = parseContent(parser)
+    }
+
+    @RestrictTo(RestrictTo.Scope.TESTS)
+    constructor(parent: Base, content: ((Fallback) -> List<Content>?)? = null) : super(parent) {
+        this.content = content?.invoke(this).orEmpty()
+    }
```
TheNoodleMoose, line 32: Is this deciding which parser to use based on the name?

TheNoodleMoose (approved):
Got some questions, but at a top-level I think I understand what this is doing.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/ui/controller/ParentController.kt
frett: I override it in `FallbackController`, this allows rendering of fallbacks to only render the first element instead of rendering every element it contains.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/ui/controller/ParentController.kt
frett: ViewGroup is the base class for View that contains child views. Up until we added the `FallbackController` all of our renderer used `LinearLayout` to contain child elements. The `Fallback` rendering logic is using a `FrameLayout` instead, which necessitated widening the class for `contentContainer`

library/xml-model/src/test/resources/org/cru/godtools/xml/model/fallback_paragraph.xml
frett: this is a small snippet of content that is similar to the content the API returns. We are using it to test out the XML parser to ensure it correctly handles various parts of the content spec

ui/tract-renderer/src/main/java/org/cru/godtools/tract/ui/controller/FallbackController.kt
frett: the concept of the fallback content element (and this controller) is that it will render the first child element that is supported by the device. This will allow us to introduce animations and video to the content xml, but still provide "fallback" behavior for older versions of the app.

library/xml-model/src/main/java/org/cru/godtools/xml/model/Content.kt
frett: fallback is meant to be purely render the first child element that is supported. right now we are leveraging paragraphs to provide a mostly backwards compatible mechanism to introduce the fallback concept until apps are upgraded and directly support the `fallback` node directly

library/xml-model/src/main/java/org/cru/godtools/xml/model/Fallback.kt
frett: we are using an `XmlPullParser`, so there is a `parser` object that processes an XML stream and exposes encountered nodes via methods on the `parser` object.

what the first constructor is doing is taking in the parser object, requiring that it is currently processing either `<content:paragraph fallback="true">` or `<content:fallback>`, then calls `parseContents()` to actually parse any child content.

The second constructor is meant for tests and allows us to completely bypass the xml parsing logic and build a `Fallback` node with the related objects we want.

PR #1417, opened 2020-11-18
allow the Firebase User id to be set to null when the user logs out

frett:


library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -86,7 +95,8 @@ class FirebaseAnalyticsService @MainThread @Inject internal constructor(
 
     init {
         oktaUserProfileProvider.userInfoFlow()
-            .mapNotNull { it?.ssoGuid }
+            .map { it?.ssoGuid }
+            .distinctUntilChanged()
```
TheNoodleMoose, line 99: Does this just listen for change and once there is it will update the value? For example, logging out, the values is changed, so we set the userId to ```null```.

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -29,13 +31,20 @@ private const val VALUE_APP_TYPE_INSTANT = "instant"
...
 
 @Singleton
-class FirebaseAnalyticsService @MainThread @Inject internal constructor(
+class FirebaseAnalyticsService @VisibleForTesting internal constructor(
     app: Application,
     eventBus: EventBus,
-    oktaUserProfileProvider: OktaUserProfileProvider
+    oktaUserProfileProvider: OktaUserProfileProvider,
+    private val firebase: FirebaseAnalytics,
+    coroutineScope: CoroutineScope = CoroutineScope(Dispatchers.Default)
```
TheNoodleMoose, line 39: Did you have to add the coroutineScope for testing purposes?

TheNoodleMoose (approved):

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
frett: yeah, I had to expose it so that I could provide a testing coroutineScope

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
frett: it's a small optimization.

if you have a stream of values that looks like: `null`, `a`, `a`, `b`, `null`, `null` it would filter out duplicate values that occur and wouldn't change anything. it would become: `null`, `a`, `b`, `null`.

frett: the `distinctUntilChanged()` flow operator will probably go away when we start reporting additional user profile values, but with just a single value it makes sense for now.

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
frett: there is a coroutines-test library that provides a `TestCoroutineScope`, this test scope provides the following:
- coroutines execute immediately and the original call doesn't return until the coroutine is complete
- full control over the execution clock
- methods to ensure that there are no pending coroutines once your test is complete

frett: the first capability is the critical one here, without it we wouldn't be able to immediately test that code in coroutines ran, we would have to introduce an artificial delay into our test and hope that the code runs before the delay expires.

frett: here's an example test utilizing control over the execution clock: https://github.com/CruGlobal/godtools-android/blob/develop/ui/tract-renderer/src/test/java/org/cru/godtools/tract/ui/controller/HeroControllerTest.kt#L60-L89

PR #1428, opened 2020-11-23
GT-912 use a lottie animation for the remote share menu item

frett:


ui/tract-renderer/src/main/java/org/cru/godtools/tract/util/LottieComposition.kt
```diff
@@ -0,0 +1,5 @@
+package org.cru.godtools.tract.util
+
+import com.airbnb.lottie.LottieComposition
+
+fun LottieComposition.scaleTo(width: Float, height: Float) = minOf(width / bounds.width(), height / bounds.height())
```
TheNoodleMoose, line 5: So are you adding a method to ```LottieComposition``` or tapping into an existing method and overriding it?

ui/tract-renderer/src/main/java/org/cru/godtools/tract/util/LottieUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+    LottieCompositionFactory.fromRawRes(context, rawRes).addListener { comp ->
+        icon = LottieDrawable().apply {
+            composition = comp
+            repeatCount = LottieDrawable.INFINITE
+            val size = TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP, 24f, context.resources.displayMetrics)
+            scale = comp.scaleTo(size, size)
+            playAnimation()
+        }
+    }
+}
```
TheNoodleMoose, line 20: Similar thing here, are you creating a new method on ```MenuItem```? Also:

1. Where are the values in the size value coming from(TypedValue.COMPLEX_UNIT_DIP, etc)
2. Composition is equal to the lottie composition itself, which is the same as the animation right? So when you do ```comp.scaleTo```, you're referring to the method you added/created, right?
3. ```playAnimation()``` is just a method from the lottie library right?

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.kt
```diff
@@ -469,9 +470,11 @@ class TractActivity : BaseToolActivity<TractActivityBinding>(R.layout.tract_acti
     private var liveShareMenuObserver: Observer<Pair<State, State>>? = null
     private fun Menu.setupLiveShareMenuItemVisibility() {
         liveShareMenuObserver?.let { liveShareState.removeObserver(it) }
+
+        val liveShareItem = findItem(R.id.action_live_share_active)
+        liveShareItem?.loadAnimation(this@TractActivity, R.raw.anim_tract_live_share)
         liveShareMenuObserver = Observer<Pair<State, State>> { (publisherState, subscriberState) ->
-            findItem(R.id.action_live_share_active)?.isVisible =
-                publisherState == State.On || subscriberState == State.On
+            liveShareItem?.isVisible = publisherState == State.On || subscriberState == State.On
```
TheNoodleMoose, line 477: The publisherState /subscriberState is referring to web socket logic for liveShare right? publisher being the host(not sure the correct term?), while the subscriber is the person who joins. And then the animation only shows if they are connected via live share?

TheNoodleMoose (approved):

ui/tract-renderer/src/main/java/org/cru/godtools/tract/util/LottieComposition.kt
frett: I'm defining an extension method. It's Kotlin syntactic sugar that can "add" functionality onto an existing class.

https://kotlinlang.org/docs/reference/extensions.html

In reality it gets compiled as a static method that would look like this:
```
fun scaleTo(comp: LottieComposition, width: Float, height: Float)
```

ui/tract-renderer/src/main/java/org/cru/godtools/tract/util/LottieUtils.kt
frett: 1. action icons are supposed to be `24dp` x `24dp`, the `TypedValue` method will resolve the raw pixel size utilizing the `resources.displayMetrics`.
2. yeah, `LottieComposition` represents a loaded animation (`LottieCompositionFactory.fromRawRes()`), calling the `scaleTo()` extension method will calculate how to scale the animation to fit within the 24 x 24 display size.
3. yeah, the method is `LottieDrawable.playAnimation()`. I'm using `.apply` on line 12, which makes `this` the `LottieDrawable`, and I can call methods on `this` without needing to directly reference the object

ui/tract-renderer/src/main/java/org/cru/godtools/tract/activity/TractActivity.kt
frett: yes, the icon shows if the user is publishing or subscribing to a live share stream

frett:
I need to fix the unit tests before merging this, playing the animation isn't working within Robolectric.

PR #1431, opened 2020-11-24
GT-986 the updateShortcutActor should always accept requests

frett:
On older versions of android there was a non-fatal exception thrown anytime the shortcut manager would have ran because the `updateShortcutActor` would finish processing and shutdown on launch.

### DIFF
--- a/ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
+++ b/ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
@@ -205,8 +205,8 @@ class GodToolsShortcutManager @VisibleForTesting internal constructor(
     @VisibleForTesting
     @OptIn(ObsoleteCoroutinesApi::class)
     internal val updateShortcutsActor = coroutineScope.actor<Unit>(capacity = CONFLATED) {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
-            channel.consumeAsFlow().conflate().collectLatest {
+        channel.consumeAsFlow().conflate().collectLatest {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
                 delay(DELAY_UPDATE_SHORTCUTS)
                 updateShortcuts()
             }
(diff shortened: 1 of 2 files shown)


gyasistory (approved):

ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
```diff
@@ -205,8 +205,8 @@ class GodToolsShortcutManager @VisibleForTesting internal constructor(
     @VisibleForTesting
     @OptIn(ObsoleteCoroutinesApi::class)
     internal val updateShortcutsActor = coroutineScope.actor<Unit>(capacity = CONFLATED) {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
-            channel.consumeAsFlow().conflate().collectLatest {
+        channel.consumeAsFlow().conflate().collectLatest {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
```
TheNoodleMoose, line 209: What about reversing these causes it to always accept requests? And what exactly is this doing?

frett: the if statement targets logic to newer versions of Android only, in this case it's needed because the system `ShortcutManager` service was added late in Android Nouget.

the `channel.consumeAsFlow().conflate().collectLatest` statement is the more important one for processing requests. This utilizes several coroutines concepts/objects.

- The first part is `channel`, a coroutine `actor` provides a `Channel` that things can send requests to, within execution of the actor you can consume requests off of that channel.
- We are consuming the channel as a coroutine `Flow`. This is very similar to a `Channel`, but has some differences as to when and how it is evaluated. The main reason I'm converting from a `Channel` to a `Flow` is to access some additional APIs for consuming requests
- `conflate()` indicates to the `Flow` that it only needs to deliver the most recent request if multiple requests come in at the same time. This prevents unnecessary duplicate processing.
- `collectLatest` will take requests off of the flow and process them, `Latest` indicates that if another request shows up while processing the current one, the current one will be canceled and the new request will start processing.

the reason why switching the order fixes the problem is because the `channel.consumeAsFlow().conflate().collectLatest` statement will loop/suspend until the channel is closed. If we only start that on newer versions of Android, then on older versions we never consume the channel and the actor block exits immediately. Once the actor exits, it closes the channel and it's not possible for code to send requests to the actor anymore. What we needed to do was to always loop over actor requests, but only really process the requests on newer versions of Android

PR #1432, opened 2020-11-24
GT-911 Video content element

frett:
Support embedding videos in tool content. So far the only place that a video is embedded is in the header tip for "Renderer Test"

Dependent on:
- [x] https://github.com/CruGlobal/mobile-content-api/pull/521


gyasistory (changes requested):
I am trying to test this PR but I don't see any video on the "Renderer Testing".  What I see is illustrated below.

Android_Emulator_-_Pixel_4_XL_API_30_5554

frett:
because the test video is inside a tip, you need to open the tool in training mode

gyasistory (approved):

PR #1452, opened 2020-12-16
launch the key signup page instead of Okta

frett:
Since Okta self-service registration is still not enabled we are going with the temporary solution of using the key account creation form to create a user account, and then the user will have to know to use login after creating the account to continue.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -45,6 +45,9 @@ android {
 
             buildConfigField "String", "OKTA_CLIENT_ID", "\"0oatqowo16nlzqJRV0h7\""
             buildConfigField "String", "OKTA_DISCOVERY_URI", "\"https://cru.oktapreview.com\""
+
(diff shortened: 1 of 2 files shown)


gyasistory (commented):
Will this work since the login is _**Okta**_ and the signup is _**Key**_

frett:
It will, but isn't the best user experience. The key internally forwards all requests to Okta.

Completing the key signup flow will not actually end with the user logged into GodTools. The user will actually need to select the login menu item after they create the account in order to actually log in

TheNoodleMoose (approved):

TheNoodleMoose:
So they will signup with the key, be dropped back into the login page for okta, and then dropped into the app logged in?

frett:
@TheNoodleMoose yeah, it's not ideal but its functional. Due to technical constraints within Okta we are unable to use Okta's self-service registration form.

PR #1459, opened 2020-12-17
Analytics cleanup

frett:


library/xml-model/src/main/java/org/cru/godtools/xml/model/AnalyticsEvent.kt
```diff
@@ -47,15 +50,20 @@ class AnalyticsEvent {
...
+    constructor(
+        action: String? = null,
+        delay: Int = 0,
+        systems: Set<AnalyticsSystem>? = null,
+        attributes: Map<String, String>? = null
+    ) {
+        this.action = action
         this.delay = delay
-        systems = emptySet()
+        this.systems = systems.orEmpty()
```
TheNoodleMoose, line 61: This just returns an empty string if there is no other string value right?

library/xml-model/src/main/java/org/cru/godtools/xml/model/AnalyticsEvent.kt
```diff
@@ -47,15 +50,20 @@ class AnalyticsEvent {
...
         trigger = Trigger.DEFAULT
-        attributes = emptyMap()
+        this.attributes = attributes.orEmpty()
     }
 
-    internal constructor(parser: XmlPullParser) {
+    internal constructor(parent: Base, parser: XmlPullParser) {
```
TheNoodleMoose, line 66: Ah, so we need the ```parent``` to know which tool the event is coming from?

TheNoodleMoose (approved):
Looks good! Thanks for cleaning this up 🙂

library/xml-model/src/main/java/org/cru/godtools/xml/model/AnalyticsEvent.kt
frett: yeah, the parent gives us access to the parsed object hierarchy. The Manifest at the root of the hierarchy has the tool and locale

library/xml-model/src/main/java/org/cru/godtools/xml/model/AnalyticsEvent.kt
frett: empty set because `systems` is a `Set<AnalyticsSystem>`

TheNoodleMoose: Aahh gotcha its just returning an empty value for whatever the type is. Got it!

PR #1468, opened 2020-12-30
Adobe-Cleanup

TheNoodleMoose:


ui/tract-renderer/src/main/java/org/cru/godtools/tract/analytics/model/ContentAnalyticsActionEvent.kt
```diff
@@ -10,20 +9,5 @@ class ContentAnalyticsActionEvent(@VisibleForTesting internal val event: Analyti
     AnalyticsActionEvent(action = event.action.orEmpty()) {
```
TheNoodleMoose, line 9: This is currently not working since the event names are still adobe based, but once the tools XML is updated to firebase it will work.

library/analytics/src/main/java/org/cru/godtools/analytics/model/AnalyticsActionEvent.kt
```diff
@@ -14,20 +13,13 @@ open class AnalyticsActionEvent(
...
         this(action, label, locale, setOf(system))
 
     open val adobeAttributes: Map<String, *>? get() = null
 
     open val firebaseEventName get() = action
-    open val firebaseParams
-        get() = Bundle().apply {
-            adobeAttributes?.forEach { putString(it.key.sanitizeAdobeNameForFirebase(), it.value?.toString()) }
-        }
+    open val firebaseParams: Map<String, *>? get() = null
```
frett, line 22: I actually want to keep `firebaseParams` returning a `Bundle` since the firebase track event method takes a bundle directly. It provides more direct access to how additional attributes are passed to firebase without needing to assume all values should be stringified.

library/analytics/src/main/java/org/cru/godtools/analytics/model/ExitLinkActionEvent.kt
```diff
@@ -7,5 +7,5 @@ private const val FIREBASE_PARAM_EXIT_LINK = "cru_mobileexitlink"
 
 class ExitLinkActionEvent(link: Uri) : AnalyticsActionEvent(action = ACTION_EXIT_LINK) {
     override fun isForSystem(system: AnalyticsSystem) = system == AnalyticsSystem.FIREBASE
-    override val adobeAttributes = mapOf<String, Any>(FIREBASE_PARAM_EXIT_LINK to link.toString())
+    override val firebaseParams = mapOf<String, Any>(FIREBASE_PARAM_EXIT_LINK to link.toString())
```
frett: this should be converted to populate a `Bundle` object

library/analytics/src/main/java/org/cru/godtools/analytics/model/AnalyticsActionEvent.kt
TheNoodleMoose: Yeah I was trying to find a way to pass the values in order to map through them in the bundle but couldn't figure out an approach that wasn't either making it a function or using a separate value like we are currently doing(ex. ```adobeAttributes```) .

frett: We can work through it next week when I'm back

library/analytics/src/main/java/org/cru/godtools/analytics/model/ExitLinkActionEvent.kt
TheNoodleMoose: Sounds good 🙂

library/analytics/src/main/java/org/cru/godtools/analytics/model/AnalyticsActionEvent.kt
TheNoodleMoose: Sounds good 🙂

library/analytics/src/main/java/org/cru/godtools/analytics/model/AnalyticsActionEvent.kt
```diff
@@ -14,20 +13,13 @@ open class AnalyticsActionEvent(
     locale: Locale? = null,
     systems: Collection<AnalyticsSystem>? = null
 ) : AnalyticsBaseEvent(locale, systems) {
-    companion object {
-        fun String.sanitizeAdobeNameForFirebase(): String = replace(Regex("[ \\-.]"), "_").toLowerCase(Locale.ROOT)
-    }
```
frett, line 19: we won't be able to remove this until all the tool updates are made, I would leave this alone for now

library/analytics/src/main/java/org/cru/godtools/analytics/model/AnalyticsActionEvent.kt
frett: instead of changing `firebaseParams` the focus should be on removing usages of `adobeAttributes` by converting them to return a `Bundle` for `firebaseParams` instead.

frett:
One of the big things this PR should be doing is removing the `AdobeAnalyticsService` and the adobe library

library/analytics/src/main/java/org/cru/godtools/analytics/model/AnalyticsActionEvent.kt
TheNoodleMoose: Ah so make ```firebaseParams``` be equal to ```adobeAttributes``` but change ```adobeAttributes``` to return a bundle instead?

frett: not exactly, I'll send you a google meet request

frett: `adobeAttributes` should be what is being refactored out

ui/base-tool/src/main/java/org/cru/godtools/base/tool/analytics/model/ShareActionEvent.kt
```diff
@@ -3,11 +3,8 @@ package org.cru.godtools.base.tool.analytics.model
 import org.cru.godtools.analytics.model.AnalyticsActionEvent
 import org.cru.godtools.analytics.model.AnalyticsSystem
 
-private const val ACTION_SHARE = "Share Icon Engaged"
-private const val ADOBE_SHARE_CONTENT = "cru.shareiconengaged"
+private const val ACTION_SHARE = "share_icon_engaged"
 
 object ShareActionEvent : AnalyticsActionEvent(action = ACTION_SHARE) {
-    override fun isForSystem(system: AnalyticsSystem) = system == AnalyticsSystem.ADOBE
-    override val adobeAttributes = mapOf<String, Any>(ADOBE_SHARE_CONTENT to 1)
```
frett, line 11: I'm assuming you are just dropping the `ADOBE_SHARE_CONTENT` param since it isn't needed?

TheNoodleMoose: Yep, all the ```cru.``` params don't need to be carried over anymore.

library/analytics/src/main/java/org/cru/godtools/analytics/model/ExitLinkActionEvent.kt
```diff
@@ -1,12 +1,14 @@
...
+private const val FIREBASE_PARAM_EXIT_LINK = "cru_mobileexitlink"
 
 class ExitLinkActionEvent(link: Uri) : AnalyticsActionEvent(action = ACTION_EXIT_LINK) {
-    override fun isForSystem(system: AnalyticsSystem) = system == AnalyticsSystem.ADOBE
-    override val adobeAttributes = mapOf<String, Any>(ADOBE_EXIT_LINK to link.toString())
-    override val firebaseEventName = "exit_link_engaged"
+    override fun isForSystem(system: AnalyticsSystem) = system == AnalyticsSystem.FIREBASE
+    override val firebaseParams = Bundle().apply {
+        putString(FIREBASE_PARAM_EXIT_LINK, link.toString())
+    }
```
frett, line 13:
```suggestion
    override val firebaseParams get() = Bundle().apply {
        putString(FIREBASE_PARAM_EXIT_LINK, link.toString())
    }
```

something subtle you should do here is define the property with a getter, you will also have to make the `link: Uri` constructor param a `private val`.

Using the `get()` syntax will regenerate a fresh bundle every time it is called. otherwise it will generate the Bundle once when the object is initially created, and always return the same `Bundle`. Because `Bundles` are mutable, any changes made to the returned `Bundle` will persist for subsequent calls to `firebaseParams`.

We currently don't modify the returned Bundle, but I can't say that will always be the case.

frett: it was one trade-off I made when switching from `Map` to `Bundle`

ui/tract-renderer/src/main/java/org/cru/godtools/tract/analytics/model/ContentAnalyticsActionEvent.kt
frett: thankfully `firebaseEventName` on this event will correctly sanitize the names when the analytics event was for Adobe originally.

library/analytics/src/main/java/org/cru/godtools/analytics/firebase/FirebaseAnalyticsService.kt
```diff
@@ -67,7 +68,7 @@ class FirebaseAnalyticsService @VisibleForTesting internal constructor(
     @MainThread
     @Subscribe(threadMode = ThreadMode.MAIN_ORDERED)
     fun onAnalyticsEvent(event: AnalyticsBaseEvent) {
-        if (event.isForSystem(AnalyticsSystem.FIREBASE) || event.isForSystem(AnalyticsSystem.ADOBE)) when (event) {
+        if (event.isForSystem(AnalyticsSystem.FIREBASE)) when (event) {
```
frett, line 71: I think we need to leave this for the time being due to the content analytics event

library/analytics/src/test/java/org/cru/godtools/analytics/model/AnalyticsActionEventTest.kt
```diff
@@ -1,30 +0,0 @@
...
-        val firebaseParams = event.firebaseParams
-        assertTrue(firebaseParams.containsKey("cru_contentlanguagesecondary"))
-        assertEquals("en", firebaseParams["cru_contentlanguagesecondary"])
-        assertFalse(firebaseParams.containsKey("cru.contentlanguagesecondary"))
-    }
-
-    @Test
-    fun testSanitizeAdobeNameForFirebase() {
-        assertEquals("a_b_c_d", "a.B c-d".sanitizeAdobeNameForFirebase())
-    }
```
frett, line 29: you should restore this test due to the `sanitizeAdobeNameForFirebase` still being used within the Content analytics event.

`testFirebaseParamsAdobeMigration` will disappear once you finishing phasing out the `adobeAttributes` property.

ui/tract-renderer/src/test/java/org/cru/godtools/tract/analytics/model/ContentAnalyticsActionEventTest.kt
```diff
@@ -1,50 +0,0 @@
...
-            ContentAnalyticsActionEvent(
-                AnalyticsEvent(systems = setOf(AnalyticsSystem.FIREBASE), attributes = mapOf("cru.Key" to "value"))
-            )
-        with(firebaseEvent.firebaseParams) {
-            assertFalse(containsKey("cru_key"))
-            assertTrue(containsKey("cru.Key"))
-            assertEquals("value", getString("cru.Key"))
-        }
-    }
-}
```
frett, line 50: these tests should be restored while we still have content events that need to be updated. although once you phase out `adobeAttributes` you can remove the assertion that was checking `adobeAttributes`

library/analytics/src/main/java/org/cru/godtools/analytics/model/ExitLinkActionEvent.kt
TheNoodleMoose: Would creating the ```private val``` just be doing something like ```private val link = link```?

frett:
```kotlin
class ExitLinkActionEvent(private val link: Uri) : AnalyticsActionEvent(action = ACTION_EXIT_LINK) {
```

library/analytics/src/main/java/org/cru/godtools/analytics/model/AnalyticsActionEvent.kt
```diff
@@ -26,8 +26,8 @@ open class AnalyticsActionEvent(
     open val firebaseEventName get() = action
     open val firebaseParams
         get() = Bundle().apply {
-            adobeAttributes?.forEach { putString(it.key.sanitizeAdobeNameForFirebase(), it.value?.toString()) }
-        }
+        adobeAttributes?.forEach { putString(it.key.sanitizeAdobeNameForFirebase(), it.value?.toString()) }
+    }
```
frett, line 30: I'm kind of surprised that this didn't trigger a ktlint violation for improper indention

TheNoodleMoose: Yeah, I thought it didn't look right but ktlint didn't yell at me. I went ahead and fixed it though.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/analytics/model/ToggleLanguageAnalyticsActionEvent.kt
```diff
@@ -1,25 +1,22 @@
...
+private const val ACTION_TOGGLE_LANGUAGE = "parallel_language_toggled"
 
 class ToggleLanguageAnalyticsActionEvent(tool: String?, locale: Locale) :
     AnalyticsActionEvent(action = ACTION_TOGGLE_LANGUAGE) {
     override fun isForSystem(system: AnalyticsSystem) =
-        system === AnalyticsSystem.ADOBE || system === AnalyticsSystem.FACEBOOK
+        system === AnalyticsSystem.FIREBASE || system === AnalyticsSystem.FACEBOOK
 
     override val appSection = tool
     @OptIn(ExperimentalStdlibApi::class)
```
frett, line 18: you can actually remove this `@OptIn` annotation, it was present for the use of `buildMap` which is no longer being used.

ui/tract-renderer/src/main/java/org/cru/godtools/tract/analytics/model/ToggleLanguageAnalyticsActionEvent.kt
```diff
@@ -1,25 +1,22 @@
...
-    override val adobeAttributes = buildMap<String, Any> {
-        put(ADOBE_ATTR_TOGGLE_LANGUAGE, 1)
-        put(ADOBE_ATTR_LANGUAGE_SECONDARY, toLanguageTag(locale))
+    override val firebaseParams = Bundle().apply {
```
frett, line 19:
```suggestion
    override val firebaseParams get() = Bundle().apply {
```

TheNoodleMoose: After adding this I also had to wrap ```toLanguageTag(locale)``` with this instead ``` locale?.let { toLanguageTag(it) }``` because there was a type mismatch. Is there a better way to do this or is this alright?

frett: hmm, that's changing the locale that gets sent. we will need to preserve the `locale` constructor param, but don't want it to override `locale` from the base analytics event

frett: what it'll probably look like:

```kotlin
class ToggleLanguageAnalyticsActionEvent(tool: String?, private val selectedLocale: Locale) :
    AnalyticsActionEvent(action = ACTION_TOGGLE_LANGUAGE) {

    override val firebaseParams get() = Bundle().apply {
        putString(PARAM_LANGUAGE_SECONDARY, LocaleCompat.toLanguageTag(selectedLocale))
    }
```

TheNoodleMoose: Ah, gotcha! Yeah, I tried just adding ```private val locale: Locale``` initially,  but it yelled at me. It makes sense why though.

frett (approved):

PR #1474, opened 2021-01-12
don't crash when a page is not a tract page

frett:
This is in preperation for lesson page parsing. This will prevent lesson pages from crashing the parser before we actually support lesson pages


frett:
I'm going to skip making this change for now, it doesn't appear to actually be necessary

PR #1476, opened 2021-01-13
GT-963-Implement-Correct-Back-Button-Behavior-For-Tool-Training-Tutorial

TheNoodleMoose:


ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
```diff
@@ -29,14 +29,16 @@ import org.cru.godtools.tutorial.databinding.TutorialActivityBinding
 private const val ARG_PAGE_SET = "pageSet"
 private const val ARG_FRMT_ARGS = "formatArgs"
 
+const val START_TUTORIAL = 102
```
TheNoodleMoose, line 32: Not sure if this value matters or if it just needs to be unique, but at one point it was ```100``` which matched another request code so Okta login was being launched instead.

frett: it needs to be unique within the context of the activity or fragment. if request codes are re-used you will start getting weird behavior.

Also, the request values should be defined where you are starting activities for result, and result codes should be defined where you are returning a result.

In this case it means that this constant should be moved to `ToolDetailsFragment` because that is where you are starting the tutorial from

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsActivity.kt
```diff
@@ -21,6 +28,23 @@ fun Activity.startToolDetailsActivity(toolCode: String) {
...
+) {
+    val code = tool?.code ?: return
+    val locale = translation?.languageCode
+
+    if (force || isDiscovered) openToolActivity(code, type, locale, showTips = true)
+    else startTutorialActivity(PageSet.TIPS, Bundle().apply {
+        putString(TutorialActivity.FORMAT_ARG_TOOL_NAME, translation.getName(tool).toString())
+    }
+    )
+}
```
frett, line 46: this should be a regular method within `ToolDetailsFragment`

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -129,17 +144,14 @@ class ToolDetailsFragment() :
 
     fun openToolTraining(tool: Tool?, translation: Translation?) {
         val code = tool?.code ?: return
-        val locale = translation?.languageCode ?: return
+        translation?.languageCode ?: return
```
frett, line 147: you should still assign the locale to a local property and then just use the local property

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragmentDataModel.kt
```diff
@@ -23,10 +26,15 @@ import org.cru.godtools.shortcuts.GodToolsShortcutManager
...
 
+private const val KEY_TIPS_LANGUAGE = "KEY_TIPS_LANGUAGE"
+private const val KEY_TIPS_TOOL = "KEY_TIPS_TOOL"
+private const val KEY_TIPS_TYPE = "KEY_TIPS_TYPE"
+
 class ToolDetailsFragmentDataModel @ViewModelInject constructor(
     private val dao: GodToolsDao,
     private val downloadManager: GodToolsDownloadManager,
     manifestManager: ManifestManager,
+    @Assisted private val savedStateHandle: SavedStateHandle,
```
frett, line 37: just a minor nit, I would put this as the last constructor parameter because it is not a service singleton like everything else

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
```diff
@@ -29,14 +29,16 @@ import org.cru.godtools.tutorial.databinding.TutorialActivityBinding
...
+
 fun Context.buildTutorialActivityIntent(pageSet: PageSet, formatArgs: Bundle? = null) =
     Intent(this, TutorialActivity::class.java)
         .putExtra(ARG_PAGE_SET, pageSet)
         .putExtra(ARG_FRMT_ARGS, formatArgs)
 
 @JvmOverloads
 fun Activity.startTutorialActivity(pageSet: PageSet, stringArgs: Bundle? = null) =
-    startActivity(buildTutorialActivityIntent(pageSet, stringArgs))
+    startActivityForResult(buildTutorialActivityIntent(pageSet, stringArgs), START_TUTORIAL)
```
frett, line 41: this change shouldn't be made here, this should be handled within `ToolDetailsFragment`.

it will be something like:
```kotlin
startActivityForResult(context.buildTutorialActivityIntent(), REQUEST_TUTORIAL_TIPS)
```

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
```diff
@@ -183,7 +185,10 @@ class TutorialActivity : BaseActivity<TutorialActivityBinding>(), TutorialCallba
                 setResult(RESULT_OK)
                 finish()
             }
-            R.id.action_training_finish, R.id.action_tips_finish -> finish()
+            R.id.action_training_finish, R.id.action_tips_finish -> {
+                setResult(RESULT_OK)
+                finish()
+            }
```
frett, line 191: instead of doing this, just move the `R.id.action_tips_finish` id to the previous section where we are already setting OK and finishing.

It doesn't hurt anything moving `R.id.action_training_finish` there as well, but there is no need to move it. So it doesn't matter to me if you consolidate that in as well or not.

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsActivity.kt
```diff
@@ -21,6 +28,23 @@ fun Activity.startToolDetailsActivity(toolCode: String) {
...
+    translation: Translation,
+    isDiscovered: Boolean = false,
+    force: Boolean
+) {
+    val code = tool?.code ?: return
+    val locale = translation?.languageCode
+
+    if (force || isDiscovered) openToolActivity(code, type, locale, showTips = true)
+    else startTutorialActivity(PageSet.TIPS, Bundle().apply {
+        putString(TutorialActivity.FORMAT_ARG_TOOL_NAME, translation.getName(tool).toString())
```
frett, line 43: I also don't think we need to send this string to the tutorial anymore. I'm pretty sure we removed the tool name from the tutorial

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/PageSet.kt
```diff
@@ -65,6 +65,7 @@ enum class PageSet(
         )
     ),
     TIPS(
+        menu = R.menu.tutorial_tips_menu,
```
TheNoodleMoose, line 68: I saw that the ticket mentioned also adding an exit option when going through the tutorial. Is the skip fine or should it be an X instead?

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
TheNoodleMoose: I was confused by this at first but I see you're just saying to consolidate it to the section above since it's doing the same thing. Will do!

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -28,12 +29,13 @@ import org.cru.godtools.model.Translation
...
 import org.cru.godtools.shortcuts.PendingShortcut
 import org.cru.godtools.tutorial.PageSet
-import org.cru.godtools.tutorial.activity.TutorialActivity.Companion.FORMAT_ARG_TOOL_NAME
-import org.cru.godtools.tutorial.activity.startTutorialActivity
+import org.cru.godtools.tutorial.activity.buildTutorialActivityIntent
 import org.cru.godtools.ui.tools.analytics.model.AboutToolButtonAnalyticsActionEvent
 import org.cru.godtools.util.openToolActivity
 import splitties.fragmentargs.arg
 
+const val REQUEST_TUTORIAL_TIPS = 102
```
frett, line 37: this can be `private`

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -76,6 +78,22 @@ class ToolDetailsFragment() :
...
 
+    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+        when (requestCode) {
+            REQUEST_TUTORIAL_TIPS -> when (resultCode) {
+                Activity.RESULT_OK -> {
+                    settings.setFeatureDiscovered("$FEATURE_TUTORIAL_TIPS${dataModel.tipsTool}")
+                    dataModel.tipsTool?.let {
+                        tool -> dataModel.tipsType?.let {
+                        type -> dataModel.tipsLanguage?.let {
+                        language -> requireActivity().openToolActivity(tool, type, language, showTips = true) } } }
```
frett, line 89: I'm not a fan of nesting `let` blocks like this. a better pattern is:
```kotlin
val code = dataModel.tipsTool ?: return
val type = dataModel.tipsType ?: return
val language = dataModel.tipsLanguage ?: return
```

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -132,16 +150,24 @@ class ToolDetailsFragment() :
...
+            dataModel.tipsType = tool.type
+            launchTrainingTips(code, tool.type, locale, isToolDiscovered, false)
         }
     }
+
+    private fun launchTrainingTips(
+        code: String,
+        type: Tool.Type,
+        locale: Locale,
+        isDiscovered: Boolean = false,
```
frett, line 165: don't pass this into this method, calculate it directly within the if statement

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
frett: instead of calling `openToolActivity` here, call `launchTrainingTips` with `force = true`

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -132,16 +150,24 @@ class ToolDetailsFragment() :
         val locale = translation?.languageCode ?: return
 
         activity?.run {
```
frett, line 152: you no longer need `activity?.run` for this logic

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -132,16 +150,24 @@ class ToolDetailsFragment() :
...
-            openToolActivity(code, tool.type, locale, showTips = true)
-            if (settings.getFeatureDiscoveredCount("$FEATURE_TUTORIAL_TIPS$code") < 3) {
-                settings.setFeatureDiscovered("$FEATURE_TUTORIAL_TIPS$code")
-                startTutorialActivity(
-                    PageSet.TIPS,
-                    Bundle().apply { putString(FORMAT_ARG_TOOL_NAME, translation.getName(tool).toString()) }
-                )
-            }
+            val isToolDiscovered = settings.isFeatureDiscovered("$FEATURE_TUTORIAL_TIPS$code")
+            dataModel.tipsLanguage = translation.languageCode
```
frett, line 154:
```suggestion
            dataModel.tipsLanguage = locale
```

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
TheNoodleMoose: Oh yeah, Elvis operator! I like that better and will change.

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -76,6 +78,22 @@ class ToolDetailsFragment() :
         menu.setupPinShortcutAction()
     }
 
+    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+        when (requestCode) {
+            REQUEST_TUTORIAL_TIPS -> when (resultCode) {
+                Activity.RESULT_OK -> {
+                    settings.setFeatureDiscovered("$FEATURE_TUTORIAL_TIPS${dataModel.tipsTool}")
```
frett, line 85: let's move this into `launchTrainingTips`, set the feature discovered when it is taking the branch that is launching the tract and not the tutorial

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -131,16 +149,21 @@ class ToolDetailsFragment() :
...
-                settings.setFeatureDiscovered("$FEATURE_TUTORIAL_TIPS$code")
-                startTutorialActivity(
-                    PageSet.TIPS,
-                    Bundle().apply { putString(FORMAT_ARG_TOOL_NAME, translation.getName(tool).toString()) }
-                )
-            }
-        }
+        dataModel.tipsLanguage = locale
+        dataModel.tipsTool = code
+        dataModel.tipsType = tool.type
```
frett, line 154: move these into `launchTrainingTips`, set these values when you are launching the tutorial

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -131,16 +149,21 @@ class ToolDetailsFragment() :
...
+
+    private fun launchTrainingTips(
+        code: String,
+        type: Tool.Type,
+        locale: Locale,
+        force: Boolean
+    ) {
+        if (force || settings.isFeatureDiscovered("$FEATURE_TUTORIAL_TIPS$code"))
+            requireActivity().openToolActivity(code, type, locale, showTips = true)
+        else startActivityForResult(context?.buildTutorialActivityIntent(PageSet.TIPS), REQUEST_TUTORIAL_TIPS)
```
frett, line 166: change `context?` to `requireContext()`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/PageSet.kt
frett: I think having a skip action is fine

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/activity/TutorialActivity.kt
```diff
@@ -88,6 +85,11 @@ class TutorialActivity : BaseActivity<TutorialActivityBinding>(), TutorialCallba
             finish()
             true
         }
+        R.id.action_tips_skip -> {
```
frett, line 88: can be combined with `R.id.action_live_share_skip`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/PageSet.kt
frett: wow, I evidently added the skip menu item, but forgot to include the menu in the tutorial page set. good catch that it already existed and just wasn't wired up

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
TheNoodleMoose: So something like this?
```
if (force || settings.isFeatureDiscovered("$FEATURE_TUTORIAL_TIPS$code")) {
     settings.setFeatureDiscovered("$FEATURE_TUTORIAL_TIPS${dataModel.tipsTool}")
     requireActivity().openToolActivity(code, type, locale, showTips = true)
}
```
Or should I have a specific case for just ```force``` being true and a case for when ```settings.isFeatureDiscovered("$FEATURE_TUTORIAL_TIPS$code")``` is true?

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
TheNoodleMoose: I assume ```requireContext``` just makes sure the context is available before running the function?

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
frett: your example is what I'd expect, no special case needed for `force`.

frett: also, if you want you can rename `force` to `skipTutorial`, it would make the concept of the flag more straight-forward

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
frett: yeah, `requireContext()` returns a non-null context or throws an error. without that change, when `context` is null, you will have a null intent, which would throw a different error when trying to use `startActivityForResult`.

frett: similar concept as `requireActivity()` vs `activity`

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
TheNoodleMoose: Yeah I like ```skipTutorial``` better, will change

app/src/main/java/org/cru/godtools/ui/tooldetails/ToolDetailsFragment.kt
```diff
@@ -131,15 +150,23 @@ class ToolDetailsFragment() :
...
+    }
+
+    private fun launchTrainingTips(
+        code: String,
+        type: Tool.Type,
+        locale: Locale,
+        skipTutorial: Boolean
+    ) {
+        if (skipTutorial || settings.isFeatureDiscovered("$FEATURE_TUTORIAL_TIPS$code")) {
+            settings.setFeatureDiscovered("$FEATURE_TUTORIAL_TIPS${dataModel.tipsTool}")
```
frett, line 163: use `$code` instead of `${dataModel.tipsTool}`

TheNoodleMoose: Okay, done!

frett (approved):

PR #1499, opened 2021-01-25
GT-1036 Lesson tool type parsing

frett:


frett:
@TheNoodleMoose I assigned this one to you since it's touching the XML parsing logic which you will be interacting with for a few of the other jira tasks you have

library/xml-model/src/test/java/org/cru/godtools/xml/model/ManifestTest.kt
```diff
@@ -3,32 +3,63 @@ package org.cru.godtools.xml.model
...
-@RunWith(AndroidJUnit4::class)
 class ManifestTest {
     @Test
-    fun verifyParseEmptyManifest() {
+    fun testManifestTypeParsing() {
+        assertNull(Manifest.Type.parseOrNull(null))
+        assertEquals(Manifest.Type.ARTICLE, Manifest.Type.parseOrNull("article"))
+        assertEquals(Manifest.Type.LESSON, Manifest.Type.parseOrNull("lesson"))
+        assertEquals(Manifest.Type.TRACT, Manifest.Type.parseOrNull("tract"))
+        assertEquals(Manifest.Type.UNKNOWN, Manifest.Type.parseOrNull("nasldkja"))
```
TheNoodleMoose, line 21: How do we handle parsing the unknown case?

TheNoodleMoose (approved):
Looks pretty good 🙂 Going through and asking questions on that last PR made looking at this one easier!

library/xml-model/src/test/java/org/cru/godtools/xml/model/ManifestTest.kt
frett: the `UNKNOWN` type is meant to reflect future tool types. basically when `Manifest.Type.parseOrNull()` is called with a type it doesn't recognize, it's meant to return `UNKNOWN`. This allows us to write logic that can gracefully handle future manifests we don't know about yet.

frett: I typed some random characters that will probably never be a valid tool type to test out the `UNKNOWN` state

PR #1510, opened 2021-01-28
Revert "Bump deps.androidX.hilt from 1.0.0-alpha02 to 1.0.0-alpha03"

frett:
Reverts CruGlobal/godtools-android#1509


frett:
this issue needs to be addressed before we upgrade: https://github.com/google/dagger/issues/2328

PR #1514, opened 2021-01-29
Update-cru_appname-Value

TheNoodleMoose:
The analytics team requested we update this value to be the same across all platforms/apps. So it should always be "app name + App".


frett (approved):

frett:
@TheNoodleMoose do you want me to include this in the bugfix release so it gets out soon?

TheNoodleMoose:
@Daniel Yeah it can go in!

PR #1517, opened 2021-01-29
GT-1044 Fix ShortcutManager crash on Instant App

frett:
Instant Apps don't have access to the ShortcutManager system service. This led to a crash where updating dynamic shortcuts would utilize `ShortcutManagerCompat` which assumes the `ShortcutManager` is available on newer versions of Android. This led to an NPE.

we now only attempt to update dynamic shortcuts if we have an actual `ShortcutManager`

### DIFF
--- a/ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
+++ b/ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
@@ -233,22 +233,25 @@ class GodToolsShortcutManager @VisibleForTesting internal constructor(
 
     @VisibleForTesting
     @RequiresApi(Build.VERSION_CODES.N_MR1)
-    internal suspend fun updateDynamicShortcuts(shortcuts: Map<String, ShortcutInfoCompat>) =
-        withContext(ioDispatcher) {
-            val dynamicShortcuts = Query.select<Tool>()
+    internal suspend fun updateDynamicShortcuts(shortcuts: Map<String, ShortcutInfoCompat>) {
+        val manager = shortcutManager ?: return
+
+        val dynamicShortcuts = withContext(ioDispatcher) {
+            Query.select<Tool>()
                 .where(ToolTable.FIELD_ADDED.eq(true))
                 .orderBy(ToolTable.SQL_ORDER_BY_ORDER)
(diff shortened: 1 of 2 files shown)


ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
```diff
@@ -233,22 +233,25 @@ class GodToolsShortcutManager @VisibleForTesting internal constructor(
...
+        val manager = shortcutManager ?: return
+
+        val dynamicShortcuts = withContext(ioDispatcher) {
+            Query.select<Tool>()
                 .where(ToolTable.FIELD_ADDED.eq(true))
                 .orderBy(ToolTable.SQL_ORDER_BY_ORDER)
                 .get(dao).asSequence()
                 .mapNotNull { shortcuts[it.shortcutId]?.toShortcutInfo() }
-                .take(ShortcutManagerCompat.getMaxShortcutCountPerActivity(context))
+                .take(manager.maxShortcutCountPerActivity)
```
TheNoodleMoose, line 245: Can you explain the main reason we don't have to pass context anymore to this? I think I have an idea but just want to make sure it's the right one haha.

TheNoodleMoose (approved):

ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
frett: `ShortcutManagerCompat` provides backwards compatible wrappers around new APIs. In this case it was providing a wrapper around `ShortcutManager.getMaxShortcutCountPerActivity()` which is only available starting with API 25.

`ShortcutManagerCompat` requires the `context` object so it can actually look up the `ShortcutManager` when it is run on a supported version of Android, the new logic already has the `ShortcutManager` in `manager`, so it doesn't need the `context` object to look it up

frett: `ShortcutManagerCompat.getMaxShortcutCountPerActivity()` was a static method, `manager.maxShortcutCountPerActivity` is an instance method on the `ShortcutManager` class

PR #1523, opened 2021-02-01
GT-1048 GT-1060 Card background images

frett:
utilize a matrix scale type to position/scale background images for tools, pages, and cards.

I'm not totally happy with how I had to write tests, but I'm not sure of a better way to programmatically test how an ImageView renders a drawable without needing to run an emulator.

Instead I'm testing to make sure that the Matrix transform is generated in an expected manner based upon the drawable and image size


frett:
I'm not totally happy with how I had to write tests, but I'm not sure of a better way to programmatically test how an ImageView renders a drawable without needing to run an emulator.

Instead I'm testing to make sure that the Matrix transform is generated in an expected manner based upon the drawable and image size

gyasistory (approved):

PR #1530, opened 2021-02-04
GT-592

TheNoodleMoose:


library/xml-model/src/main/java/org/cru/godtools/xml/model/tract/TractPage.kt
```diff
@@ -73,6 +73,11 @@ class TractPage : BaseModel, Styles {
     @get:ColorInt
     val cardTextColor get() = _cardTextColor ?: textColor
 
+    @ColorInt
+    private val _cardBackgroundColor: Int?
+    @get:ColorInt
+    val cardBackgroundColor get() = _cardBackgroundColor ?: manifest.cardBackgroundColor
```
frett, line 79: I'm actually glad you implemented this here, rethinking the one long elvis operator chain, I was going to have you refactor to utilizing this.

library/xml-model/src/main/java/org/cru/godtools/xml/model/tract/Card.kt
```diff
@@ -53,11 +53,10 @@ class Card : BaseModel, Styles, Parent {
     @get:ColorInt
     override val textColor get() = _textColor ?: page.cardTextColor
 
-    // TODO: implement card-background-color on Page & Manifest
     @ColorInt
     private val _backgroundColor: Int?
     @get:ColorInt
-    internal val backgroundColor get() = _backgroundColor ?: manifest.backgroundColor
+    internal val backgroundColor get() = _backgroundColor ?: page.cardBackgroundColor ?: manifest.cardBackgroundColor ?: manifest.backgroundColor
```
frett, line 59: because you have the page `card-background-color` fallback to the manifest `card-background-color` defined in `TractPage`, you can leave `manifest.cardBackgroundColor` out of this elvis operator chain

library/xml-model/src/main/java/org/cru/godtools/xml/model/Manifest.kt
```diff
@@ -95,7 +95,9 @@ class Manifest : BaseModel, Styles {
     val backgroundImageScaleType: ImageScaleType
 
     @ColorInt
-    val cardBackgroundColor: Int?
+    private val _cardBackgroundColor: Int?
+    @get:ColorInt
+    val cardBackgroundColor: Int get() = _cardBackgroundColor ?: manifest.backgroundColor
```
frett, line 100: no need for `manifest.` here, you are already in the manifest :)

library/xml-model/src/test/java/org/cru/godtools/xml/model/tract/CardTest.kt
```diff
@@ -68,4 +72,21 @@ class CardTest {
...
+        )
+        val card = Card(page, content = {
+            listOf(
+                InlineTip(it, "tip1"),
+                Paragraph(it, content = { listOf(InlineTip(it, "tip2")) }),
+                InlineTip(it, "tip3"),
+                InlineTip(it, "tip1"),
+            )
+        })
+        assertThat(card.backgroundColor, equalTo(page.cardBackgroundColor))
```
frett, line 90: I would explicitly set `page.cardBackgroundColor` to something like `Color.GREEN`, that way you know the card color is definitely using the page color.

library/xml-model/src/test/java/org/cru/godtools/xml/model/tract/CardTest.kt
```diff
@@ -68,4 +72,21 @@ class CardTest {
...
+            Manifest(),
+            cards = { listOf(Card(it, isHidden = false), Card(it, isHidden = false), Card(it, isHidden = true)) }
+        )
+        val card = Card(page, content = {
+            listOf(
+                InlineTip(it, "tip1"),
+                Paragraph(it, content = { listOf(InlineTip(it, "tip2")) }),
+                InlineTip(it, "tip3"),
+                InlineTip(it, "tip1"),
+            )
```
frett, line 88: there should be no need to define card content since this test isn't doing anything with the content

library/xml-model/src/test/java/org/cru/godtools/xml/model/tract/CardTest.kt
```diff
@@ -68,4 +72,21 @@ class CardTest {
         assertTrue(page.cards[1].isLastVisibleCard)
         assertFalse(page.cards[2].isLastVisibleCard)
     }
+
+    @Test
+    fun testCardBackgroundColorFallbackBehaviorForTractPage() {
+        val page = TractPage(
+            Manifest(),
+            cards = { listOf(Card(it, isHidden = false), Card(it, isHidden = false), Card(it, isHidden = true)) }
```
frett, line 80: you aren't using any of these cards, so no need to create them here

library/xml-model/src/test/java/org/cru/godtools/xml/model/tract/TractPageTest.kt
```diff
@@ -43,6 +43,8 @@ class TractPageTest {
         assertThat(page.cards, hasSize(2))
         assertEquals("Card 1", page.cards[0].label!!.text)
         assertEquals("Card 2", page.cards[1].label!!.text)
+        assertEquals(page.cards[0].backgroundColor, page.cardBackgroundColor)
+        assertEquals(page.cards[1].backgroundColor, page.cardBackgroundColor)
```
frett, line 47: this is basically testing the card parsing (already tested in `CardTest`), as well as the page card background color fallback (also already tested in `CardTest`).

I'd rather see that the parsing logic correctly reads page.cardBackgroundColor from the XML

library/xml-model/src/main/java/org/cru/godtools/xml/model/Manifest.kt
TheNoodleMoose: So do I need just have ```val cardBackgroundColor: Int?``` or should I have it be ```DEFAULT_BACKGROUND_COLOR``` instead of ```manifest.backgroundColor```?

library/xml-model/src/main/java/org/cru/godtools/xml/model/Manifest.kt
```diff
@@ -93,6 +94,11 @@ class Manifest : BaseModel, Styles {
     internal val backgroundImageGravity: ImageGravity
     val backgroundImageScaleType: ImageScaleType
 
+    @ColorInt
+    private val _cardBackgroundColor: Int?
+    @get:ColorInt
+    val cardBackgroundColor: Int get() = _cardBackgroundColor ?: manifest.backgroundColor
```
frett, line 100: This should be all you need here

```suggestion
    @ColorInt
    private val _cardBackgroundColor: Int?
    @get:ColorInt
    val cardBackgroundColor: Int get() = _cardBackgroundColor ?: backgroundColor
```

frett: all `manifest` does on the manifest itself is return `this`

library/xml-model/src/main/java/org/cru/godtools/xml/model/Manifest.kt
frett: you do want it to fallback to the manifest background color, but you are already in the manifest object, so there is no need to use `manifest.` to access it

TheNoodleMoose: Ah okay gotcha!

library/xml-model/src/test/java/org/cru/godtools/xml/model/tract/TractPageTest.kt
TheNoodleMoose: So should I put that inside of a separate test or could I add the assertions inside of something like `verifyParseCards()`  where we don't define a ```cardBackgroundColor``` so it should fall back to ```manifest.cardBackgroundColor```?

frett: I would define a separate test method to test the page cardBackgroundColor fallback behavior

TheNoodleMoose:
@frett So I think I have all the tests needed. I have these currently:

1. Test parsing background-color from card.xml
2. Test fall back behavior: card.backgroundColor -> page.cardBackgroundColor
3. Test fall back behavior: page.cardBackgroundColor -> manifest.cardBackgroundColor
4. Test fall back behavior: manifest.cardBackgroundColor -> manifest.backgroundColor

Let me know if there are any additional ones I need or ways to improve what I have 🙂

library/xml-model/src/test/java/org/cru/godtools/xml/model/ManifestTest.kt
```diff
@@ -75,6 +75,12 @@ class ManifestParsingRobolectricTest {
         assertEquals(0, manifest.tips.size)
     }
 
+    @Test
+    fun testCardBackgroundColorFallbackBehavior() {
+        val manifest = parseManifest("manifest_tract.xml")
```
frett, line 80: for testing the fallback behavior I want to avoid parsing xml, so create your objects like this instead:

```kotlin
val manifest = Manifest(backgroundColor = Color.GREEN)
```

frett: This logic should cover testing both fallback scenarios:

```kotlin
assertEquals(Color.GREEN, Manifest(backgroundColor = Color.RED, cardBackgroundColor = Color.GREEN))
assertEquals(Color.GREEN, Manifest(backgroundColor = Color.GREEN))
```

frett: you will probably need to add `backgroundColor` and `cardBackgroundColor` to the Manifest test constructor

library/xml-model/src/test/java/org/cru/godtools/xml/model/tract/CardTest.kt
```diff
@@ -25,9 +30,10 @@ class CardTest {
         val listenerEvents = Event.Id.parse(TOOL_CODE, "listener1 listener2")
         val dismissListenerEvents = Event.Id.parse(TOOL_CODE, "dismiss-listener1 dismiss-listener2")
 
-        val card = parseCardXml("card.xml")
+        val card = parseCardXml("card.xml", )
```
frett, line 33: you shouldn't have the trailing comma here, guessing that's one of the lint failures

library/xml-model/src/test/java/org/cru/godtools/xml/model/tract/TractPageTest.kt
```diff
@@ -53,5 +53,11 @@ class TractPageTest {
         assertEquals("Modal 2", page.modals[1].title!!.text)
     }
 
+    @Test
+    fun testCardBackgroundColorFallbackBehavior() {
+        val page = parsePageXml("page_cards.xml")
+        assertEquals(page.cardBackgroundColor, manifest.cardBackgroundColor)
```
frett, line 59: similar to the manifest test, I want to avoid making the page cardBackgroundColor fallback behavior test dependent on parsing xml

library/xml-model/src/test/java/org/cru/godtools/xml/model/ManifestTest.kt
TheNoodleMoose: Oh yeah, will do!

TheNoodleMoose: So ```backgroundColor``` doesn't have a private val ```_backgroundColor``` like ```cardBackgroundColor``` so trying to add ```backgroundColor``` to the test constructor is throwing errors. So I changed it to this
```
@ColorInt
private val _backgroundColor: Int?
@get:ColorInt
val backgroundColor: Int get() = _backgroundColor ?: DEFAULT_BACKGROUND_COLOR
```

Is this correct or should I just make the test constructor argument be ```_backgroundColor```?

frett: neither, you will just need to prefix where you assign it with `this.backgroundColor`

frett: take a look at how `code` and `locale` are set in the test constructor

TheNoodleMoose: Ah cool!

library/xml-model/src/test/java/org/cru/godtools/xml/model/tract/CardTest.kt
```diff
@@ -68,4 +74,11 @@ class CardTest {
         assertTrue(page.cards[1].isLastVisibleCard)
         assertFalse(page.cards[2].isLastVisibleCard)
     }
+
+    @Test
+    fun testCardBackgroundColorFallbackBehavior() {
+        val page = TractPage(Manifest(), cardBackgroundColor = Color.GREEN)
+        val card = Card(page, backgroundColor = null)
+        assertThat(card.backgroundColor, equalTo(page.cardBackgroundColor))
```
frett, line 82: This will test both fallback scenarios
```suggestion
        assertEquals(Color.GREEN, Card(page).backgroundColor)
        assertEquals(Color.BLUE, Card(page, backgroundColor = Color.BLUE).backgroundColor)
```

TheNoodleMoose:
Okay, I think this is good 🙂 If you could double-check my commit where I resolved merge conflicts I'd appreciate it!

frett (approved):
lgtm

PR #1536, opened 2021-02-08
GT-1072-Add-Hindi-To-Onboarding

TheNoodleMoose:
@frett This seemed pretty straightforward and worked when I tested locally. The ticket says the files are on OneSky but they appeared to be in the ```res``` folder under ```values-hi```. Not sure if there is anything else I need to add so just let me know!

### DIFF
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/PageSet.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/PageSet.kt
@@ -29,6 +29,7 @@ enum class PageSet(
             Locale.ENGLISH,
             Locale("es"),
             Locale.FRENCH,
+            Locale("hi"),
             Locale("in"),
             Locale("ru"),
             Locale.SIMPLIFIED_CHINESE,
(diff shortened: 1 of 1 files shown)


frett (approved):

frett:
I periodically update the strings from OneSky, so if the translation had been done for a while I would have already imported them. I'm sure I'll run another import before we publish a new version anyways.

right now importing the strings is a manual process

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/PageSet.kt
```diff
@@ -29,6 +29,7 @@ enum class PageSet(
             Locale.ENGLISH,
             Locale("es"),
             Locale.FRENCH,
+            Locale("hi"),
```
frett, line 32: I actually think you need to add it to the `TRAINING` PageSet as well. that controls whether the user can open up the training tutorial from the nav drawer

frett (changes requested):
I think the language needs to be added to the TRAINING PageSet as well

frett (approved):

PR #1540, opened 2021-02-09
GT-1031-Add-Support-For-Start/End-Images-On-Text-Content-Elements

TheNoodleMoose:


library/xml-model/src/main/java/org/cru/godtools/xml/model/Text.kt
```diff
@@ -8,6 +8,7 @@ import org.ccci.gto.android.common.util.xmlpull.skipTag
 import org.cru.godtools.xml.XMLNS_CONTENT
 import org.xmlpull.v1.XmlPullParser
 
+private const val XML_RESOURCE = "resource"
```
frett, line 11: so, this is the incorrect attribute name for parsing `start-image` and `end-image` (or whatever I actually named the attributes)

library/xml-model/src/main/java/org/cru/godtools/xml/model/Text.kt
```diff
@@ -65,6 +76,9 @@ class Text : Content {
         _textColor = parser.getAttributeValueAsColorOrNull(XML_TEXT_COLOR)
         _textScale = parser.getAttributeValue(null, XML_TEXT_SCALE)?.toDoubleOrNull()
 
+        imageStartName = parser.getAttributeValue(null, XML_RESOURCE)
+        imageEndName = parser.getAttributeValue(null, XML_RESOURCE)
```
frett, line 80: you'll want to use the corresponding attribute names with this methods

library/xml-model/src/main/java/org/cru/godtools/xml/model/Text.kt
```diff
@@ -44,18 +45,28 @@ class Text : Content {
...
+        textAlign: Align? = null,
+        imageStart: String? = null,
+        imageEnd: String? = null
     ) : super(parent) {
         this.text = text
         _textAlign = textAlign
         _textColor = textColor
         _textScale = textScale
+        this.imageStartName = imageStart
+        this.imageEndName = imageEnd
```
frett, line 69: no need for `this.` on these because there isn't ambiguity between the properties and constructor params

library/xml-model/src/main/java/org/cru/godtools/xml/model/Text.kt
TheNoodleMoose: Aahhhh, gotcha! So this const is just for referencing the specific attribute you are trying to parse and will change depending on the attribute name.

frett: yeah, it's just extracting a string constant out

ui/base-tool/src/main/java/org/cru/godtools/base/tool/databinding/TextViewAdapters.kt
```diff
@@ -3,9 +3,44 @@ package org.cru.godtools.base.tool.databinding
...
 
 @BindingAdapter("android:text", "android:textSize", "defaultTextColor", requireAll = false)
 fun TextView.bindTextNode(text: Text?, textSize: Float?, @ColorInt defaultTextColor: Int?) =
     text.bindTo(this, textSize, defaultTextColor)
+
+@BindingAdapter("android:drawableStart", "startImageSize")
+fun TextView.bindDrawableStartResource(resource: Resource?, startImageSize: Int) {
+    val target = TextViewDrawableStartTarget.of(this)
+    val file = resource?.localName?.let { context.fileManager.getFileBlocking(it) }
+    val imageSize = (startImageSize * resources.displayMetrics.density).toInt()
```
frett, line 22: I like using `TypedValue` to do the conversion for us instead of hard-coding the formula.

```
TypedValue.applyDimension(COMPLEX_UNIT_DIP, startImageSize, resources.displayMetrics)
```

ui/base-tool/src/main/java/org/cru/godtools/base/tool/databinding/TextViewAdapters.kt
```diff
@@ -3,9 +3,44 @@ package org.cru.godtools.base.tool.databinding
...
+import org.cru.godtools.base.fileManager
 import org.cru.godtools.base.tool.model.view.bindTo
+import org.cru.godtools.xml.model.Resource
 import org.cru.godtools.xml.model.Text
 
 @BindingAdapter("android:text", "android:textSize", "defaultTextColor", requireAll = false)
 fun TextView.bindTextNode(text: Text?, textSize: Float?, @ColorInt defaultTextColor: Int?) =
     text.bindTo(this, textSize, defaultTextColor)
+
+@BindingAdapter("android:drawableStart", "startImageSize")
```
frett, line 18: I would rather name the binding attribute `drawableStartSize`

frett: that way in the view layout it's easier to see the correlation between `drawableStart` and `drawableStartSize`

ui/base-tool/src/main/java/org/cru/godtools/base/tool/databinding/TextViewAdapters.kt
frett: At some point I'm going to play around with the layout a little bit to see if data binding does any automatic conversions for us in the case of using a `dimen` value directly for the size attribute.

If so I might adjust where we do the dp -> px conversion to better mesh with data binding

TheNoodleMoose: @frett So to fix the typing issues this is what I have
```
val imageSize = TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP, startImageSize.toFloat(), resources.displayMetrics).toInt()
```

Does that look alright to you?  ```applyMethod``` expects a ```Float``` for the second argument so it's kind of weird to have to turn the Int -> Float and then convert the whole thing to an Int.

frett: hmm, let me do a bit of research

frett: stick with what you have for the time being, I might revisit this later and implement some Kotlin utils to make something like this possible:

```kotlin
Dp(startImageSize).toPx(resources.displayMetrics).toInt()
```

ui/base-tool/src/main/res/layout/tool_content_text.xml
```diff
@@ -12,6 +12,8 @@
         android:layout_marginRight="@dimen/tool_content_margin_horizontal"
         android:drawableStart="@{model.startImage}"
         android:drawableEnd="@{model.endImage}"
+        android:drawablePadding="8dp"
```
TheNoodleMoose, line 15: @frett I wasn't current if I should make this a const value or not. Let me know and I can change it.

frett: I think leaving this here is fine, it's not really being used in multiple places

library/xml-model/src/test/java/org/cru/godtools/xml/model/TextTest.kt
```diff
@@ -25,4 +28,36 @@ class TextTest {
...
+        val text = Text(manifest, "text", startImage = "image.png", startImageSize = 65)
+        assertEquals(resource, text.startImage)
+        assertEquals(65, text.startImageSize)
+    }
+
+    @Test
+    fun testTextStartImageNull() {
+        val text = Text(mockManifest(), "text", startImage = null)
+        assertEquals(null, text.startImage)
+        assertEquals(40, text.startImageSize)
```
frett, line 45: you should use `DEFAULT_IMAGE_SIZE` here instead of `40`

frett: you will need to make `DEFAULT_IMAGE_SIZE` `internal` and add a `@VisibleForTesting` annotation to indicate that the increased visibility is only for testing purposes

library/xml-model/src/test/java/org/cru/godtools/xml/model/TextTest.kt
```diff
@@ -25,4 +28,36 @@ class TextTest {
...
+        val text = Text(manifest, "text", endImage = "image.png", endImageSize = 65)
+        assertEquals(resource, text.endImage)
+        assertEquals(65, text.endImageSize)
+    }
+
+    @Test
+    fun testTextEndImageNull() {
+        val text = Text(mockManifest(), "text", endImage = null)
+        assertEquals(null, text.endImage)
+        assertEquals(40, text.endImageSize)
```
frett, line 61: same thing about using the constant here

frett (changes requested):
just a couple minor comments about the test

frett (approved):

PR #1542, opened 2021-02-09
GT-1052 create Lesson renderer

frett:


ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonActivity.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    override fun updateToolbarTitle() {
+        title = ""
+    }
+    // endregion Toolbar
+
+    // region Pages
+    @Inject
+    lateinit var lessonPageAdapterFactory: LessonPageAdapter.Factory
+    private val lessonPageAdapter by lazy {
```
TheNoodleMoose, line 35: I looked at the [documentation](https://kotlinlang.org/docs/delegated-properties.html#lazy-properties) for ```by lazy``` but could use some explanation of the value of it being used here. Is it because the factory is not created initially, so we create it the one time, and then that value is remembered for all sequential calls so it doesn't create another factory?

ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonPageAdapter.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.cru.godtools.tool.lesson.ui.controller.LessonPageController
+import org.cru.godtools.tool.lesson.ui.controller.bindController
+import org.cru.godtools.xml.model.Manifest
+
+class LessonPageAdapter @AssistedInject internal constructor(
+    @Assisted lifecycleOwner: LifecycleOwner,
+    private val controllerFactory: LessonPageController.Factory
+) : SimpleDataBindingAdapter<LessonPageBinding>(lifecycleOwner), Observer<Manifest?> {
+    @AssistedFactory
+    interface Factory {
```
TheNoodleMoose, line 21: I haven't seen Factory before but searching the code base we use it in a few spots. What makes it different from a regular ```Class```/```Interface```? Or is it just a design pattern?

ui/base-tool/build.gradle
```diff
@@ -33,7 +33,7 @@ dependencies {
     implementation "org.ccci.gto.android:gto-support-eventbus:${deps.gtoSupport}"
     implementation "org.ccci.gto.android:gto-support-lottie:${deps.gtoSupport}"
     implementation "org.ccci.gto.android:gto-support-material-components:${deps.gtoSupport}"
-    implementation "org.ccci.gto.android:gto-support-picasso:${deps.gtoSupport}"
+    api "org.ccci.gto.android:gto-support-picasso:${deps.gtoSupport}"
```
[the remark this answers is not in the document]
frett: `api` vs `implementation` relates to transitive dependency visibility.

if: A depends on B and B depends on C

if the `C` dependency is defined as an `api` dependency, then `A` can access code out of `C`. if the dependency is an `implementation` dependency, the code in `C` is not visible to `A`.

The main reason for this distinction is to scope code so that you only have direct access to what is a public API vs an internal implementation detail.

ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonActivity.kt
frett: there are a few things I want to unpack here.
1. the `by` keyword identifies property delegation. this is a powerful Kotlin feature where you can create property delegates that implement some behavior around setting/getting a property, and all accesses of the property triggers this behavior.
2. the `lazy` delegate defers creation of the property value until it's first use
3. the `lessonPageAdapterFactory` on the previous line is a `lateinit var` and specifically being injected by Dagger once the activity starts.

The reason we need to use `lazy` here is because when the activity object is first created we don't have a value set in `lessonPageAdapterFactory` yet. this means that if we try to create the adapter at object creation it will fail with an NPE. The `lazy` delegate allows us to defer that creation to a latter point after `lessonPageAdapterFactory` has been initialized.

ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonPageAdapter.kt
frett: this is a Dagger feature, take a look at this documentation: https://dagger.dev/dev-guide/assisted-injection.html

I can explain in more detail how this works via a video call later if desired, it's not something that would be easy to capture in github comments

frett: `Factory` is just the name I gave the interface, it's not a kotlin keyword and doesn't have any significance.

TheNoodleMoose: Ah okay, gotcha. The ```@AssistedFactory``` is what threw me off.

TheNoodleMoose (approved):

PR #1573, opened 2021-03-02
GT-1108 use HackyDrawerLayout to workaround a framework crash relating to MotionEvents

frett:
Depends on:
- [x] https://github.com/CruGlobal/android-gto-support/pull/780

### DIFF
--- a/app/src/main/res/layout/activity_dashboard.xml
+++ b/app/src/main/res/layout/activity_dashboard.xml
@@ -9,7 +9,7 @@
         <variable name="selectedPage" type="LiveData&lt;org.cru.godtools.ui.dashboard.Page&gt;" />
     </data>
 
-    <androidx.drawerlayout.widget.DrawerLayout
+    <org.ccci.gto.android.common.androidx.drawerlayout.widget.HackyDrawerLayout
         android:id="@+id/drawer_layout"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/activity_dashboard.xml
```diff
@@ -9,7 +9,7 @@
         <variable name="selectedPage" type="LiveData&lt;org.cru.godtools.ui.dashboard.Page&gt;" />
     </data>
 
-    <androidx.drawerlayout.widget.DrawerLayout
+    <org.ccci.gto.android.common.androidx.drawerlayout.widget.HackyDrawerLayout
```
TheNoodleMoose, line 12: Could I ask what makes this a hacky implementation or should I just go take a look at it?

frett: it's an android bug with multi-touch motion events, sometimes invalid events are fed into the touch methods and it causes the app to crash. We are basically hacking around the framework bug by looking for those crashes and suppressing them so the app will continue to function.

frett: https://github.com/CruGlobal/android-gto-support/blob/master/gto-support-androidx-drawerlayout/src/main/java/org/ccci/gto/android/common/androidx/drawerlayout/widget/HackyDrawerLayout.kt#L16-L21

PR #1581, opened 2021-03-10
GT-1052 Minor adjustments for lessons

frett:


gyasistory (approved):

ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
```diff
@@ -238,7 +239,10 @@ class GodToolsShortcutManager @VisibleForTesting internal constructor(
 
         val dynamicShortcuts = withContext(ioDispatcher) {
             Query.select<Tool>()
-                .where(ToolTable.FIELD_ADDED.eq(true))
+                .where(
+                    ToolTable.FIELD_TYPE.`in`(*constants(Tool.Type.ARTICLE, Tool.Type.TRACT))
```
TheNoodleMoose, line 243: What is doing exactly?

frett: it's updating the SQL query that fetches favorited tools. we only want to include tract and article tool types, and exclude the new lesson tool type

frett: the actual SQL it generates is: `SELECT * FROM tools WHERE type in ('article', 'tract') AND added = true`

TheNoodleMoose: Ah cool, thanks!

PR #1591, opened 2021-03-16
GT-1119 Utilize WorkManager for updating shortcuts

frett:


gyasistory (approved):

ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
```diff
@@ -267,6 +257,11 @@ class GodToolsShortcutManager @VisibleForTesting internal constructor(
     }
     // endregion Update Existing Shortcuts
 
+    internal suspend fun refreshShortcutsNow() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) updateShortcuts()
+        updatePendingShortcuts()
+    }
+
```
TheNoodleMoose, line 264: What is the checking against exactly in order to know if it should refresh the shortcuts immediately?

app/src/main/java/org/cru/godtools/dagger/ServicesModule.kt
```diff
@@ -42,7 +40,7 @@ abstract class ServicesModule {
     companion object {
         @Provides
         @Singleton
-        fun workManager(@ApplicationContext context: Context, workerFactory: DaggerWorkerFactory): WorkManager {
+        fun workManager(@ApplicationContext context: Context, workerFactory: HiltWorkerFactory): WorkManager {
```
TheNoodleMoose, line 43: What's different between the ```DaggerWorkerFactory``` and the ```HiltWorkerFactory```? Or does it just let you use Hilt instead of Dagger?

build.gradle
```diff
@@ -60,6 +60,7 @@ buildscript {
                 core              : '1.3.2',
                 databinding       : deps.gradleAndroidPlugin,
                 fragment          : '1.3.1',
+                hilt              : '1.0.0-beta01',
```
TheNoodleMoose, line 63: If I remember correctly, ```hilt``` helps to remove/generate the code from ```Dagger``` but does the same thing(dependency injection)?

library/sync/src/main/java/org/cru/godtools/sync/SyncModule.kt
```diff
@@ -1,31 +1,14 @@
...
-    @Binds
-    @IntoMap
-    @WorkerKey(SyncFollowupWorker::class)
-    abstract fun syncFollowupWorker(factory: SyncFollowupWorker.Factory): AssistedWorkerFactory<SyncFollowupWorker>
-
-    @Binds
-    @IntoMap
-    @WorkerKey(SyncToolSharesWorker::class)
-    abstract fun syncToolSharesWorker(factory: SyncToolSharesWorker.Factory):
-        AssistedWorkerFactory<SyncToolSharesWorker>
```
TheNoodleMoose, line 30: Was this all able to be removed because you added ```@HiltWorker``` to the ```SyncFollowupWorker``` class below?

TheNoodleMoose (approved):

app/src/main/java/org/cru/godtools/dagger/ServicesModule.kt
frett: `DaggerWorkerFactory` is a `WorkerFactory` I implemented to allow Dagger to manage Workers.
`HiltWorkerFactory` is a `WorkerFactory` from AndroidX Hilt which we won't have to maintain.

build.gradle
frett: Hilt is part of Dagger, this hilt version deals with the AndroidX Hilt extensions, which provide the `HiltWorkerFactory`

library/sync/src/main/java/org/cru/godtools/sync/SyncModule.kt
frett: yeah, this moved it away from our custom `DaggerWorkerFactory` which required Multibinding to work to `HiltWorkerFactory` which automatically generates that logic for us.

ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
frett: this method allows us to trigger an immediate refresh of all shortcuts. right now the only place this is triggered from is the `UpdateShortcutsWorker`, which is only scheduled after the user changes their device language.

The reason we have this execution path enabled is because we use the device language to translate the tool names for shortcuts, so when the device language changes we need to update the shortcuts to use the new device language.

frett: Originally the system locale update broadcast would directly do this work and not rely on work manager to schedule it, but there are timing constraints that come with updating immediately while processing the broadcast.

Broadcast Receivers start execution on the main thread, they can offload processing to a background thread with `goAsync()`, but they additionally have a 10 second time limit to complete execution during. The recommended solution for any heavy background processing is to use Work Manager to schedule the work to be done at a later time, with assurances that it will fully execute at some point in the future.

PR #1596, opened 2021-03-19
GT-978 Skip the Categories UI when it's not necessary

frett:


ui/article-renderer/src/main/java/org/cru/godtools/article/ui/ArticlesActivity.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    override fun onOptionsItemSelected(item: MenuItem) = when {
+        item.itemId == android.R.id.home && supportFragmentManager.backStackEntryCount > 0 -> {
+            supportFragmentManager.popBackStack()
+            true
+        }
+        else -> super.onOptionsItemSelected(item)
+    }
+
+    override fun onCategorySelected(category: Category?) = showFragment(articlesFragment(category), true)
```
TheNoodleMoose, line 56: I saw there was logic in the ```CategoriesActivity``` to only load the fragment if needed, but I guess it still had to run logic to check if it was needed. This solution is better because it's only loaded once the category is selected and the logic to check is no longer needed right?

TheNoodleMoose (approved):
So you consolidated the ArticlesActivity and CategoriesActivity into one activity so the categories UI only gets loaded when needed? Are categories closely coupled with articles? Could you explain where a category fragment may be seen in the app?

ui/article-renderer/src/main/java/org/cru/godtools/article/ui/ArticlesActivity.kt
frett: the `loadFragmentIfNecessary()` logic would create the fragment the first time the activity is created, but not recreate it if the activity was recreated.

This logic still exists [here](https://github.com/CruGlobal/godtools-android/pull/1596/files#diff-bf7f611f9f31954e845ae85b090a1a5e3987f71f0f7a6ce9dab07a23792e87d4R78-R86) but now intelligently chooses what that initial fragment being loaded is

frett:
These activities/fragments are used within the article renderer, "Questions About God" is the only articles tool that currently exists.

The normal flow is:
- open Questions About God which will show you a list of article categories
- select a category which will then show you a list of articles
- select an article which will then show you the article content

Ann wanted to be able to skip the categories screen when there wasn't enough articles to justify categories

PR #1597, opened 2021-03-19
GT-1052 Adjust a couple things in the Lessons Renderer

frett:


ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonActivity.kt
```diff
@@ -82,9 +85,12 @@ class LessonActivity :
         pages.registerOnPageChangeCallback(object : ViewPager2.OnPageChangeCallback() {
             override fun onPageScrollStateChanged(state: Int) {
                 if (state == SCROLL_STATE_IDLE) {
-                    val currentItemId = dataModel.pages.value?.getOrNull(pages.currentItem)?.id
-                    dataModel.visiblePages.removeAll { it != currentItemId }
-                    updateProgressIndicator()
```
TheNoodleMoose, line 87: Why is ```updateProgressIndicator()``` no longer needed? I know it's related to using coroutines but not exactly sure why.

TheNoodleMoose (approved):

ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonActivity.kt
frett: actually it wasn't due to coroutines, I don't remember the exact reason I removed it, but it was probably related to it being triggered multiple times from multiple locations and not being necessary here. I'm going to revisit this before I merge to make sure I didn't miss something

frett: I took a look, we call `updateProgressIndicator` in 2 locations,
1. when a page selected event is fired (user scrolls to another page)
   - This fires off before the scroll state change event
   - This is supposed to fire off on any change to the `selectedItem`
2. when the pages in the data model are updated
   - This is to capture the total number of pages visible (ignoring hidden pages)

The logic that got moved to a coroutine makes a change to visiblePages, which cascades an update into the pages in the data model, which automatically triggers `updateProgressIndicator()`. This means calling it a second time here would have no effect

PR #1614, opened 2021-03-26
GT-1074 Support rendering text styles

frett:


gyasistory (approved):

ui/base-tool/src/main/java/org/cru/godtools/base/tool/model/view/TextViewUtils.kt
```diff
@@ -10,10 +14,16 @@ import org.cru.godtools.xml.model.defaultTextColor
...
+import splitties.bitflags.withFlag
 
 internal fun Text?.bindTo(view: TextView, textSize: Float? = null, @ColorInt defaultTextColor: Int? = null) {
     view.text = this?.text
-    view.typeface = this?.getTypeface(view.context)
+    view.setTypeface(this?.getTypeface(view.context), this?.typefaceStyle ?: NORMAL)
+    view.paintFlags = view.paintFlags.let {
+        if (Text.Style.UNDERLINE in this?.textStyles.orEmpty()) it.withFlag(UNDERLINE_TEXT_FLAG)
+        else it.minusFlag(UNDERLINE_TEXT_FLAG)
+    }
```
TheNoodleMoose, line 26: I see that ```splitties``` is adding or removing the underline. Is there not an easy way to do this natively and that's why ```splitties``` is needed?

frett: Unfortunately there isn't a simple API to toggle the underline style.

This logic utilizes setting/unsetting the appropriate paint flag when we want to render an underline vs remove the underline. The `Splitties` bit flag library provides us some extension methods to toggle the appropriate bit flags.

without splitties the logic would look something like this:
```kotlin
    view.paintFlags = view.paintFlags.let {
        if (Text.Style.UNDERLINE in this?.textStyles.orEmpty()) it or UNDERLINE_TEXT_FLAG
        else it and ~UNDERLINE_TEXT_FLAG
    }
```

I prefer using the splitties extension methods because it makes it a bit more readable as to what flags are being toggled

frett: https://github.com/LouisCAD/Splitties#splitties

https://github.com/LouisCAD/Splitties/blob/main/modules/bitflags/src/commonMain/kotlin/splitties/bitflags/BitFlags.kt#L18-L19

PR #1635, opened 2021-04-09
GT-1140 Update Firebase launch event

frett:
https://jira.cru.org/browse/GT-1140

### DIFF
--- a/library/analytics/src/main/java/org/cru/godtools/analytics/model/LaunchAnalyticsActionEvent.kt
+++ b/library/analytics/src/main/java/org/cru/godtools/analytics/model/LaunchAnalyticsActionEvent.kt
@@ -1,13 +1,16 @@
 package org.cru.godtools.analytics.model
 
-private const val ACTION_LAUNCH = "App Launch"
-private const val FIREBASE_EVENT_LAUNCH_PREFIX = "launch_"
+private const val ACTION_LAUNCH_PREFIX = "launch_"
+private const val ACTION_LAUNCH_AT_LEAST_PREFIX = "launch_atleast_"
 
-class LaunchAnalyticsActionEvent(private val launches: Int) : AnalyticsActionEvent(action = ACTION_LAUNCH) {
-    override fun isForSystem(system: AnalyticsSystem) = when (launches) {
-        1, 2, 3, 5, 10 -> system == AnalyticsSystem.FIREBASE
-        else -> false
-    }
+class LaunchAnalyticsActionEvent(private val launches: Int) : AnalyticsActionEvent(action = launches.buildAction()) {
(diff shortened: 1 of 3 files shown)


library/analytics/src/main/java/org/cru/godtools/analytics/model/LaunchAnalyticsActionEvent.kt
```diff
@@ -1,13 +1,16 @@
...
-        else -> false
-    }
+class LaunchAnalyticsActionEvent(private val launches: Int) : AnalyticsActionEvent(action = launches.buildAction()) {
+    override fun isForSystem(system: AnalyticsSystem) = launches > 0 && system == AnalyticsSystem.FIREBASE
+}
 
-    override val firebaseEventName get() = "$FIREBASE_EVENT_LAUNCH_PREFIX$launches"
+private fun Int.buildAction() = when (this) {
+    1, 2 -> "$ACTION_LAUNCH_PREFIX$this"
+    in 3..4 -> "${ACTION_LAUNCH_AT_LEAST_PREFIX}3"
```
TheNoodleMoose, line 12: Why is one wrapped with ```{ }``` while the other is not?

library/analytics/src/main/java/org/cru/godtools/analytics/model/LaunchAnalyticsActionEvent.kt
```diff
@@ -1,13 +1,16 @@
...
+    in 5..9 -> "${ACTION_LAUNCH_AT_LEAST_PREFIX}5"
```
TheNoodleMoose, line 13: I'm assuming we can just ignore these ```magic number``` warnings?

TheNoodleMoose (approved):

library/analytics/src/main/java/org/cru/godtools/analytics/model/LaunchAnalyticsActionEvent.kt
frett: yeah, I turned on some automated scanning, I'm still figuring out what is useful and then what I need to disable

library/analytics/src/main/java/org/cru/godtools/analytics/model/LaunchAnalyticsActionEvent.kt
frett: without `{ }` it would include `3` in the variable name it's trying to resolve

TheNoodleMoose: Gotcha, figured but wanted to make sure!

PR #1640, opened 2021-04-09
GT-1151 Allow text-scale to be set on a manifest and pages

frett:


library/xml-model/src/test/java/org/cru/godtools/xml/model/tract/TractPageTest.kt
```diff
@@ -59,5 +61,15 @@ class TractPageTest {
         assertEquals(Color.BLUE, TractPage(Manifest(cardBackgroundColor = Color.BLUE)).cardBackgroundColor)
     }
 
+    @Test
+    fun testTextScale() {
```
TheNoodleMoose, line 65: Should this test use the string format for the test name(ex. ```fun `parseManifest - Lesson Manifest`()```), or are we only using that on certain tests?

TheNoodleMoose (approved):

library/xml-model/src/test/java/org/cru/godtools/xml/model/tract/TractPageTest.kt
frett: I haven't established a best practice with it yet, I've been most experimenting with the space separate name style

frett: also, ```fun `testTextScale`()``` will cause a line error because the ``` ` ``` character isn't necessary for the name

PR #1654, opened 2021-04-21
GT-1155 Constrain the lifecycle of non-visible lesson pages

frett:
Update the lesson renderer so that each page has it's own `LifecycleOwner`, and any page that isn't currently visible has it's `LifecycleOwner` restricted from reaching the `RESUMED` state.

Dependent on:
- [x] https://github.com/CruGlobal/android-gto-support/pull/819
- [x] https://github.com/CruGlobal/android-gto-support/pull/823


gyasistory (approved):

ui/base/src/main/java/org/cru/godtools/base/ui/databinding/YouTubePlayerViewBindingAdapter.kt
```diff
@@ -11,6 +11,14 @@ import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils.YouTube
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.YouTubePlayerView
 import org.cru.godtools.base.ui.R
 
+@BindingAdapter("pauseVideoWhen")
+internal fun YouTubePlayerView.pauseVideoWhen(pause: Boolean) {
```
TheNoodleMoose, line 15: So before you added this, would it continue to play the video even if you moved on to the next page?

ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonPageAdapter.kt
```diff
@@ -43,21 +47,44 @@ class LessonPageAdapter @AssistedInject internal constructor(
...
 
     // region Lifecycle
-    override fun onCreateViewDataBinding(parent: ViewGroup, viewType: Int) =
-        LessonPageBinding.inflate(LayoutInflater.from(parent.context), parent, false).apply {
-            callbacks = _callbacks
-            bindController(controllerFactory)
+    override fun onAttachedToRecyclerView(recyclerView: RecyclerView) {
+        primaryItemObserver = onUpdatePrimaryItem(recyclerView) { primary, previous ->
+            previous?.binding?.controller?.lifecycleOwner?.apply { maxState = minOf(maxState, Lifecycle.State.STARTED) }
+            primary?.binding?.controller?.lifecycleOwner?.apply { maxState = maxOf(maxState, Lifecycle.State.RESUMED) }
```
TheNoodleMoose, line 60: Could you explain what the ```maxState``` is? Also to confirm what I "believe" this is doing, the ```primary``` and ```previous``` refers to the current and last page, and you are binding a specific state to the individual lifecycles of those pages?

TheNoodleMoose (approved):

ui/base/src/main/java/org/cru/godtools/base/ui/databinding/YouTubePlayerViewBindingAdapter.kt
frett: yes and no.

last week the video would have stopped because the RecyclerView only kept the current page rendered, this caused other UX issues because images wouldn't be initially loaded when you navigated to a page for the first time.

To address that issue I increased the number of adjacent pages to pre-render. This caused the video to continue to play while it is on an adjacent page because it wasn't being destroyed immediately

I had used the AndroidX Lifecycle framework elsewhere to work around this issue, so I did the work to instrument the lesson renderer to use the lifecycle framework and keep it consistent with the other renderers

ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonPageAdapter.kt
frett: relevant class: https://github.com/CruGlobal/android-gto-support/blob/master/gto-support-androidx-lifecycle/src/main/java/org/ccci/gto/android/common/androidx/lifecycle/ConstrainedStateLifecycleOwner.kt

Basically the `ConstrainedStateLifecycleOwner` has a `maxState` property the limits the maximum state the lifecycle can reach. Whenever the "primaryItem" is updated, the previous primary item is constrained to a max state of `STARTED`, and the new primary item is constrained to a max state of `RESUMED`.

I use the `minOf` and `maxOf` functions to not potentially pave over a more restrictive `maxState` set elsewhere, although I don't believe there are any code paths that would make that necessary

PR #1676, opened 2021-05-03
GT-1167 Fix a crash when AemArticleManager is created on the Main Thread

frett:


ui/article-aem-renderer/src/main/java/org/cru/godtools/article/aem/db/AemImportDao.kt
```diff
@@ -12,9 +12,8 @@ import org.cru.godtools.article.aem.model.AemImport.AemImportArticle
 
 @Dao
 interface AemImportDao {
-    @get:WorkerThread
```
TheNoodleMoose, line 15: This annotation was what was allowing it to run on the main thread? Is the issue because it's an IO operation and could take some time to finish, which is why doing it on the main thread would be a bad idea?

TheNoodleMoose (approved):

ui/article-aem-renderer/src/main/java/org/cru/godtools/article/aem/service/AemArticleManager.kt
```diff
@@ -156,9 +156,11 @@ class AemArticleManager @VisibleForTesting internal constructor(
...
 
     init {
-        aemDb.aemImportDao().all
-            .filter { it.isStale() }
-            .forEach { coroutineScope.launch { syncAemImport(it.uri, false) } }
+        coroutineScope.launch {
+            aemDb.aemImportDao().getAll()
+                .filter { it.isStale() }
+                .forEach { launch { syncAemImport(it.uri, false) } }
+        }
```
TheNoodleMoose, line 163: So before only the ```syncAemImport``` was wrapped in coroutine scope, but now all of this is. Was there a chance before it was wrapped in coroutine scope, that it would try to run ```syncAemImport``` before ```getAll``` finished?

ui/article-aem-renderer/src/main/java/org/cru/godtools/article/aem/db/AemImportDao.kt
frett: The `@WorkerThread`, `@UiThread`, `@MainThread` are purely used by lint rules to try and detect improper thread usage of methods/objects/etc. They aren't foolproof though.

The main problem was that this was being called during creation of `AemArticleManager`, which could happen on the main thread in certain scenarios.

ui/article-aem-renderer/src/main/java/org/cru/godtools/article/aem/service/AemArticleManager.kt
frett: nope, `getAll()` would have to return before either the `filter` statement or the `syncAemImport` method even started executing.

The entire block is wrapped in a `launch` because it needs to process on a background thread, then I launch separate coroutines for each found stale `AemImport` so that they can execute in parallel on background threads.

ui/article-aem-renderer/src/main/java/org/cru/godtools/article/aem/db/AemImportDao.kt
frett: converting the `Dao`/`Repository` methods to be suspend methods make them safe to call on any thread.

ui/article-aem-renderer/src/main/java/org/cru/godtools/article/aem/service/AemArticleManager.kt
TheNoodleMoose: Gotcha, thanks!

PR #1744, opened 2021-06-17
GT-1217 Utilize shared tool model

frett:
This updated the Android app to use the new shared multiplatform GodTools tool parser library.

This is primarily about getting it to functionally work, there will be several subsequent PRs that utilize some updates to the shared library.


frett:
coverage dropped because most of our tests were for the parsed model which is now an external library

PR #1801, opened 2021-07-23
GT-1164 GT-1165 Implement the Multiselect element for tools

frett:
This PR adds renderer support for the Multiselect element.

You can test out this element in the lesson on the stage environment

### DIFF
--- a/ui/base-tool/src/main/java/org/cru/godtools/base/tool/ui/controller/TextController.kt
+++ b/ui/base-tool/src/main/java/org/cru/godtools/base/tool/ui/controller/TextController.kt
@@ -18,6 +18,10 @@ internal class TextController private constructor(
     @AssistedFactory
     interface Factory : BaseController.Factory<TextController>
 
+    init {
+        binding.controller = this
+    }
+
     public override fun onBind() {
         super.onBind()
         binding.model = model
--- a/ui/base-tool/src/main/java/org/cru/godtools/base/tool/viewmodel/ToolStateHolder.kt
+++ b/ui/base-tool/src/main/java/org/cru/godtools/base/tool/viewmodel/ToolStateHolder.kt
@@ -1,0 +1,13 @@
+package org.cru.godtools.base.tool.viewmodel
+
+import androidx.lifecycle.SavedStateHandle
+import androidx.lifecycle.ViewModel
+import org.cru.godtools.tool.state.State
+
+private const val STATE_TOOL_STATE = "toolState"
+
+class ToolStateHolder(private val savedState: SavedStateHandle) : ViewModel() {
+    val toolState by lazy {
+        savedState[STATE_TOOL_STATE] ?: State().also { savedState[STATE_TOOL_STATE] = it }
+    }
+}
--- a/ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonActivity.kt
+++ b/ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonActivity.kt
@@ -17,6 +17,7 @@ import org.cru.godtools.base.tool.activity.BaseSingleToolActivity
 import org.cru.godtools.base.tool.activity.BaseSingleToolActivityDataModel
 import org.cru.godtools.base.tool.model.Event
 import org.cru.godtools.base.tool.service.ManifestManager
+import org.cru.godtools.base.tool.viewmodel.ToolStateHolder
 import org.cru.godtools.download.manager.GodToolsDownloadManager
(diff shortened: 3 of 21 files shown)


ui/base-tool/src/main/java/org/cru/godtools/base/tool/ui/controller/TextController.kt
```diff
@@ -18,6 +18,10 @@ internal class TextController private constructor(
     @AssistedFactory
     interface Factory : BaseController.Factory<TextController>
 
+    init {
+        binding.controller = this
+    }
```
TheNoodleMoose, line 23: Is this needed for overriding ```textIsSelectable```?

ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonActivity.kt
```diff
@@ -80,7 +82,7 @@ class LessonActivity :
     // region Pages
     @Inject
     lateinit var lessonPageAdapterFactory: LessonPageAdapter.Factory
-    private val lessonPageAdapter by lazy { lessonPageAdapterFactory.create(this, this) }
+    private val lessonPageAdapter by lazy { lessonPageAdapterFactory.create(this, this, toolState.toolState) }
```
TheNoodleMoose, line 85: ```by lazy``` simply calls the method when ```get()``` first runs and then returns the result of that first call if ```get()``` gets called again? Is that to just prevent unnecessary calls to that method?  And if i'm understanding the syntax correctly(I could be way off here haha):
```kotlin
private val lessPageAdapter = lessPageAdapterFactory.create(this, this, toolState.toolState)
```
is the same as 
```kotlin
private val lessonPageAdapter
        get() = lessonPageAdapterFactory.create(this, this, toolState.toolState)
```
which is the same as the ```by lazy``` syntax except for the additional things mentioned above?

ui/base-tool/src/main/java/org/cru/godtools/base/tool/viewmodel/ToolStateHolder.kt
```diff
@@ -0,0 +1,13 @@
+package org.cru.godtools.base.tool.viewmodel
+
+import androidx.lifecycle.SavedStateHandle
+import androidx.lifecycle.ViewModel
+import org.cru.godtools.tool.state.State
+
+private const val STATE_TOOL_STATE = "toolState"
+
+class ToolStateHolder(private val savedState: SavedStateHandle) : ViewModel() {
+    val toolState by lazy {
```
TheNoodleMoose, line 10: So will these quizzes be in multiple parts of the app, hence why you added the tool state to different activities, controllers, etc? And each will instantiate its own individual state.

TheNoodleMoose (approved):

gyasistory (approved):
@frett I was

ui/base-tool/src/main/java/org/cru/godtools/base/tool/ui/controller/TextController.kt
frett: yes, because in the layout we need to reference the controller now

ui/lesson-renderer/src/main/java/org/cru/godtools/tool/lesson/ui/LessonActivity.kt
frett: you got the syntax right. We want to reuse the same `lessonPageAdapter` for the lifetime of the Activity, which is the reason we don't use `get() =`. But we also can't initialize the adapter on object creation because `lessonPageAdapterFactory` won't be initialized, plus the parameters sent to `create` won't be in the correct state until the activity is actually started.

TheNoodleMoose: Sweet, makes sense! I feel like the syntax is beginning to click more and more.

ui/base-tool/src/main/java/org/cru/godtools/base/tool/viewmodel/ToolStateHolder.kt
frett: yes, `ToolState` is a new concept that extends beyond just quizzes. it will allow content elements to store arbitrary data and other content elements to change their behavior on what data is stored.

The scope of `ToolState` is across the entirety of the a single tool during a session using it. So state stored on the first page of a tool will be able to be referenced on the last page of that same tool. But if the user backs out of a tool and re-enters it the state will be reset.

PR #1863, opened 2021-09-01
record the number of launches in a firebase user property

frett:
Dependent on:
- [ ] https://github.com/CruGlobal/android-gto-support/pull/960


frett:
Go ahead and review, but this may be redundant since you basically accomplish the same thing for IAM using Audiences in Firebase.

PR #1874, opened 2021-09-07
GT-1275 Fix multiselect layout when rendering a single column

frett:
when the multiselect option is created the `layout_constraintWidth_percentage` wouldn't initially be set. This would cause all options to be rendered in a single row.

Once databinding runs it would attempt to set the percentage to 100%, but would short-circuit because the unset internal value is 100%. This would result in the view not being re-laid out.


frett:
This is a quick fix that works as is, I'm still doing a bit of research to see if there is a more correct way to accomplish this.

frett:
I think I found the attribute I was missing that will correctly configure the ConstraintLayout params

TheNoodleMoose (approved):

PR #1887, opened 2021-09-20
Use coroutines Flow to monitor for primary/parallel language changes

frett:
- expose the primary and parallel languages as coroutine flows
- update the GodToolsShortcutManager to monitor primary & parallel language changes via coroutine flows
- remove unused register/unregister listener methods

### DIFF
--- a/library/base/src/main/java/org/cru/godtools/base/Settings.kt
+++ b/library/base/src/main/java/org/cru/godtools/base/Settings.kt
@@ -209,12 +209,4 @@ class Settings @Inject internal constructor(
         trackFirstLaunchVersion()
     }
     // endregion Launch tracking
-
-    fun registerOnSharedPreferenceChangeListener(listener: SharedPreferences.OnSharedPreferenceChangeListener?) {
-        prefs.registerOnSharedPreferenceChangeListener(listener)
-    }
-
-    fun unregisterOnSharedPreferenceChangeListener(listener: SharedPreferences.OnSharedPreferenceChangeListener?) {
-        prefs.unregisterOnSharedPreferenceChangeListener(listener)
-    }
 }
--- a/library/base/src/main/java/org/cru/godtools/base/Settings.kt
+++ b/library/base/src/main/java/org/cru/godtools/base/Settings.kt
@@ -13,9 +13,12 @@ import dagger.hilt.android.qualifiers.ApplicationContext
 import java.util.Locale
 import javax.inject.Inject
 import javax.inject.Singleton
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.map
 import org.ccci.gto.android.common.androidx.lifecycle.getBooleanLiveData
 import org.ccci.gto.android.common.androidx.lifecycle.getIntLiveData
 import org.ccci.gto.android.common.androidx.lifecycle.getStringLiveData
+import org.ccci.gto.android.common.kotlin.coroutines.getStringFlow
 import org.ccci.gto.android.common.okta.oidc.clients.sessions.oktaUserId
 import org.ccci.gto.android.common.util.toLocale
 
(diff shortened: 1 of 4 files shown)


library/base/src/main/java/org/cru/godtools/base/Settings.kt
```diff
@@ -74,6 +77,9 @@ class Settings @Inject internal constructor(
         prefs.getStringLiveData(PREF_PRIMARY_LANGUAGE, defaultLanguage.toLanguageTag()).distinctUntilChanged()
             .map { it?.toLocale() ?: defaultLanguage.also { primaryLanguage = it } }
     }
+    val primaryLanguageFlow
+        get() = prefs.getStringFlow(PREF_PRIMARY_LANGUAGE, defaultLanguage.toLanguageTag()).distinctUntilChanged()
+            .map { it?.toLocale() ?: defaultLanguage.also { primaryLanguage = it } }
```
TheNoodleMoose, line 82: So is the ```distinctUntilChanged()``` what is listening for changes to the ```getStringFlow(PREF_PRIMARY_LANGUAGE, defaultLanguage.toLanguageTag())``` flow and then updating this flow accordingly? And this flow just returns that value that is set and ```val primaryLanguage``` is the value that is actually used?

library/base/src/main/java/org/cru/godtools/base/Settings.kt
```diff
@@ -86,6 +92,9 @@ class Settings @Inject internal constructor(
         prefs.getStringLiveData(PREF_PARALLEL_LANGUAGE, null).distinctUntilChanged()
             .map { it?.toLocale() }
     }
+    val parallelLanguageFlow
+        get() = prefs.getStringFlow(PREF_PARALLEL_LANGUAGE, null).distinctUntilChanged()
```
TheNoodleMoose, line 96: the ```getStringFlow``` here has ```null``` as the second value because it isn't required to have one set right? While the ```getStringFlow``` for the primary language is passed a default language because one has to be set no matter what.

ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
```diff
@@ -178,9 +168,13 @@ class GodToolsShortcutManager @VisibleForTesting internal constructor(
...
     @VisibleForTesting
-    @OptIn(ObsoleteCoroutinesApi::class)
+    @OptIn(ExperimentalCoroutinesApi::class, ObsoleteCoroutinesApi::class)
     internal val updatePendingShortcutsActor = coroutineScope.actor<Unit>(capacity = CONFLATED) {
-        channel.consumeAsFlow().conflate().collectLatest {
+        merge(
+            settings.primaryLanguageFlow,
+            settings.parallelLanguageFlow,
+            channel.consumeAsFlow()
+        ).conflate().collectLatest {
```
TheNoodleMoose, line 177: Or does ```collectLatest``` listen for the changes to the flows? And I assume we merge them here in order for ```collectLatest``` to listen for changes to either flow? And then we run these callbacks accordingly after a change to the flow has occurred.

TheNoodleMoose (approved):
Looks good 🙂 Just a few questions I had.

library/base/src/main/java/org/cru/godtools/base/Settings.kt
frett: 1. `getStringFlow()` will emit the current value of the shared preference as it is changed internally.
2. `distinctUntilChanged()` will then mutate the flow and only emit the value when it changed from the previous value.
3. then the `String` value that is being emitted is mutated again to convert it to a `Locale` object
4. If that `Locale` object is null, we then emit `defaultLanguage` instead, and in the process update the stored `primaryLanguage`

library/base/src/main/java/org/cru/godtools/base/Settings.kt
frett: correct, `primaryLanguage` will always be non-null, `parallelLanguage` can be null

ui/shortcuts/src/main/java/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
frett: `conflate` and `collectLatest` kind of do the same thing, but at different points during execution.

- `collectLatest` will start processing the most recent value, if a new value shows up before it finishes, it will cancel the previous execution and start processing the new value
- `conflate` will only buffer a single value for execution, if a new value shows up it will drop the old value and buffer the new one

library/base/src/main/java/org/cru/godtools/base/Settings.kt
frett: in reality I should probably move the `distinctUntilChanged()` flow operator to be last, that way it will eliminate a duplicate value if the underlying `primaryLanguage` preference is ever set to null

PR #1998, opened 2021-11-22
GT-1363 Separate Dispatcher logic from ShortcutManager

frett:
This splits some of the ShortcutManager functionality to make it easier to test certain behaviors. This is in preparation for Kotlin Coroutines 1.6.0 which changes how coroutines tests run


frett:
Test coverage went down because we are no longer hitting some logic that we never verified the behavior of

PR #2021, opened 2021-12-03
GT-1368 Hide tools that are set to hidden

frett:


library/db/src/main/kotlin/org/keynote/godtools/android/db/GodToolsDatabase.kt
```diff
@@ -77,6 +79,10 @@ class GodToolsDatabase @Inject internal constructor(@ApplicationContext private
                     }
                     43 -> db.execSQL(ToolTable.SQL_V43_ALTER_CATEGORY)
                     44 -> db.execSQL(TrainingTipTable.SQL_V44_CREATE_TRAINING_TIPS)
+                    45 -> {
```
TheNoodleMoose, line 82: I think I understand what's happening here, but could use an explanation just to make sure 🙂

library/db/src/main/kotlin/org/keynote/godtools/android/db/ToolMapper.kt
```diff
@@ -55,5 +57,6 @@ internal object ToolMapper : BaseMapper<Tool>() {
         defaultOrder = c.getInt(COLUMN_DEFAULT_ORDER, 0)
         order = c.getInt(COLUMN_ORDER, Int.MAX_VALUE)
         isAdded = getBool(c, COLUMN_ADDED, false)
+        isHidden = getBool(c, COLUMN_HIDDEN, false)
```
TheNoodleMoose, line 60: So I tracked this all the way down to https://github.com/CruGlobal/android-gto-support/blob/28f0bc272625df820e0cfe4a75c08e3e86969836/gto-support-db/src/main/java/org/ccci/gto/android/common/db/util/CursorUtils.java#L26 and just wanted to double-check something. I see in the ```tool_hidden_true.json``` you have ```"attr-hidden": "TruE"```. Does this get that value of ```TruE``` and then convert it to a boolean, or is there somewhere that does that and then passes it to this?

TheNoodleMoose (approved):
Looks good, just a few questions 🙂

gyasistory (approved):

app/src/main/java/org/cru/godtools/ui/tools/ToolsFragmentDataModel.kt
```diff
@@ -35,7 +35,7 @@ class ToolsFragmentDataModel @Inject constructor(private val dao: GodToolsDao, s
                 MODE_LESSONS -> constants(Tool.Type.LESSON)
                 else -> constants(Tool.Type.TRACT, Tool.Type.ARTICLE, Tool.Type.CYOA)
             }
-        )
+        ).and(ToolTable.FIELD_HIDDEN.ne(true))
```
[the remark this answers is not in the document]
frett: Yeah

library/db/src/main/kotlin/org/keynote/godtools/android/db/ToolMapper.kt
frett: The jaonapi converter takes care of converting Json to a boolean, which is handled in a different part of the app

library/db/src/main/kotlin/org/keynote/godtools/android/db/GodToolsDatabase.kt
frett: Database migration to update the db definition for users that have previously installed the app.

TheNoodleMoose: Okay, I was right 🙂 I was going to use the words "database migration", but didn't want to be completely wrong lol. I remember doing something similar in the missionhub app, but it was using redux instead of a database.

PR #2022, opened 2021-12-03
GT-1325 Add background support to content pages

frett:
- add horizontal padding for content pages
- add support for background color and images on content pages

### DIFF
--- a/ui/cyoa-renderer/src/main/res/layout/cyoa_page_content.xml
+++ b/ui/cyoa-renderer/src/main/res/layout/cyoa_page_content.xml
@@ -10,10 +10,10 @@
     </data>
 
     <org.ccci.gto.android.common.widget.HackyNestedScrollView
-        android:id="@+id/initial_page_content"
-        android:layout_marginTop="@{contentInsets.top}"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
+        android:layout_marginHorizontal="@dimen/tool_content_margin"
+        android:layout_marginTop="@{contentInsets.top}"
         android:fillViewport="true"
         android:layoutDirection="@{BaseUtilsKt.getLayoutDirection(page)}">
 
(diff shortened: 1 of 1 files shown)


ui/cyoa-renderer/src/main/res/layout/cyoa_page_content.xml
```diff
@@ -1,26 +1,44 @@
...
+    <FrameLayout
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        android:fillViewport="true"
-        android:layoutDirection="@{BaseUtilsKt.getLayoutDirection(page)}">
+        android:background="@{PageKt.getBackgroundColor(page)}">
 
-        <LinearLayout
-            android:id="@+id/content"
+        <org.cru.godtools.base.tool.widget.SimpleScaledPicassoImageView
```
TheNoodleMoose, line 19: So ```FrameLayout``` handles the background color while ```SimpleScaledPicassoImageView``` handles the background image if there is one. That way we can have a background color of, let's say, blue, and then a scaled background image and be able to see both at the same time. I guess you could technically scale the image to cover the whole screen if you wanted right?

TheNoodleMoose (approved):

ui/cyoa-renderer/src/main/res/layout/cyoa_page_content.xml
frett: yes, we want to layer our 2 background layers here, and then put the content on top of it. `FrameLayout` is a very lightweight android `ViewGroup`. it can have multiple children that are drawn on top of each other from first to last. so we draw any defined background image as the first child of the FrameLayout, and then the content stack is drawn on top of the background as the second child in the `FrameLayout`.

once we have that to get the backgroundColor behind the backgroundImage we attach it to the `FrameLayout` as the wrapper element.

PR #2041, opened 2021-12-15
GT-1327 Implement up navigation respecting the parentPage property for pages

frett:
Dependent on:
- [x] https://github.com/CruGlobal/android-gto-support/pull/1080


frett:
There is an interaction issue between dagger and jacoco that prevents coverage from being reported on hilt annotated classes that have bytecode updates (e.g. `CyoaActivity`)

TheNoodleMoose (approved):

PR #2067, opened 2022-01-06
GT-1349 Use WorkManager to update tools when the user is online

frett:
- support deferred initialization of WorkManager
- keep existing work to avoid reseting backoff timers
- when syncing tools fails, reschedule with workmanager
- use WorkManager to retry languages sync the next time the device is online
- use WorkManager to reschedule a tool download if it failed

### DIFF
--- a/app/src/main/java/org/cru/godtools/GodToolsApplication.kt
+++ b/app/src/main/java/org/cru/godtools/GodToolsApplication.kt
@@ -2,21 +2,26 @@ package org.cru.godtools
 
 import android.app.Application
 import android.content.Context
+import android.util.Log
 import androidx.appcompat.app.AppCompatDelegate
+import androidx.hilt.work.HiltWorkerFactory
+import androidx.work.Configuration
 import com.google.android.instantapps.InstantApps
 import com.google.android.play.core.missingsplits.MissingSplitsManagerFactory
 import com.google.android.play.core.splitcompat.SplitCompat
 import com.google.firebase.crashlytics.FirebaseCrashlytics
+import dagger.Lazy
 import dagger.hilt.android.HiltAndroidApp
 import java.util.Locale
 import javax.inject.Inject
+import org.ccci.gto.android.common.androidx.work.TimberLogger
 import org.ccci.gto.android.common.dagger.eager.EagerSingletonInitializer
 import org.ccci.gto.android.common.firebase.crashlytics.timber.CrashlyticsTree
 import org.ccci.gto.android.common.util.LocaleUtils
 import timber.log.Timber
 
 @HiltAndroidApp
-open class GodToolsApplication : Application() {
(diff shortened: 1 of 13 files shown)


library/sync/src/main/java/org/cru/godtools/sync/work/SyncToolSharesWorker.kt
```diff
@@ -13,10 +13,10 @@ import org.cru.godtools.sync.task.ToolSyncTasks
 private const val WORK_NAME = "SyncToolShares"
 
 internal fun WorkManager.scheduleSyncToolSharesWork() =
-    enqueueUniqueWork(WORK_NAME, ExistingWorkPolicy.REPLACE, SyncWorkRequestBuilder<SyncToolSharesWorker>().build())
+    enqueueUniqueWork(WORK_NAME, ExistingWorkPolicy.KEEP, SyncWorkRequestBuilder<SyncToolSharesWorker>().build())
```
TheNoodleMoose, line 16: What was the change to ```ExistingWorkPolicy``` needed for? Does it have something to do with the top-level ```getWorkManagerConfiguration``` we now have?

app/src/main/java/org/cru/godtools/GodToolsApplication.kt
```diff
@@ -60,4 +65,14 @@ open class GodToolsApplication : Application() {
         }
         Timber.plant(CrashlyticsTree())
     }
+
+    // region WorkManager Configuration.Provider
+    init {
+        TimberLogger(Log.ERROR).install()
```
TheNoodleMoose, line 71: I'm trying to figure out what this is doing exactly. I have a few ideas, but not sure which is right.

TheNoodleMoose (approved):

library/sync/src/main/java/org/cru/godtools/sync/work/SyncToolSharesWorker.kt
frett: nope, just a policy adjustment that made sense while I was adding the other network related tasks.

`REPLACE` will discard the existing work, `KEEP` will keep the existing work. The actual work being run would be the same between the old and new, the difference is in current backoff timers. I wanted to preserve the existing backoff timer if the code decides to re-schedule an existing work

app/src/main/java/org/cru/godtools/GodToolsApplication.kt
frett: the class name isn't the most helpful in this case. this `TimberLogger` is a WorkManager Logger that sends log messages to Timber. This is creating the logger, then installing it in WorkManager.

There isn't an official API to do this yet, which is the reason it's not in the work manager configuration.

TheNoodleMoose: Ah okay, that makes sense. Thanks!

PR #2119, opened 2022-01-31
GT-1259   prep tools page for spotlight and categories 

gyasistory:
https://jira.cru.org/browse/GT-1259

This is an initial PR to set up the Tool Fragment that will replace the old one with the new design.  This is just placeholder fragments and views to understand the way it will be created.  Upon approval of this  PR, I will delete the current ToolsFragment and replace it with the ToolsContainerFragment.


app/src/main/java/org/cru/godtools/ui/tools/ToolsCategoryFragment.kt
```diff
@@ -0,0 +1,9 @@
+package org.cru.godtools.ui.tools
```
frett, line 1: put all of this in the `org.cru.godtools.ui.dashboard.tools` package

app/src/main/java/org/cru/godtools/ui/tools/ToolsContainerFragment.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.cru.godtools.fragment.BasePlatformFragment
+
+@AndroidEntryPoint
+class ToolsContainerFragment : BasePlatformFragment<ToolsContainerFragmentBinding>(R.layout.tools_container_fragment) {
+
+    private val dataModel: ToolsContainerFragmentDataModel by viewModels()
+    override fun onBindingCreated(binding: ToolsContainerFragmentBinding, savedInstanceState: Bundle?) {
+        super.onBindingCreated(binding, savedInstanceState)
+        binding.refresh.setupSwipeRefresh()
+        binding.hasSpotLight = dataModel.hasSpotlight
```
frett, line 17: try and keep case on names consistent. because `spotlight` is a single word, you shouldn't have a capital `L` in the middle of the name.

frett:
You won't be able to delete the existing `ToolsFragment`, it's also used by the list of lessons and the list of favorites currently.

app/src/main/java/org/cru/godtools/ui/tools/ToolsCategoryFragment.kt
gyasistory: @frett did you mean `org.cru.godtools.base.ui.dashboard`.

frett: no, as a sub package of this: https://github.com/CruGlobal/godtools-android/tree/develop/app/src/main/java/org/cru/godtools/ui/dashboard

frett: so, keep it in the `app` module, just move the java package under the `ui/dashboard` package that already exists

gyasistory: OK I see thanks for the clarification

app/src/main/java/org/cru/godtools/ui/tools/ToolsContainerFragment.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import android.os.Bundle
+import androidx.fragment.app.viewModels
+import dagger.hilt.android.AndroidEntryPoint
+import org.cru.godtools.R
+import org.cru.godtools.databinding.ToolsContainerFragmentBinding
+import org.cru.godtools.fragment.BasePlatformFragment
+
+@AndroidEntryPoint
+class ToolsContainerFragment : BasePlatformFragment<ToolsContainerFragmentBinding>(R.layout.tools_container_fragment) {
```
frett, line 11: let's name this either: `ToolsFragment` or `DashboardToolsFragment`.

I'm leaning more towards the first one, but I know it will have some temporary import conflicts as you try to use it within `MainActivity`. I think we will probably want to rename the pre-existing `ToolsFragment` to be `ToolsListFragment` (don't do that in this PR, we will address that later).

app/src/main/res/layout/tools_container_fragment.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
```
frett, line 2: let's name this layout `dashboard_tools_fragment.xml`

app/src/main/res/layout/tools_spotlight_fragment.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:app="http://schemas.android.com/apk/res-auto"
```
frett, line 2: let's name this layout `dashboard_tools_spotlight_fragment.xml`

app/src/main/res/layout/tools_container_fragment.xml
frett: reason being, this is being positioned as a sub-component of the dashboard. so I'd like the layout names to reflect that hierarchy.

app/src/main/res/layout/tools_container_fragment.xml
```diff
@@ -0,0 +1,44 @@
...
+                android:layout_height="wrap_content"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toTopOf="parent"
+                app:visibleIf="@{hasSpotlight}"
+                tools:layout="@layout/tools_spotlight_fragment" />
+
+            <androidx.fragment.app.FragmentContainerView
+                android:id="@+id/fragmentCategory"
+                android:name="org.cru.godtools.ui.dashboard.tools.ToolsCategoryFragment"
```
frett, line 34: I know this is just to start setting up the structure, but you will probably need to change how you layout this UI in order for it to correctly scroll.

You will probably need a `RecyclerView` as the the root view under the `SwipeRefreshLayout`, then have the spotlight fragment as the first item in the RecyclerView, the category selector as the second item, and then all the available tools after that.

There really needs to be a better way in Android to setup a scrollable stack of content where some of those items are dynamic lists.

frett (approved):

PR #2122, opened 2022-02-01
GT-1259   create and implement tools category

gyasistory:
https://jira.cru.org/browse/GT-1259


app/src/main/java/org/cru/godtools/ui/dashboard/tools/CategoryAdapter.kt
```diff
@@ -0,0 +1,46 @@
+package org.cru.godtools.ui.dashboard.tools
+
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.lifecycle.Observer
+import androidx.recyclerview.widget.RecyclerView
+import org.cru.godtools.databinding.DashboardListItemCategoryBinding
+
+class CategoryAdapter(val callbacks: CategoryAdapterCallbacks) : RecyclerView.Adapter<CategoryAdapter.ViewHolder>(), Observer<List<String>> {
```
frett, line 9: you should extend `SimpleDataBindingAdapter`

app/src/main/res/layout/dashboard_list_item_category.xml
```diff
@@ -0,0 +1,50 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content">
+
+        <androidx.cardview.widget.CardView
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:minHeight="64dp"
+            android:layout_margin="4dp"
+            android:padding="@dimen/tools_card_corner_radius"
+            android:shadowColor="@color/tintable"
```
frett, line 23: this is the wrong color constant to use here, you aren't planning on tinting the shadow are you?

app/src/main/res/layout/dashboard_list_item_category.xml
```diff
@@ -0,0 +1,50 @@
...
+    <FrameLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content">
+
+        <androidx.cardview.widget.CardView
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:minHeight="64dp"
+            android:layout_margin="4dp"
+            android:padding="@dimen/tools_card_corner_radius"
```
frett, line 22: this isn't related to the tools renderer, so you should be using tools renderer dimensions

app/src/main/res/layout/dashboard_list_item_category.xml
```diff
@@ -0,0 +1,50 @@
...
+            android:shadowColor="@color/tintable"
+            tools:cardBackgrounColor="@color/gray_80"
+            app:cardBackgroundColor="@{isSelected ? @color/gray_80 : @color/white }"
+            app:cardCornerRadius="@dimen/card_corner_radius"
+            android:onClick="@{() -> callbacks.onCategorySelected(category)}"
+            app:cardElevation="4dp">
+
+            <androidx.constraintlayout.widget.ConstraintLayout
+                android:layout_width="match_parent"
+                android:layout_height="match_parent">
```
frett, line 32: is there a reason you are using `ConstraintLayout` here and not just putting the `TextView` directly in the `CardView`?

app/src/main/res/layout/dashboard_tools_fragment.xml
```diff
@@ -33,8 +32,10 @@
                 android:name="org.cru.godtools.ui.dashboard.tools.ToolsCategoryFragment"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
+                tools:layout_height="match_parent"
+                android:tag="ToolsCategory"
```
frett, line 36: what's your reason for setting a view tag?

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryAdapter.kt
```diff
@@ -0,0 +1,57 @@
...
+    override fun onBindViewHolder(holder: ViewHolder, position: Int) {
+        val tool = tools[position]
+        holder.binding.tool = tool
+
+        tool.code?.let {
+            holder.binding.parallelLanguage = dataModel.getToolViewModel(it).parallelLanguage
+            holder.binding.parallelTranslation = dataModel.getToolViewModel(it).parallelTranslation
+            holder.binding.primaryTranslation = dataModel.getToolViewModel(it).firstTranslation
+            holder.binding.downloadProgress = dataModel.getToolViewModel(it).downloadProgress
+            holder.binding.banner = dataModel.getToolViewModel(it).banner
```
gyasistory, line 54: @frett these values are returning null.  Is there some type of sync that I need to call first?

frett: you aren't setting a `LifecycleOwner` to data binding, so LiveData objects won't work in it

gyasistory: I knew it was something simple

app/src/main/res/layout/dashboard_tools_fragment.xml
gyasistory: @frett I thought it was best practice

app/src/main/java/org/cru/godtools/ui/tools/ToolsAdapter.kt
```diff
@@ -61,7 +65,13 @@ class ToolsAdapter(lifecycleOwner: LifecycleOwner, viewModelProvider: ViewModelP
 
     override fun onCreateViewDataBinding(parent: ViewGroup, viewType: Int) = when (viewType) {
         VIEW_TYPE_LESSON -> ToolsListItemLessonBinding.inflate(LayoutInflater.from(parent.context), parent, false)
-        else -> ToolsListItemToolBinding.inflate(LayoutInflater.from(parent.context), parent, false)
+        else -> {
+            if(isAllTools){
+                DashboardListItemToolsBinding.inflate(LayoutInflater.from(parent.context), parent, false)
+            }else {
+                ToolsListItemToolBinding.inflate(LayoutInflater.from(parent.context), parent, false)
+            }
```
frett, line 70: you should add this as a new `viewType` in the `getItemViewType` method.

frett: the `viewType` affects how RecyclerView caches/reuses items. If in the future we make `isAllTools` dynamic, that would cause the `RecyclerView` cache to become corrupted and we would get a weird intermix of different layout types.

app/src/main/java/org/cru/godtools/ui/tools/ToolsAdapter.kt
```diff
@@ -57,18 +58,13 @@ class ToolsAdapter(
 
     override fun getItemViewType(position: Int) = when (getItem(position)?.type) {
         Tool.Type.LESSON -> VIEW_TYPE_LESSON
-        else -> VIEW_TYPE_TOOL
+        else -> if (isAllTools) { VIEW_TYPE_ALL_TOOL } else { VIEW_TYPE_TOOL }
```
frett, line 61: instead of packing this into an if statement in the when else condition, flatten it out into the when statement itself

app/src/main/res/layout/dashboard_list_item_tools.xml
```diff
@@ -0,0 +1,130 @@
...
+                    <androidx.appcompat.widget.AppCompatImageView
+                        android:layout_width="match_parent"
+                        android:layout_height="match_parent"
+                        android:padding="3dp"
+                        android:tint="@color/gt_blue"
+                        app:selected="@{tool.isAdded}"
+                        app:srcCompat="@drawable/ic_favorite" />
+                    <!--                 />-->
+
+                </androidx.cardview.widget.CardView>
```
frett, line 77: don't use a `CardView` for the favorites icon, just use an image with elevation.

you might be able to use a `ShapeableImageView` with a white background and the foreground being the current favorite icons.

app/src/main/res/layout/dashboard_list_item_tools.xml
```diff
@@ -0,0 +1,130 @@
...
+
+
+                <androidx.constraintlayout.widget.Guideline
+                    android:id="@+id/guideline2"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:elevation="4dp"
```
frett, line 84: why do you have elevation on a guideline?

app/src/main/res/layout/dashboard_list_item_tools.xml
```diff
@@ -0,0 +1,130 @@
...
+                    android:textStyle="bold"
+                    app:layout_constraintStart_toStartOf="parent"
+                    app:layout_constraintTop_toBottomOf="@id/guideline2"
+                    tools:text="Tool Name" />
+
+                <TextView
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:layout_marginStart="16dp"
+                    android:text="@{tool.category.replace('_', ' ')}"
```
frett, line 105: there is a `getToolCategory()` method that you should be using for the layout

app/src/main/res/layout/dashboard_list_item_tools.xml
frett: `ShapeableImageView` would involve some kotlin code. a different approach that should work instead is to create a simple xml shape drawable, and set this as the background image directly

app/src/main/java/org/cru/godtools/ui/dashboard/tools/CategoryAdapter.kt
```diff
@@ -0,0 +1,46 @@
...
+
+class CategoryAdapter() :
+    SimpleDataBindingAdapter<DashboardListItemCategoryBinding>(), Observer<List<String>> {
+
+    val callbacks = ObservableField<CategoryAdapterCallbacks>()
+    var selectedCategory: String? = null
+        set(value) {
+            field = value
+            notifyDataSetChanged()
+        }
```
frett, line 18: you should just provide the selected category LiveData directly instead of duplicating it's state to this variable and having to re-bind all categories when the selection changes

app/src/main/java/org/cru/godtools/ui/dashboard/tools/CategoryAdapter.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    override fun getItemCount() = categories.size
+
+    override fun onChanged(t: List<String>) {
+        categories = t
+    }
+
+    override fun onCreateViewDataBinding(parent: ViewGroup, viewType: Int) = DashboardListItemCategoryBinding
+        .inflate(LayoutInflater.from(parent.context)).also {
+            it.callbacks = callbacks.get()
```
frett, line 34: using `callbacks.get()` defeats the purpose of using an `ObservableField`

app/src/main/java/org/cru/godtools/ui/dashboard/tools/CategoryAdapter.kt
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+    override fun onBindViewDataBinding(binding: DashboardListItemCategoryBinding, position: Int) {
+        val category = categories[position]
+        binding.category = category
+        binding.isSelected = category == selectedCategory
```
frett, line 40: this comparison can be handled in databinding if you set the selected category LiveData to the binding

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    val categories = allTools.map {
+        it.mapNotNull { tool ->
+            tool.category
+        }.distinct()
+    }
+
+    val viewTools: LiveData<List<Tool>> = selectedCategory.combineWith(allTools) { selectedCategory, tools ->
+        if (selectedCategory == null) {
+            return@combineWith tools
```
frett, line 37: you don't need to use the `return` operator within a closure, the last statement that executes is automatically returned

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.keynote.godtools.android.db.GodToolsDao
+
+@HiltViewModel
+class ToolsCategoryDataModel @Inject constructor(dao: GodToolsDao) : ViewModel() {
+
+    val selectedCategory = MutableLiveData<String?>(null)
+
+    private val where =
+        Contract.ToolTable.FIELD_TYPE.`in`(*constants(Tool.Type.TRACT, Tool.Type.ARTICLE, Tool.Type.CYOA))
+            .and(Contract.ToolTable.FIELD_HIDDEN.ne(true))
```
frett, line 24: don't define this as an instance variable, that means every data model that is created will recreate this query, when DB queries are immutable.

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -1,10 +1,87 @@
...
+    private val categoryAdapter: CategoryAdapter by lazy {
+        CategoryAdapter().also { adapter ->
+            adapter.callbacks.set(this)
+            lifecycle.onDestroy { adapter.callbacks.set(null) }
+            dataModel.categories.observe(this, adapter)
+        }
+    }
+
+    private val toolsAdapter: ToolsAdapter by lazy {
+        ToolsAdapter(viewLifecycleOwner, toolsDataModel, true).also { adapter ->
```
frett, line 47: you shouldn't be referencing the `viewLifecycleOwner` within Kotlin lazy initialization.

app/src/main/java/org/cru/godtools/ui/tools/ToolsAdapter.kt
```diff
@@ -53,11 +58,15 @@ class ToolsAdapter(lifecycleOwner: LifecycleOwner, private val dataModel: ToolsA
 
     override fun getItemViewType(position: Int) = when (getItem(position)?.type) {
         Tool.Type.LESSON -> VIEW_TYPE_LESSON
-        else -> VIEW_TYPE_TOOL
+        else -> when (isAllTools) {
+            true -> VIEW_TYPE_ALL_TOOL
+            false -> VIEW_TYPE_TOOL
+        }
```
frett, line 64: this when should be collapsed to be part of the parent when

app/src/main/res/layout/dashboard_list_item_category.xml
```diff
@@ -0,0 +1,45 @@
...
+            app:cardElevation="4dp"
+            tools:cardBackgrounColor="@color/gray_80">
+
+            <TextView
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginStart="20dp"
+                android:layout_marginTop="15dp"
+                android:layout_marginBottom="15dp"
+                android:text="@{category.replace('_', ' ')}"
```
frett, line 35: you should use the getToolCategory method to resolve the display name taking localization into account

app/src/main/res/layout/dashboard_list_item_category.xml
```diff
@@ -0,0 +1,45 @@
...
+                app:layout_constraintBottom_toBottomOf="parent"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toTopOf="parent"
+                tools:text="Categroy" />
```
frett, line 40: misspelled category

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -1,10 +1,87 @@
...
+            lifecycle.onDestroy { adapter.callbacks.set(null) }
+            dataModel.viewTools.observe(this, adapter)
+        }
+    }
+
+    override fun onBindingCreated(binding: DashboardToolsCategoryFragmentBinding, savedInstanceState: Bundle?) {
+        super.onBindingCreated(binding, savedInstanceState)
+        binding.lifecycleOwner = viewLifecycleOwner
+        binding.categoryRecyclerView.adapter = categoryAdapter
+        binding.toolsRecyclerView.adapter = toolsAdapter
```
frett, line 58: these 2 statements introduce the possibility of leaking the corresponding `RecyclerView`. whenever an adapter is assigned to a `RecyclerView`, a reference to the `RecyclerView` is added to the adapter itself. Because your adapters are fixed `vals` on the fragment, they won't be destroyed when the views are destroyed.

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -0,0 +1,43 @@
...
+    val selectedCategory = MutableLiveData<String?>(null)
+
+    private val allTools = Query.select<Tool>().where(
+        Contract.ToolTable.FIELD_TYPE.`in`(*constants(Tool.Type.TRACT, Tool.Type.ARTICLE, Tool.Type.CYOA))
+            .and(Contract.ToolTable.FIELD_HIDDEN.ne(true))
+    ).getAsLiveData(dao)
+
+    val categories = allTools.map {
+        it.mapNotNull { tool ->
+            tool.category?.let { category -> Pair(category, tool.getCategory(getApplication(), Locale.getDefault())) }
```
frett, line 32: what are you trying to do here?

app/src/main/java/org/cru/godtools/ui/tools/ToolsAdapter.kt
gyasistory: @frett not really sure how I would flatten this into one when, since I am checking two conditions?

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
gyasistory: I am getting the category from the tool and the category that will be used in the view in one pair.

gyasistory: It returns a `LiveData<List<Pair<String, String>>>`

app/src/main/java/org/cru/godtools/ui/tools/ToolsAdapter.kt
frett:
```
when {
   getItem(position)?.type == Tool.Type.LESSON -> VIEW_TYPE_LESSON
   isAllTools -> VIEW_TYPE_ALL_TOOLS
   ...
}
```

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
frett: you shouldn't need the actual `tool` object to resolve the display name of the category

frett: `categories` should be a `LiveData<List<String>>`. the display name of the category should be handled within data binding using a `getToolCategory` method to convert the internal category name to a localized category name

frett: you will probably need to create a new variation on the `getToolCategory` method that doesn't require the `Tool` object and takes the category as a string

gyasistory: @frett disregard the comments above.

app/src/main/java/org/cru/godtools/ui/dashboard/tools/CategoryAdapter.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.databinding.ObservableField
+import androidx.lifecycle.Observer
+import org.ccci.gto.android.common.recyclerview.adapter.SimpleDataBindingAdapter
+import org.cru.godtools.databinding.DashboardListItemCategoryBinding
+
+class CategoryAdapter() :
+    SimpleDataBindingAdapter<DashboardListItemCategoryBinding>(), Observer<List<String>> {
+
+    val callbacks = ObservableField<CategoryAdapterCallbacks>()
+    var selectedCategory = ObservableField<String>()
```
frett, line 14: instead of messing around with observing the live data to populate a `selectedCategory` field, you should just set the LiveData object directly here to directly set on the layout

app/src/main/java/org/cru/godtools/ui/dashboard/tools/CategoryAdapter.kt
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+    override fun getItemCount() = categories.size
+
+    override fun onChanged(t: List<String>) {
+        categories = t
+    }
+
+    override fun onCreateViewDataBinding(parent: ViewGroup, viewType: Int) = DashboardListItemCategoryBinding
+        .inflate(LayoutInflater.from(parent.context)).also {
```
frett, line 29: you should pass in the parent `ViewGroup` when inflating this binding so it correctly inflates layout params on the root view

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val allTools = Query.select<Tool>().where(
+        Contract.ToolTable.FIELD_TYPE.`in`(*constants(Tool.Type.TRACT, Tool.Type.ARTICLE, Tool.Type.CYOA))
+            .and(Contract.ToolTable.FIELD_HIDDEN.ne(true))
+    ).getAsLiveData(dao)
+
+    val categories = allTools.map {
+        it.mapNotNull { tool ->
+            tool.category
+        }.distinct()
+    }
```
frett, line 32: make this a single line of code, it's pretty straightforward what it is doing and having it multiple lines doesn't really improve any readability

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    val viewTools: LiveData<List<Tool>> = selectedCategory.combineWith(allTools) { selectedCategory, tools ->
```
frett, line 34: personal preference I would write this as `allTools.combineWith(selectedCategory)`. imo it makes it more clear that you are manipulating `allTools` instead of manipulating `selectedCategory`

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -1,10 +1,94 @@
...
+    }
+
+    override fun onDestroyBinding(binding: DashboardToolsCategoryFragmentBinding) {
+        binding.categoryRecyclerView.adapter = null
+        binding.toolsRecyclerView.adapter = null
+    }
+    //endregion lifecycle
+
+    override fun onCategorySelected(category: String) {
+        val selectedCategory: String? = if (categoryAdapter.selectedCategory.get() == category) {
```
frett, line 70: you shouldn't be referencing the selected category out of the `categoryAdapter` here, that's what the dataModel is for

ui/base/src/main/kotlin/org/cru/godtools/base/ui/util/ModelUtils.kt
```diff
@@ -34,4 +34,5 @@ private fun Context.getToolCategoryStringRes(category: String) =
         0 -> null
         else -> resources.getString(id)
     }
+fun String.getLocalStringFromToolCategory(context: Context) = context.getToolCategoryStringRes(this)
```
frett, line 37: this doesn't account for backwards compatibility if a new category is added to the API. old versions of the app will then show `null` and not fallback to the new category value.

frett (dismissed):
see comments

frett:
https://user-images.githubusercontent.com/47818/153434263-0f5f7518-8122-4f48-b5f6-deed38386a3f.mp4

the categories should scroll off the top of the screen when the user is scrolling the list of tools

frett:
Screen Shot 2022-02-10 at 10 12 50 AM
font size and alignment of the parallel language label is incorrect

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -0,0 +1,38 @@
...
+    private val allTools = Query.select<Tool>().where(
+        Contract.ToolTable.FIELD_TYPE.`in`(*constants(Tool.Type.TRACT, Tool.Type.ARTICLE, Tool.Type.CYOA))
+            .and(Contract.ToolTable.FIELD_HIDDEN.ne(true))
+    ).orderBy(Contract.ToolTable.COLUMN_DEFAULT_ORDER).getAsLiveData(dao)
+
+    val categories = allTools.map {
+        it.mapNotNull { tool -> tool.category }.distinct()
+    }
+
+    fun getFormattedCategory(category: String) = category.getLocalStringFromToolCategory(getApplication())
```
frett, line 33: you don't need this method, you can call the ModelUtils method directly from within data binding

ui/base/src/main/kotlin/org/cru/godtools/base/ui/util/ModelUtils.kt
```diff
@@ -34,4 +34,6 @@ private fun Context.getToolCategoryStringRes(category: String) =
         0 -> null
         else -> resources.getString(id)
     }
+fun String.getLocalStringFromToolCategory(context: Context) =
+    context.getToolCategoryStringRes(this) ?: this.replace("_", " ")
```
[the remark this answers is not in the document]
frett: which shouldn't be a problem because the first thing that method does is just read the category out of the tool and then generates the name exclusively from that category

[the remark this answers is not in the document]
frett: it'll probably look something like this:
```kotlin
fun Tool?.getCategory(context: Context, locale: Locale? = null) = getToolCategoryName(this?.category, context, locale)

fun getToolCategoryName(category: String?, context: Context, locale: Locale? = null)
```

app/src/main/res/values/styles_tools.xml
```diff
@@ -20,6 +20,11 @@
         <item name="android:progressDrawable">@drawable/progress_download</item>
     </style>
 
+    <style name="ShapeAppearance.GodTools.Tools.ShapeableImageView.Shape.Circle" parent="">
```
frett, line 23: maybe something like: `ShapeAppearance.GodTools.Tools.Action.Favorite` where it describes what you are using the ShapeAppearance for and not just that it's a circle

app/src/main/java/org/cru/godtools/ui/dashboard/tools/CategoryAdapter.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    override fun onCreateViewDataBinding(parent: ViewGroup, viewType: Int) = DashboardListItemCategoryBinding
+        .inflate(LayoutInflater.from(parent.context), parent, false).also {
+            it.callbacks = callbacks
+        }
+
+    override fun onBindViewDataBinding(binding: DashboardListItemCategoryBinding, position: Int) {
+        val category = categories[position]
+        binding.category = category
+        binding.formattedCategory = dataModel.getFormattedCategory(category)
```
frett, line 36: you shouldn't need to set this on the binding, you should be able to directly call the ModelUtil method from the binding itself using the `category` name that you already have there.

The whole point is separation of concerns, when looking at the layout XML I want to be able to quickly tell how we are generating the category display name and not have to hunt where the name is being set to a variable

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    val viewTools: LiveData<List<Tool>> = allTools.combineWith(selectedCategory) { tools, selectedCategory ->
+        tools.filter { tool -> selectedCategory?.let { tool.category == it } ?: true }
```
frett, line 36: how about `tools.filter { selectedCategory == null || it.category == selectedCategory }`?

frett: that is much more straightforward to read and understand vs using `?.let {} ?: true`

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -1,10 +1,93 @@
...
+    }
+
+    override fun onDestroyBinding(binding: DashboardToolsCategoryFragmentBinding) {
+        binding.categoryRecyclerView.adapter = null
+        binding.toolsRecyclerView.adapter = null
+    }
+    //endregion lifecycle
+
+    override fun onCategorySelected(category: String) {
+        val selectedCategory: String? = if (dataModel.selectedCategory.value == category) {
```
frett, line 69: there is no point in creating the temporary `selectedCategory` variable, you can just directly assign it to `dataModel.selectedCategory.value`

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsFragment.kt
```diff
@@ -16,4 +17,10 @@ class ToolsFragment : BasePlatformFragment<DashboardToolsFragmentBinding>(R.layo
         binding.refresh.setupSwipeRefresh()
         binding.hasSpotlight = dataModel.hasSpotlight
     }
+
+    override fun onSyncData(helper: SwipeRefreshSyncHelper, force: Boolean) {
+        super.onSyncData(helper, force)
+        helper.sync(syncService.syncTools(force))
+        helper.sync(syncService.syncLanguages(force))
```
frett, line 24: why are you syncing languages from the `ToolsFragment`?

app/src/main/java/org/cru/godtools/ui/tools/ToolsAdapter.kt
```diff
@@ -23,9 +26,14 @@ private typealias VH = DataBindingDraggableItemViewHolder<ViewDataBinding>
 
 private const val VIEW_TYPE_TOOL = 1
 private const val VIEW_TYPE_LESSON = 2
-
-class ToolsAdapter(lifecycleOwner: LifecycleOwner, private val dataModel: ToolsAdapterViewModel) :
-    SimpleDataBindingDraggableItemAdapter<ViewDataBinding>(lifecycleOwner), Observer<List<Tool>> {
+private const val VIEW_TYPE_ALL_TOOL = 3
```
frett, line 29: how about `VIEW_TYPE_CUSTOM`?

app/src/main/java/org/cru/godtools/ui/tools/ToolsAdapter.kt
```diff
@@ -51,13 +59,15 @@ class ToolsAdapter(lifecycleOwner: LifecycleOwner, private val dataModel: ToolsA
         tools = t
     }
 
-    override fun getItemViewType(position: Int) = when (getItem(position)?.type) {
-        Tool.Type.LESSON -> VIEW_TYPE_LESSON
+    override fun getItemViewType(position: Int) = when {
+        getItem(position)?.type == Tool.Type.LESSON -> VIEW_TYPE_LESSON
+        itemLayout == R.layout.dashboard_list_item_tools -> VIEW_TYPE_ALL_TOOL
```
frett, line 64: reverse the order of these and make the first condition `itemLayout != null -> VIEW_TYPE_CUSTOM`

app/src/main/java/org/cru/godtools/ui/tools/ToolsAdapter.kt
```diff
@@ -51,13 +59,15 @@ class ToolsAdapter(lifecycleOwner: LifecycleOwner, private val dataModel: ToolsA
...
         else -> VIEW_TYPE_TOOL
     }
 
     override fun onCreateViewDataBinding(parent: ViewGroup, viewType: Int) = when (viewType) {
         VIEW_TYPE_LESSON -> ToolsListItemLessonBinding.inflate(LayoutInflater.from(parent.context), parent, false)
+        VIEW_TYPE_ALL_TOOL -> DashboardListItemToolsBinding.inflate(LayoutInflater.from(parent.context), parent, false)
```
frett, line 70: because we want to be able to support any custom layout, you shouldn't use this method to inflate custom layouts, use `DataBindingUtil` instead.

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
gyasistory:
```

            <TextView
                android:layout_marginStart="20dp"
                android:layout_marginTop="15dp"
                android:layout_marginBottom="15dp"
                android:text="@{ModelUtils.getLocalStringFromToolCategory(category, context)}"
                app:layout_constraintBottom_toBottomOf="parent"
                app:layout_constraintEnd_toEndOf="parent"
                app:layout_constraintStart_toStartOf="parent"
                tools:text="Category" />

```
@frett Whenever I add the code above to implement it in binding the app crashes.

frett: it's probably because category is null before you actually bind anything to the model

frett: and that method currently requires a non-null string

gyasistory: @frett the way you asked me to switch the code below actually worked in the binding so disregard the comment above.

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsFragment.kt
gyasistory: I was thinking it was needed to sync the parallel and primary languages in sync.

app/src/main/java/org/cru/godtools/ui/dashboard/tools/CategoryAdapter.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.databinding.ObservableField
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.Observer
+import org.ccci.gto.android.common.recyclerview.adapter.SimpleDataBindingAdapter
+import org.cru.godtools.databinding.DashboardListItemCategoryBinding
+
+class CategoryAdapter(lifecycleOwner: LifecycleOwner, private val dataModel: ToolsCategoryDataModel) :
```
frett, line 11: it looks like the only thing you are using out of the dataModel is the `selectedCategory`. let's just pass that LiveData into the constructor instead of the full data model

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -0,0 +1,35 @@
...
+    val selectedCategory = MutableLiveData<String?>(null)
+
+    private val allTools = Query.select<Tool>().where(
+        ToolTable.FIELD_TYPE.`in`(*constants(Tool.Type.TRACT, Tool.Type.ARTICLE, Tool.Type.CYOA))
+            .and(ToolTable.FIELD_HIDDEN.ne(true))
+    ).orderBy(ToolTable.COLUMN_DEFAULT_ORDER).getAsLiveData(dao)
+
+    val categories = allTools.map {
+        it.mapNotNull { tool -> tool.category }.distinct()
+    }
```
frett, line 30:
```suggestion
    val categories = allTools.map { it.mapNotNull { it.category }.distinct() }.distinctUntilChanged()
```

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -1,10 +1,93 @@
...
+            adapter.callbacks.set(this)
+            lifecycle.onDestroy { adapter.callbacks.set(null) }
+            dataModel.viewTools.observe(this, adapter)
+        }
+    }
+
+    // region lifecycle
+    override fun onBindingCreated(binding: DashboardToolsCategoryFragmentBinding, savedInstanceState: Bundle?) {
+        super.onBindingCreated(binding, savedInstanceState)
+        binding.lifecycleOwner = viewLifecycleOwner
```
frett, line 57: I actually think this is handled automatically by `BaseFragment`

frett: also, I don't think you are doing anything that would require a `LifecycleOwner` on the fragment binding itself currently

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -1,10 +1,93 @@
...
+    }
+    //endregion lifecycle
+
+    override fun onCategorySelected(category: String) {
+        val selectedCategory: String? = if (dataModel.selectedCategory.value == category) {
+            null
+        } else {
+            category
+        }
+        dataModel.selectedCategory.value = selectedCategory
```
frett, line 74:
```suggestion
        with(dataModel.selectedCategory) { value = if (value != category) category else null }
```

this works, I'm not sure if there is a more clear way to toggle the value

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -1,10 +1,93 @@
...
+
+    override fun removeTool(tool: Tool?, translation: Translation?) {
+        tool?.code?.let { downloadManager.unpinToolAsync(it) }
+    }
+
+    override fun onToolInfo(code: String?) {
+        code?.let { requireActivity().startToolDetailsActivity(code) }
+    }
+
+    override fun onToolsReordered(vararg ids: Long) {}
```
frett, line 90:
```suggestion
    override fun onToolsReordered(vararg ids: Long) = Unit
```

minor nit, I prefer doing the `= Unit` syntax for methods that are no-ops like this

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -1,10 +1,93 @@
...
+    override fun openTool(tool: Tool?, primary: Translation?, parallel: Translation?) {}
```
frett, line 91: same syntax here

app/src/main/java/org/cru/godtools/ui/tools/ToolsAdapter.kt
```diff
@@ -51,15 +52,15 @@ class ToolsAdapter(lifecycleOwner: LifecycleOwner, private val dataModel: ToolsA
...
+        else -> R.layout.tools_list_item_tool
     }
 
-    override fun onCreateViewDataBinding(parent: ViewGroup, viewType: Int) = when (viewType) {
-        VIEW_TYPE_LESSON -> ToolsListItemLessonBinding.inflate(LayoutInflater.from(parent.context), parent, false)
-        else -> ToolsListItemToolBinding.inflate(LayoutInflater.from(parent.context), parent, false)
-    }.also { it.setVariable(BR.callbacks, callbacks) }
+    override fun onCreateViewDataBinding(parent: ViewGroup, viewType: Int) =
+        DataBindingUtil.inflate<ViewDataBinding>(LayoutInflater.from(parent.context), viewType, parent, false)
+            .also { it.setVariable(BR.callbacks, callbacks) }
```
frett, line 63: @gyasistory take a look at how I cleaned up the `ToolsAdapter`

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -0,0 +1,29 @@
...
+
+@HiltViewModel
+class ToolsCategoryDataModel @Inject constructor(dao: GodToolsDao) : ViewModel() {
+    private val tools = Query.select<Tool>().where(
+        ToolTable.FIELD_TYPE.`in`(*constants(Tool.Type.TRACT, Tool.Type.ARTICLE, Tool.Type.CYOA))
+            .and(ToolTable.FIELD_HIDDEN.ne(true))
+    ).orderBy(ToolTable.COLUMN_DEFAULT_ORDER).getAsLiveData(dao)
+
+    val categories = tools.map { it.mapNotNull { it.category }.distinct() }
+    val selectedCategory = MutableLiveData<String?>(null)
```
frett, line 24: @gyasistory have you done anything with `SavedStateHandle` in a ViewModel yet? this is a simple place to utilize it to persist the currently selected category through app relaunches.

I can point you to a couple examples in our codebase if needed

gyasistory: @frett  no I have done anything with that before.

frett: k, let's jump on a quick call

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -13,17 +13,25 @@ import org.cru.godtools.model.Tool
...
     ).orderBy(ToolTable.COLUMN_DEFAULT_ORDER).getAsLiveData(dao)
 
-    val categories = tools.map { it.mapNotNull { it.category }.distinct() }
-    val selectedCategory = MutableLiveData<String?>(null)
+    val categories = tools.map { t -> t.mapNotNull { it.category }.distinct() }
 
-    val filteredTools = tools.combineWith(selectedCategory) { tools, category ->
-        tools.filter { category == null || it.category == category }
+    val filteredTools = tools.combineWith(selectedCategoryLiveData) { tools, category ->
+        tools.filter { category == null || it.category == category } ?: tools
```
frett, line 35: why the `?: tools`?

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -36,10 +36,13 @@ class ToolsCategoryFragment() :
...
 
     private val categoryAdapter: CategoryAdapter by lazy {
-        CategoryAdapter(this, dataModel.selectedCategory).also { adapter ->
+        CategoryAdapter(this, dataModel.selectedCategoryLiveData).also { adapter ->
             adapter.callbacks.set(this)
             lifecycle.onDestroy { adapter.callbacks.set(null) }
             dataModel.categories.observe(this, adapter)
+
+            // Sets live data to prevent blank tools list
+            dataModel.selectedCategoryLiveData.value = dataModel.selectedCategory
```
frett, line 45: you shouldn't need to do this

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -13,17 +13,25 @@ import org.cru.godtools.model.Tool
...
 
+private const val ATTR_SELECTED_CATEGORY = "selectedCategory"
+
 @HiltViewModel
-class ToolsCategoryDataModel @Inject constructor(dao: GodToolsDao) : ViewModel() {
+class ToolsCategoryDataModel @Inject constructor(dao: GodToolsDao, private val savedState: SavedStateHandle) :
+    ViewModel() {
+    var selectedCategory: String?
+        get() = savedState.get(ATTR_SELECTED_CATEGORY)
+        set(value) { savedState.set(ATTR_SELECTED_CATEGORY, value) }
```
frett, line 23: you probably don't need this property, you should be able to use the selectedCategoryLiveData directly

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -1,10 +1,87 @@
...
+import org.cru.godtools.model.Translation
+import org.cru.godtools.ui.tooldetails.startToolDetailsActivity
+import org.cru.godtools.ui.tools.ToolsAdapter
+import org.cru.godtools.ui.tools.ToolsAdapterCallbacks
+import org.cru.godtools.ui.tools.ToolsAdapterViewModel
 
 @AndroidEntryPoint
-class ToolsCategoryFragment :
-    BaseFragment<DashboardToolsCategoryFragmentBinding>(R.layout.dashboard_tools_category_fragment)
+class ToolsCategoryFragment() :
```
frett, line 21: I don't think you need the empty default constructor here. It got flagged by code scanning

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -1,10 +1,87 @@
...
+        binding.categoryRecyclerView.adapter = null
+        binding.toolsRecyclerView.adapter = null
+    }
+    //endregion lifecycle
+
+    override fun onCategorySelected(category: String) {
+        with(dataModel.selectedCategory) { value = if (value != category) category else null }
+    }
+
+    // region ToolsCategoryAdapterCallbacks
```
frett, line 71: minor nit: the callbacks class this region references doesn't exist

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryFragment.kt
```diff
@@ -1,10 +1,87 @@
...
-    BaseFragment<DashboardToolsCategoryFragmentBinding>(R.layout.dashboard_tools_category_fragment)
+    BaseFragment<DashboardToolsCategoryFragmentBinding>(R.layout.dashboard_tools_category_fragment),
+    CategoryAdapterCallbacks,
+    ToolsAdapterCallbacks {
+
+    @Inject
+    internal lateinit var downloadManager: GodToolsDownloadManager
+
+    @Inject
+    internal lateinit var settings: Settings
```
frett, line 30: it doesn't look like `settings` is used in this fragment

frett:
Screen Shot 2022-02-14 at 4 29 33 PM

The tools recyclerview is not efficiently laying out items. It is creating views for all the items in the list, even ones not visible. This will not scale well at all as we continue to add more tools.

Going back to my original comment before you started working on the tools categories UI, you will probably need to put the entire UI in a single RecyclerView to get the scrolling behavior we want along with efficient recycling of items without needing to layout all of our content at once

frett:
Screen Shot 2022-02-14 at 4 34 56 PM
The selected category UI is bad, it should have a blue border and the font color should change of not selected categories

frett:
Screen Shot 2022-02-14 at 4 40 01 PM
The parallel language label is still not being constrained correctly, it should be positioned off of the title of the tool. that way when the user has different accessibility settings selected it aligns properly

app/src/main/res/layout/dashboard_list_item_tools.xml
```diff
@@ -0,0 +1,125 @@
...
+                    app:selected="@{tool.isAdded}"
+                    app:shapeAppearance="@style/ShapeAppearance.GodTools.Tools.Action.Favorite"
+                    app:srcCompat="@drawable/ic_favorite" />
+
+                <androidx.constraintlayout.widget.Guideline
+                    android:id="@+id/guideline2"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:orientation="horizontal"
+                    app:layout_constraintGuide_end="62dp" />
```
frett, line 79: do you need this guideline? why not just set a fixed height for the banner image and base other measurements off the bottom edge of the banner image?

frett:
Screen Shot 2022-02-14 at 4 52 53 PM
You aren't preventing the tool name from overlapping the parallel language

frett:
Screen Shot 2022-02-14 at 4 54 26 PM

Screen Shot 2022-02-14 at 4 54 59 PM

You are missing the "Categories" label and the horizontal divider between the list of categories and the tools themselves

app/src/main/res/layout/dashboard_list_item_tools.xml
frett, line 79: Where did you get the height of `62dp` from?

app/src/main/res/layout/dashboard_list_item_tools.xml
```diff
@@ -0,0 +1,125 @@
...
+        <variable name="downloadProgress" type="LiveData&lt;org.cru.godtools.download.manager.DownloadProgress&gt;" />
+        <variable name="primaryTranslation" type="LiveData&lt;org.cru.godtools.model.Translation&gt;" />
+        <variable name="parallelTranslation" type="LiveData&lt;org.cru.godtools.model.Translation&gt;" />
+        <variable name="parallelLanguage" type="LiveData&lt;org.cru.godtools.model.Language&gt;" />
+    </data>
+
+    <FrameLayout
+        android:layout_width="match_parent"
+        android:layout_height="160dp"
+        android:padding="5dp">
```
frett, line 22: my guess is you really don't need this containing `FrameLayout`

frett: this padding is probably part of what is adding unnecessary whitespace to the UI

app/src/main/res/layout/dashboard_list_item_category.xml
```diff
@@ -0,0 +1,49 @@
...
+            app:cardElevation="4dp"
+            tools:cardBackgrounColor="@color/gray_80">
+
+            <TextView
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginStart="20dp"
+                android:layout_marginTop="15dp"
+                android:layout_marginBottom="15dp"
+                android:text="@{ModelUtils.getToolCategoryName(category, context, Locale.getDefault())}"
```
frett, line 39: We will need to get clarification, but my guess is that you will want to use the primaryLanguage as the locale for the category name

app/src/main/res/layout/dashboard_list_item_category.xml
```diff
@@ -0,0 +1,49 @@
...
+        <import type="java.util.Locale" />
+
+        <variable name="category" type="String" />
+        <variable name="selectedCategory" type="LiveData&lt;String>" />
+        <variable name="callbacks" type="ObservableField&lt;org.cru.godtools.ui.dashboard.tools.CategoryAdapterCallbacks>"/>
+    </data>
+
+    <FrameLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content">
```
frett, line 19: you shouldn't need this wrapping `FrameLayout`

app/src/main/res/layout/dashboard_list_item_tools.xml
gyasistory: image

frett: ah, I would have done it the opposite way, set a fixed height for the banner, then let the white area size automatically based on the content within it. I wouldn't have set a fixed overall height of each item, I would have done `wrap_content`. that would allow the size to scale up based on the user's accessibility settings.

frett: the cards aren't even the right fixed height currently, you have the `FrameLayout` set to `160dp`, with padding of `5dp`. then on the `CardView you have a margin of `8dp`. this yields the following height:
```
160dp
- 5dp x2
- 8dp x2
--------
134dp
```

134dp is less than the 150dp in figma. part of the redesign focus was to try and give a larger banner image on each tool.

app/src/main/res/layout/dashboard_list_item_category.xml
frett: and if we aren't using `primaryLanguage`, then you will want to add a `@JvmOverloads` annotation to the method and not provide a specific locale to the method.

app/src/main/res/layout/dashboard_list_item_tools.xml
gyasistory: yeah, I fixed it.

app/src/main/res/layout/dashboard_list_item_category.xml
```diff
@@ -14,24 +14,34 @@
...
+
+            <androidx.appcompat.widget.AppCompatImageView
+                android:layout_width="0dp"
+                android:layout_height="0dp"
+                app:layout_constraintBottom_toBottomOf="parent"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toTopOf="parent"
+                app:selected="@{category.equals(selectedCategory)}"
+                app:srcCompat="@drawable/bck_list_item_category_selector" />
```
frett, line 40: take a look at using `MaterialCardView` instead, it supposedly supports border colors for cards

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -28,4 +37,6 @@ class ToolsCategoryDataModel @Inject constructor(dao: GodToolsDao, savedState: S
     val filteredTools = tools.combineWith(selectedCategory) { tools, category ->
         tools.filter { category == null || it.category == category }
     }
+    private val primaryTranslation = settings.primaryLanguageLiveData
+    val primaryLanguage = primaryTranslation.switchMap { dao.findLiveData<Language>(it.isO3Language) }.orEmpty()
```
frett, line 41: why look up a language just to extract the Locale object back out of it?

gyasistory:
> Screen Shot 2022-02-14 at 4 29 33 PM
> 
> The tools recyclerview is not efficiently laying out items. It is creating views for all the items in the list, even ones not visible. This will not scale well at all as we continue to add more tools.
> 
> Going back to my original comment before you started working on the tools categories UI, you will probably need to put the entire UI in a single RecyclerView to get the scrolling behavior we want along with efficient recycling of items without needing to layout all of our content at once

@frett If you don't mind I would like to address this in the Spotlight PR since fixing this would require fixing that PR as well.

frett:
> @frett If you don't mind I would like to address this in the Spotlight PR since fixing this would require fixing that PR as well.

that's fine

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsCategoryDataModel.kt
```diff
@@ -37,6 +34,5 @@ class ToolsCategoryDataModel @Inject constructor(
     val filteredTools = tools.combineWith(selectedCategory) { tools, category ->
         tools.filter { category == null || it.category == category }
     }
-    private val primaryTranslation = settings.primaryLanguageLiveData
-    val primaryLanguage = primaryTranslation.switchMap { dao.findLiveData<Language>(it.isO3Language) }.orEmpty()
+    val primaryLanguage = settings.primaryLanguageLiveData.orEmpty()
```
frett, line 37: I don't think you need the `orEmpty()` here

app/src/main/res/layout/dashboard_list_item_category.xml
```diff
@@ -0,0 +1,42 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_margin="4dp"
+        android:onClick="@{() -> callbacks.onCategorySelected(category)}"
+        android:padding="6dp"
+        app:cardBackgroundColor="@color/white"
+        app:cardCornerRadius="@dimen/card_corner_radius"
+        app:cardElevation="4dp"
+        app:rippleColor="@color/gt_blue"
+        app:strokeColor="@{category.equals(selectedCategory) ? @color/gt_blue : @color/white}"
```
frett, line 28: why not `category == selectedCategory`?

frett:
Screen Shot 2022-02-16 at 9 33 48 AM
category margins aren't consistent

frett:
Screen Shot 2022-02-16 at 9 37 02 AM
There is no margin under the category name on the tool cards

frett:
Screen Shot 2022-02-16 at 9 38 54 AM
The tool name is not using the translated name

frett:
Screen Shot 2022-02-16 at 9 39 34 AM
another illustration of the category margins being incorrect, this time when using an RTL language

app/src/main/res/layout/dashboard_tools_category_fragment.xml
```diff
@@ -1,22 +1,50 @@
...
+            app:layoutManager="androidx.recyclerview.widget.GridLayoutManager"
+            app:spanCount="3"
+            tools:itemCount="5"
+            tools:listitem="@layout/dashboard_list_item_category" />
+
+        <View
+            android:layout_width="match_parent"
+            android:layout_height="1dp"
+            android:layout_margin="8dp"
+            android:background="@color/gray_E6" />
```
frett, line 38: use a `MaterialDivider` here with height of `wrap_content` and no background color

app/src/main/res/layout/dashboard_list_item_category.xml
```diff
@@ -0,0 +1,42 @@
...
+        android:layout_height="wrap_content"
+        android:layout_margin="4dp"
+        android:onClick="@{() -> callbacks.onCategorySelected(category)}"
+        android:padding="6dp"
+        app:cardBackgroundColor="@color/white"
+        app:cardCornerRadius="@dimen/card_corner_radius"
+        app:cardElevation="4dp"
+        app:rippleColor="@color/gt_blue"
+        app:strokeColor="@{category == selectedCategory ? @color/gt_blue : @color/white}"
+        app:strokeWidth="@{category == selectedCategory ? 10f : 0f}">
```
frett, line 29: the border should be `2dp` and not `10px`

ui/base/src/main/res/values/strings_base_ui.xml
```diff
@@ -5,6 +5,7 @@
 
     <!-- Tool Categories -->
     <eat-comment />
+    <string name="tool_category_label">Categories</string>
```
frett, line 8: this is a bad name for this string because if in the future we want to introduce a category named "label" this will cause a conflict.

Also, this string should be in `strings_tools.xml` within the `app` module

frett: the category names in this section are very specific because of the programmatic lookup that occurs based on values coming back from the API

app/src/main/res/layout/dashboard_list_item_category.xml
gyasistory: A little confused I thought I had to use a float.

app/src/main/res/layout/dashboard_list_item_category.xml
```diff
@@ -26,7 +26,7 @@
         app:cardElevation="4dp"
         app:rippleColor="@color/gt_blue"
         app:strokeColor="@{category == selectedCategory ? @color/gt_blue : @color/white}"
-        app:strokeWidth="@{category == selectedCategory ? 10f : 0f}">
+        app:strokeWidth="@{category == selectedCategory ? 2f : 0f}">
```
frett, line 29: this should be `2dp` and not `2px`, you will need to use a dimension value to actually resolve it within a binding expression.

app/src/main/res/layout/dashboard_list_item_category.xml
frett: no, you should be able to let the binding expression resolve the dimen constant for you and pass that value to the appropriate binding adapter.

frett: this should work, or something very similar to it

```
app:strokeWidth="@{category == selectedCategory ? @dimen/card_border_size : 0}"
```

frett (approved):

PR #2137, opened 2022-02-09
Gt 1441   create tool spotlight section

gyasistory:
depends on:
- [x] #2122 
- [x] #2146


app/src/main/res/layout/dashboard_tools_spotlight_list_item.xml
```diff
@@ -68,7 +68,7 @@
                     app:layout_constraintEnd_toEndOf="parent"
                     app:layout_constraintTop_toTopOf="parent"
                     app:selected="@{tool.isAdded}"
-                    app:shapeAppearance="@style/Widget.GodTools.Tools.ShapeableImageView.Shape"
+                    app:shapeAppearance="@style/Widget.GodTools.Tools.Card.Action.Favorite"
```
frett, line 71: this needs to be `style=` and not `app:shapeAppearance`

app/src/main/java/org/cru/godtools/ui/dashboard/tools/GridConcatAdapter.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.cru.godtools.databinding.GridConcatViewBinding
+
+class GridConcatAdapter(
+    lifecycleOwner: LifecycleOwner,
+    private val recyclerAdapter: RecyclerView.Adapter<*>,
+    private val rowCount: Int = 1,
+    private val titleText: String,
+    private val subTitleText: String? = null,
+    private val showDivider: Boolean = false
+) : SimpleDataBindingAdapter<GridConcatViewBinding>(lifecycleOwner) {
```
frett, line 18: I think you are trying to reuse too much here and making the layout much more dependent on code than it probably should be. 

you should have 2 distinct layouts, 1 for the spotlight section and 1 for the category selectors. All the configuration for those layouts should be able to be done purely in XML and you shouldn't need to pass in `rowCount`, `titleText`, `subTitleText`, or `showDivider`.

both of these layouts would have a RecyclerView that accepts an adapter, the spotlight one would be configured to use a horizontal `LinearLayoutManager` in the xml, and the categories one would use `GridLayoutManager` with 2 rows configured in the XML.

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsFragment.kt
```diff
@@ -2,24 +2,118 @@ package org.cru.godtools.ui.dashboard.tools
...
+    private val toolsSpotlightAdapter: ToolsAdapter by lazy {
+        ToolsAdapter(this, toolsDataModel, R.layout.dashboard_tools_spotlight_list_item).also { adapter ->
+            adapter.callbacks.set(this)
+            lifecycle.onDestroy { adapter.callbacks.set(null) }
+            toolsCategoryDataModel.spotlightTools.observe(this, adapter)
+        }
+    }
+
+    private val combinedToolsAdapter: ConcatAdapter by lazy {
+        ConcatAdapter(
```
frett, line 63: you should enable stable ids for the `ConcatAdapter`

app/src/main/res/layout/dashboard_categories_concat.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+    <data />
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
frett, line 7: since you are just stacking a set of views, why not use a LinearLayout?

app/src/main/res/layout/dashboard_categories_concat.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
```
frett, line 2: there is no reason to include `concat` in the layout name, you also aren't including the fact that this is part of the dashboard tools UI.

app/src/main/res/layout/dashboard_spotlight_concat.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools">
+
+    <data />
```
frett, line 6: if you aren't actually including any variables or imports, then there is no reason to include the `<data>` element.

app/src/main/res/layout/dashboard_categories_concat.xml
frett, line 2: You aren't using any actual databinding functionality in this layout, why not use ViewBinding instead of DataBinding for it?

frett: hmm, the `SimpleDataBindingAdapter` isn't setup to be able to leverage ViewBinding at the moment.

frett: hold off on getting rid of data binding for the time being, the base adapters probably need to be updated for it first.

app/src/main/java/org/cru/godtools/ui/dashboard/tools/GridConcatAdapter.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.annotation.LayoutRes
+import androidx.databinding.DataBindingUtil
+import androidx.databinding.ViewDataBinding
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.Observer
+import androidx.recyclerview.widget.RecyclerView
+import org.ccci.gto.android.common.recyclerview.adapter.SimpleDataBindingAdapter
+import org.cru.godtools.R
+
+class GridConcatAdapter(
```
frett, line 14: `Grid` doesn't make sense in this class name, adapter's have no concern of what LayoutManager is used to display their items.

frett: `Concat` also does not make sense in this adapter name, because there is no logic in the adapter about concating multiple things together.

app/src/main/java/org/cru/godtools/ui/dashboard/tools/GridConcatAdapter.kt
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    override fun onCreateViewDataBinding(parent: ViewGroup, viewType: Int): ViewDataBinding =
+        DataBindingUtil.inflate(LayoutInflater.from(parent.context), layoutRes, parent, false)
+
+    override fun onBindViewDataBinding(binding: ViewDataBinding, position: Int) {
+        binding.root.findViewById<RecyclerView>(R.id.concatRecyclerView).adapter = recyclerAdapter
+    }
+
+    override fun onChanged(t: Boolean?) {
```
frett, line 38: why is the `Boolean` nullable here? you specified the type in `Observer` to be non-null

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsFragment.kt
```diff
@@ -2,24 +2,118 @@ package org.cru.godtools.ui.dashboard.tools
...
+    @Inject
+    internal lateinit var downloadManager: GodToolsDownloadManager
 
-    private val dataModel: ToolsFragmentDataModel by viewModels()
-    override fun onBindingCreated(binding: DashboardToolsFragmentBinding, savedInstanceState: Bundle?) {
-        super.onBindingCreated(binding, savedInstanceState)
-        binding.refresh.setupSwipeRefresh()
-        binding.hasSpotlight = dataModel.hasSpotlight
+    // region Data Model
+    private val toolsCategoryDataModel: ToolsCategoryDataModel by viewModels()
```
frett, line 30:
```suggestion
    private val dataModel: ToolsFragmentDataModel by viewModels()
```

frett: the datamodel for this fragment should be renamed to actually match the fragment name

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsFragment.kt
```diff
@@ -2,24 +2,118 @@ package org.cru.godtools.ui.dashboard.tools
...
+            toolsCategoryDataModel.selectedCategory,
+            toolsCategoryDataModel.primaryLanguage
+        ).also { adapter ->
+            adapter.callbacks.set(this)
+            lifecycle.onDestroy { adapter.callbacks.set(null) }
+            toolsCategoryDataModel.categories.observe(this, adapter)
+        }
+    }
+
+    private val toolsCategoryAdapter: ToolsAdapter by lazy {
```
frett, line 46: why does this name have `Category` in the name? this should probably just be `toolsAdapter`

frett:
I have a hunch that this might leak views via the internal recycler adapters, I think you should move away from using lazily initialized adapters that are reused for the lifetime of the Fragment.

app/src/main/java/org/cru/godtools/ui/dashboard/tools/GridConcatAdapter.kt
```diff
@@ -0,0 +1,41 @@
+package org.cru.godtools.ui.dashboard.tools
```
frett, line 1: I don't think you are using this anywhere anymore

app/src/main/res/layout/dashboard_categories_concat.xml
frett: this doesn't need to be a databinding layout

app/src/main/res/layout/dashboard_spotlight_concat.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
```
frett, line 2: you aren't using any databinding functionality in this layout, so it doesn't need to be a databinding layout

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsFragment.kt
```diff
@@ -2,24 +2,116 @@ package org.cru.godtools.ui.dashboard.tools
...
+        }
+        concatAdapter.addLayout(R.layout.dashboard_tools_ui_categories) {
+            it.findViewById<RecyclerView>(R.id.concatRecyclerView).adapter = categoryAdapter
+        }.also { category ->
+            dataModel.hasCategories.observe(this) {
+                category.repeat = if (it) 1 else 0
+            }
+        }
+        concatAdapter.addAdapter(toolsAdapter)
     }
```
frett, line 78: this doesn't solve the potential memory leak issue, the leak would come from the toolsAdapter and toolsSpotlightAdapter. those would maintain a reference to the `RecyclerView` in the inflated views.

app/src/main/res/layout/dashboard_spotlight_concat.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.appcompat.widget.LinearLayoutCompat xmlns:android="http://schemas.android.com/apk/res/android"
```
frett, line 2: any reason you are using `LinearLayoutCompat` and not just `LinearLayout`? what does `LinearLayoutCompat` provide that isn't part of `LinearLayout`?

gyasistory: https://developer.android.com/reference/android/support/v7/widget/LinearLayoutCompat.html#pubmethods

It looks like it isn't supported any longer so I am going to switch.

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsFragment.kt
```diff
@@ -2,24 +2,101 @@ package org.cru.godtools.ui.dashboard.tools
...
+class ToolsFragment :
+    BasePlatformFragment<DashboardToolsFragmentBinding>(R.layout.dashboard_tools_fragment),
+    ToolCategoriesAdapter.Callbacks,
+    ToolsAdapterCallbacks {
+    @Inject
+    internal lateinit var downloadManager: GodToolsDownloadManager
 
-    private val dataModel: ToolsFragmentDataModel by viewModels()
+    // region Data Model
+    private val dataModel: ToolsCategoryDataModel by viewModels()
```
frett, line 31: the data model class should be named `ToolsFragmentDataModel`, not `ToolsCategoryDataModel`

app/src/main/res/layout/dashboard_spotlight_concat.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
frett, line 2: change the name of this layout to be `dashboard_tools_spotlight`, the `concat` is unnecessary, and just `dashboard_spotlight` doesn't indicate that it's part of the tools UI

app/src/main/res/layout/dashboard_tools_spotlight_list_item.xml
```diff
@@ -0,0 +1,105 @@
...
+                app:layout_constraintBottom_toBottomOf="@id/toolsImage"
+                tools:max="12"
+                tools:progress="5"
+                tools:visibility="visible" />
+
+            <com.google.android.material.imageview.ShapeableImageView
+                style="@style/Widget.GodTools.Tools.Card.Action.Favorite"
+                android:layout_marginTop="11dp"
+                android:layout_marginEnd="11dp"
+                android:elevation="@dimen/card_elevation"
```
frett, line 64: the elevation is centrally managed by the `Widget.GodTools.Tools.Card.Action.Favorite` style

app/src/main/res/layout/dashboard_tools_spotlight_list_item.xml
```diff
@@ -0,0 +1,105 @@
...
+                android:onClick="@{() -> tool.isAdded ? callbacks.removeTool(tool, primaryTranslation ?? parallelTranslation) : callbacks.addTool(tool.code)}"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintTop_toTopOf="parent"
+                app:selected="@{tool.isAdded}"
+                app:srcCompat="@drawable/ic_favorite" />
+
+            <TextView
```
frett, line 71: this should use the Tools Card Title widget style

app/src/main/java/org/cru/godtools/ui/dashboard/tools/ToolsFragment.kt
```diff
@@ -2,24 +2,101 @@ package org.cru.godtools.ui.dashboard.tools
...
 
     override fun onSyncData(helper: SwipeRefreshSyncHelper, force: Boolean) {
         super.onSyncData(helper, force)
         helper.sync(syncService.syncTools(force))
     }
+    //endregion Lifecycle
+
+    // region UI
+    private fun RecyclerView.setupRecyclerView() {
+        adapter = ConcatAdapter().apply {
```
frett, line 50: the `ConcatAdapter` should be configured to use stable ids

gyasistory: @frett every time I add `.setHasStableIds(true)` the app crashes with the error illustrated.

image

frett: you need to specify a config object when creating the `ConcatAdapter` to enable stable ids

app/src/main/res/layout/dashboard_tools_spotlight_list_item.xml
```diff
@@ -0,0 +1,105 @@
...
+            android:layout_height="match_parent"
+            android:layoutDirection="@{primaryTranslation.languageCode != null ? TextUtils.getLayoutDirectionFromLocale(primaryTranslation.languageCode) : View.LAYOUT_DIRECTION_INHERIT}">
+
+
+            <org.cru.godtools.base.ui.view.DaggerPicassoImageView
+                android:id="@+id/toolsImage"
+                style="@style/Widget.GodTools.Tools.Card.Banner"
+                android:layout_width="match_parent"
+                android:layout_height="161dp"
+                android:scaleType="centerCrop"
```
frett, line 42: I believe the `scaleType` attribute is already set in the Banner style

app/src/main/res/layout/dashboard_tools_spotlight_list_item.xml
```diff
@@ -0,0 +1,105 @@
...
+
+            <com.google.android.material.imageview.ShapeableImageView
+                style="@style/Widget.GodTools.Tools.Card.Action.Favorite"
+                android:layout_marginTop="11dp"
+                android:layout_marginEnd="11dp"
+                android:onClick="@{() -> tool.isAdded ? callbacks.removeTool(tool, primaryTranslation ?? parallelTranslation) : callbacks.addTool(tool.code)}"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintTop_toTopOf="parent"
+                app:selected="@{tool.isAdded}"
+                app:srcCompat="@drawable/ic_favorite" />
```
frett, line 68: I believe the favorite image is already set on the Action.Favorite style

app/src/main/res/layout/dashboard_tools_spotlight_list_item.xml
```diff
@@ -0,0 +1,105 @@
...
+                android:lines="2"
+                android:text="@{tool.name}"
+                app:layout_constraintBottom_toTopOf="@id/parallelLanguageText"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toBottomOf="@id/toolsImage"
+                tools:text="Tool Name" />
+
+            <androidx.appcompat.widget.AppCompatTextView
+                android:id="@+id/parallelLanguageText"
+                style="@style/Widget.GodTools.Tools.Card.Subtitle"
```
frett, line 86: actually looking at this again, there is a parallel language style that is the correct one to use here. my mistake in originally suggesting the subtitle one.

frett:
Screen Shot 2022-02-22 at 6 19 13 PM
both the spotlight and categories horizontal recyclerviews chop off their children before they reach the edge of the screen when they are scrolled

frett:
Screen Shot 2022-02-22 at 6 21 36 PM
The margins of the tool name in the tool spotlight cards isn't consistent

frett:
Screen Shot 2022-02-22 at 6 26 38 PM
The parallel language hides part of the tool name when accessibility settings are increased

frett:
the start margin of all the content isn't consistent, it looks like most of the views have different amounts of margin on the start edge

frett (changes requested):
You need to work on making sure the UI is clean and responsive

app/src/main/res/layout/dashboard_tools_spotlight_list_item.xml
```diff
@@ -0,0 +1,100 @@
...
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginStart="16dp"
+                android:layout_marginBottom="12dp"
+                android:text="@{parallelLanguage.getDisplayName(context)}"
+                app:layout_constraintTop_toBottomOf="@id/toolName"
+                app:layout_constraintBottom_toBottomOf="parent"
+                app:layout_constraintStart_toStartOf="parent"
+                app:visibleIf="@{tool.type.supportsParallelLanguage &amp;&amp; parallelLanguage != null &amp;&amp; !primaryTranslation.languageCode.equals(parallelLanguage.code)}"
+                tools:text="Spanish" />
```
frett, line 95: I'm checking now, but my guess is the visibility of this view is going to need to be a bit more complex than it currently is

frett: Screen Shot 2022-02-23 at 11 37 21 AM

frett: instead of using `visibleIf`, I think you will want to use a combination of `goneIf` and `invisibleIf`.

If the user doesn't have a parallel language selected, there is no reason to include the extra space for a parallel language in this card, if a parallel language is selected it should be visible for tools that have it, and the view will need to be invisible for tools that aren't available in the parallel language.

This is also a great place to write some unit tests around the visibility of the parallel language label to make sure it behaves properly

gyasistory:
@frett Ann asked me to add the following changes and I told her I would place them in this PR.  So please check for these in your review.

- Can we add 10 px above and below the horizontal line between the categories and the tools?
- How hard is it to add a 6th category called “All”? We understand that All is the default when nothing is selected but it might add clarity for the user.
- Lastly, Justin really likes the blue tap on the button - nice work! He was wondering what would be involved in fading the tools in and/or out as the button is selected.

frett:
> * Can we add 10 px above and below the horizontal line between the categories and the tools?

There is still some tightening up of the styles around that divider that does need to be done to make it look good. in the most recent screenshot I posted I actually think it has too much vertical padding now, but I was focusing on some of the bigger layout issues before nitpicking pixels here and there

> * How hard is it to add a 6th category called “All”? We understand that All is the default when nothing is selected but it might add clarity for the user.

I would actually do this as a followup PR since it should be pretty self-contained within the Categories logic itself.

> * Lastly, Justin really likes the blue tap on the button - nice work! He was wondering what would be involved in fading the tools in and/or out as the button is selected.

If we are correctly using stable ids for the tools list this should happen automatically for us and there isn't anything to do here.

frett:
other things in the current UI that may need some adjustments, but I haven't looked too in depth at yet:
- the Spotlight, Spotlight description, and Categories labels should have some styles defined for consistency and to match the styles in figma.
- padding/margin between items in the various adapters might need to be tightened up. (this would include the horizontal adapter, but could also include spacing between spotlight tools, categories, or tools)
- in spotlight cards margin alignment between the favorite action and the tool name view.

Not sure how many of these will actually need adjustments yet because I haven't dug in too deep on them.

frett:
for most dimensions I try to use `8dp` increments because that seems to be a fairly standard "grid" size in the android ecosystem, sometimes using `4dp` when necessary.

So, some places that have a dimension of 15 or a dimension of 20 within figma I'll generally normalize to `16dp` to follow closer to an `8dp` grid

app/src/main/res/layout/dashboard_tools_categories_category.xml
```diff
@@ -18,7 +18,9 @@
     <com.google.android.material.card.MaterialCardView
         android:layout_width="122sp"
         android:layout_height="wrap_content"
-        android:layout_margin="4dp"
+        android:layout_marginStart="@dimen/card_margin_horiz"
+        android:layout_marginTop="4dp"
+        android:layout_marginBottom="4dp"
```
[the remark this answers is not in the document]
frett: how you should handle extra padding at the start/end of a recyclerView is by using padding on the RecyclerView itself and disable `clipToPadding`.

for example: https://github.com/CruGlobal/godtools-android/blob/develop/ui/base-tool/src/main/res/layout/tool_share_sheet.xml#L49-L58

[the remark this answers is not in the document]
frett: Screen Shot 2022-02-23 at 2 07 05 PM
tool spotlight and categories both stop scrolling right at the edge of the screen, and should continue to scroll until they have a margin the same as the start of the RecyclerView

frett (dismissed):
still has UI changes necessary

frett:
Screen Shot 2022-02-23 at 2 12 23 PM
There is currently way too much vertical whitespace, I don't even see any tools after categories without scrolling. We really should be able to see at least part of the first tool so that the user realizes there is something down there.

app/src/main/res/layout/dashboard_tools_spotlight_list_item.xml
frett: @gyasistory have you addressed this feedback yet?

frett (changes requested):
you still haven't addressed the issue with the tool spotlight card layout

gyasistory:
> you still haven't addressed the issue with the tool spotlight card layout

@frett I think I fixed the issue.

app/src/main/res/layout/dashboard_tools_spotlight.xml
```diff
@@ -0,0 +1,36 @@
...
+    <androidx.recyclerview.widget.RecyclerView
+        android:id="@+id/tools"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:clipToPadding="false"
+        android:orientation="horizontal"
+        android:paddingStart="@dimen/dashboard_tools_spotlight_recycler_margin_horiz"
+        android:paddingEnd="@dimen/dashboard_tools_spotlight_recycler_margin_horiz"
+        app:layoutManager="androidx.recyclerview.widget.GridLayoutManager"
+        app:spanCount="1"
```
frett, line 34: can we use a `LinearLayoutManager` for this RecyclerView?

frett: I took care of this adjustment

frett (approved):

PR #2146, opened 2022-02-14
GT-1441 add spotlight to database

gyasistory:


library/db/src/main/kotlin/org/keynote/godtools/android/db/GodToolsDatabase.kt
```diff
@@ -90,6 +91,7 @@ class GodToolsDatabase @Inject internal constructor(@ApplicationContext private
                     }
                     46 -> db.execSQL(UserCounterTable.SQL_V46_CREATE_USER_COUNTERS)
                     47 -> db.execSQL(ToolTable.SQL_V47_ALTER_SCREEN_SHARE_DISABLED)
+                    48 -> db.execSQL(ToolTable.SQL_V48_POPULATE_SPOTLIGHT)
```
frett, line 94: this migration isn't currently being run, and will actually fail on upgrades

frett (dismissed):
see comments, the migration is currently invalid

library/db/src/main/kotlin/org/keynote/godtools/android/db/GodToolsDatabase.kt
frett, line 94: this migration will fail on upgrades

frett:
You will need to add your new field to the list of fields that are updated when an API sync happens, that is in `BaseDataSyncTasks`

frett (changes requested):
Migration will still fail

library/db/src/main/kotlin/org/keynote/godtools/android/db/GodToolsDatabase.kt
frett, line 94: This migration will still fail on upgrade

frett (dismissed):
You still haven't fixed the migration that will fail on upgrade.

try installing a version of the app from before you started on this PR, run it to initialize the database, then install the version from the PR.

library/db/src/main/kotlin/org/keynote/godtools/android/db/Contract.kt
```diff
@@ -179,6 +184,8 @@ object Contract : BaseContract() {
         internal const val SQL_V45_POPULATE_HIDDEN = "UPDATE $TABLE_NAME SET $COLUMN_HIDDEN = 0"
         internal const val SQL_V47_ALTER_SCREEN_SHARE_DISABLED =
             "ALTER TABLE $TABLE_NAME ADD COLUMN $SQL_COLUMN_SCREEN_SHARE_DISABLED"
+        internal const val SQL_V48_CREATE_SPOTLIGHT = "ALTER TABLE $TABLE_NAME ADD COLUMN $COLUMN_SPOTLIGHT"
```
frett, line 187: you should be using `SQL_COLUMN_SPOTLIGHT` here. you need to include the column type and not just the name of the column

frett (changes requested):
invalid SQL syntax for migration

PR #2148, opened 2022-02-15
GT-1446 Create a custom Tool Share Sheet

frett:
- wrap shareToolUri into LiveData to be able to observe changes
- get rid of unnecessary updateShareMenuItem method
- remove unnecessary message for share message subject
- create ShareBottomSheetDialogFragment
- trigger the new share sheet when there is more than 1 share item
- add Live Share item to custom share sheet


gyasistory (changes requested):
One of the test is failing

gyasistory (approved):

PR #2153, opened 2022-02-21
Fix obsolete tract pages calls

gyasistory:


frett (commented):
there aren't any changes in this PR anymore because it was all addressed already

PR #2169, opened 2022-03-01
GT-1437 Create Shareables bottom sheet for tools

frett:
- create initial shareables bottom sheet
- provide a method to get the latest translation as a Flow from the database
- create a getLatestPublishedManifestFlow to return the latest available manifest in a coroutines Flow
- enable a FileProvider for tools
- attempt to resolve the mime-type from a custom displayName
- create a bottom sheet for ShareableImages
- add shareable images to the custom share dialog


gyasistory (approved):
Everything looks good the test that is failing runs fine on my machine.

PR #2255, opened 2022-05-05
Refactor the ShortcutManager Dispatcher to no longer use EventBus

frett:
- monitor db invalidations instead of relying on EventBus events for updating pending events
- we no longer need to use an actor to trigger pending shortcut updates
- use a dao invalidation flow instead of eventbus to monitor for changes to attachments, tools, and translations
- we no longer need to use eventbus with the shortcut manager dispatcher
- convert shortcut manager dispatcher tests to MockK
- utilize runTest for all the dispatcher tests
- model the fact that the various flows emit initial values when collecting
- updateShortcutsJob no longer needs to be an actor

### DIFF
--- a/ui/shortcuts/src/main/kotlin/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
+++ b/ui/shortcuts/src/main/kotlin/org/cru/godtools/shortcuts/GodToolsShortcutManager.kt
@@ -321,13 +321,11 @@ class GodToolsShortcutManager @VisibleForTesting internal constructor(
         // endregion Events
 
         @VisibleForTesting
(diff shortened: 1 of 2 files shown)


gyasistory (changes requested):
@frett it looks like the GodToolsDownloadManagerTest is failing not sure if that needs to fixed in this PR

frett:
> @frett it looks like the GodToolsDownloadManagerTest is failing not sure if that needs to fixed in this PR

I'm not seeing any failures related to the download manager. The shortcut manager test was failing because it was using a stale version of the SNAPSHOT dependency

gyasistory (approved):

PR #2275, opened 2022-05-18
Enable Test Sharding

frett:
- enable test shard support via JUnit test categories
- add Category to robolectric tests
- move codecov.yml into .github directory
- enable test shards for GHA

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,3 +1,7 @@
+codecov:
+  notify:
+    after_n_builds: 2
+
 coverage:
   status:
     project:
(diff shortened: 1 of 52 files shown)


frett:
I'm putting this on hold because the bigger time issue within CI is building the actual app, this PR just adds a new job for the additional test shard that needs to be compiled.

A better performance improvement to GHA times would probably be setting up a remote gradle build cache. This would allow better caching of previous compilations.

frett:
I'm putting test sharding on hold at the time. This solution requires manually adding a `@Category` to every test in order for sharding to work. This isn't a maintainable solution and doesn't appear to make any performance improvements, so we will hold off on test sharding until we get a solution that works  and improves CI time

PR #2295, opened 2022-05-27
Convert About UI content to Jetpack Compose

frett:
- make it easy to enable jetpack compose for a module
- create a base GodToolsTheme for Compose
- create a new About screen layout using Compose
- use the new AboutLayout Composable instead of the AboutFragment


app/src/main/kotlin/org/cru/godtools/ui/about/AboutLayout.kt
```diff
@@ -0,0 +1,86 @@
...
+                            ?.let { uriHandler.openUri(it.item) }
+                    }
+                )
+            }
+        }
+    }
+}
+
+// TODO: extract this to gto-support when needing to reuse it for the first time
+@Composable
```
frett, line 68: I'm not happy with having to annotate this function as `@Composable`, but it was necessary to be able to access `MaterialTheme.colorScheme.primary`.

frett: whenever this gets moved to gto-support I might refactor out the linkStyle to get rid of the `@Composable` annotation. I'm sure at that point we will be more familiar with how `@Composable` works

app/src/main/kotlin/org/cru/godtools/ui/about/AboutLayout.kt
```diff
@@ -0,0 +1,86 @@
...
+import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.SpanStyle
+import androidx.compose.ui.text.buildAnnotatedString
+import androidx.compose.ui.text.style.TextDecoration
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import org.cru.godtools.R
+import org.cru.godtools.base.ui.compose.GodToolsTheme
+
+@Preview
```
gyasistory, line 25:
```suggestion
@Preview(showBackground = true)
```
This will allow the white background to be visible in the preview.

gradle/libs.versions.toml
```diff
@@ -1,5 +1,6 @@
 [versions]
 android-gradle-plugin = "7.2.1"
+androidx-compose = "1.2.0-beta02"
```
gyasistory, line 3: Why not use the release 1.1.1?

ui/base/src/main/kotlin/org/cru/godtools/base/ui/theme/GodToolsTheme.kt
```diff
@@ -0,0 +1,16 @@
+package org.cru.godtools.base.ui.compose
+
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.lightColorScheme
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.graphics.Color
+
+private val GT_BLUE = Color(0xFF3BA4DB)
```
gyasistory, line 8:
```suggestion
private val GT_BLUE = colorResource(R.color.gt_blue)
```
You could use this to maintain the resource values.

gyasistory (changes requested):
Looks good just a few comments and questions.

PR #2316, opened 2022-06-10
Putting compose in for the first page for the tutorial tips.

JasonHammar:
Compose had been added for the first page of the tutorial tips. The current version of compose that we are using doesn't like animations, so a placeholder image has been placed in there. The code for the lottie animation is commented out in the tips layout class.


frett:
@JasonHammar you missed adding the `TipsLayout()` composable

JasonHammar:
I put the TipsLayout file in.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout
```diff
@@ -0,0 +1,96 @@
+package org.cru.godtools.tutorial
```
frett, line 1: the filename should end with `.kt` to indicate that it's a Kotlin source file

JasonHammar: Just added it.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout
```diff
@@ -0,0 +1,96 @@
...
+        Spacer(modifier = Modifier.height(20.dp))
+        Text(text = "Learn to share this tool with anyone",
+            //style = MaterialTheme.typography.subtitle2,
+            fontSize = 37.sp,
+            textAlign = TextAlign.Center
+            //modifier = Modifier.align
+
+        )
+        Spacer(modifier = Modifier.height(20.dp))
+        Text("Practice using as many tips as you need to feel confident in sharing this tool with others.",
```
frett: in order for translations to actually work we need to reference the string resource that we already have defined. in this case you will want to use `stringResource(R.string.tutorial_tips_learn_text)`.

all of the strings for the tips tutorial should be defined in `strings_tutorial_tips.xml`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,96 @@
+package org.cru.godtools.tutorial
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.layout.*
```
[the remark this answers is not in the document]
frett: Android Studio by default uses wildcard imports even though it's considered a best practice to not use them. To avoid Android Studio constantly switching your imports to them you can modify the code style settings in preferences.

In the Android Studio preferences you can go here to change the setting to "Use single name import"

Screen Shot 2022-06-10 at 5 35 03 PM

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,96 @@
...
+
+@Preview(showBackground = true)
+@Composable
+internal fun TipsLayout() = GodToolsTheme() {
+    Column(
+        modifier = Modifier.fillMaxSize(),
+        //verticalArrangement = Arrangement.Center,
+        horizontalAlignment = Alignment.CenterHorizontally
+    ){
+        Spacer(modifier = Modifier.height(100.dp))
```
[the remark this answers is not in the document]
frett: you can ignore any magic number warnings for our projects, I actually probably need to go adjust the settings of this utility to only alert us about things we care about.

frett, line 31: so, you can actually implement the vertical spacer by using a `.padding(top = 100.dp)` modifier on the column.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,96 @@
...
+
+        )
+        Spacer(modifier = Modifier.height(20.dp))
+        Text(text = "Learn to share this tool with anyone",
+            //style = MaterialTheme.typography.subtitle2,
+            fontSize = 37.sp,
+            textAlign = TextAlign.Center
+            //modifier = Modifier.align
+
+        )
```
frett, line 88: We will want to use typography styles from the `MaterialTheme`. This page talks about the different styles available: https://m3.material.io/styles/typography/type-scale-tokens

I would guess we will want to use one of the title/headline styles along with the bodyLarge style.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,94 @@
...
+        horizontalAlignment = Alignment.CenterHorizontally
+    ){
+        Spacer(modifier = Modifier.height(100.dp))
+        val composition by rememberLottieComposition(
+            LottieCompositionSpec
+                .RawRes(org.cru.godtools.tutorial.R.raw.anim_tutorial_tips_people)
+        )
+        var isPlaying by remember {
+            mutableStateOf(true)
+        }
```
frett, line 43: we won't be toggling whether the animation is playing or not, it will always be playing.

so, there is no need to create this state variable, you can just directly pass in true for `isPlaying` in the `animateLottieCompositionAsState` method.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,94 @@
...
+
+// for speed
+        var speed by remember {
+            mutableStateOf(1f)
+        }
```
frett, line 48: same thing with `speed`, it won't be changing for the tutorial, so it can be hard-coded in `animateLottieCompositionAsState`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,94 @@
...
+            // when paused and play
+            // pass false to continue the animation
+            // at which is was paused
+            restartOnPlay = false
+
+        )
+        LottieAnimation(
+            composition,
+            progress,
+            modifier = Modifier.size(250.dp)
```
frett, line 75: for the size of the animation, it was previously 290dp high, and the width should be full width of the UI.

the `.size(250.dp)` modifier sets both width and height to the same dimension.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,94 @@
...
+        )
+
+        Spacer(modifier = Modifier.height(20.dp))
```
frett, line 78: I would move this to be a `.padding(top = 20.dp)` modifier on the `Text` element

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,94 @@
...
+        Text(text = "Learn to share this tool with anyone",
+            style = MaterialTheme.typography.titleMedium,
+            fontSize = 30.sp,
+            textAlign = TextAlign.Center
+            //modifier = Modifier.align
+
+        )
+        Spacer(modifier = Modifier.height(20.dp))
```
frett, line 86: this should also be moved to a padding modifier on the Text composable.

In general I want to try and keep any sort of element spacing defined directly on actual rendered content and not use a separate `Spacer` to create the space

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,94 @@
...
+        Text("Practice using as many tips as you need to feel confident in sharing this tool with others.",
+            style = MaterialTheme.typography.bodySmall,
+            fontSize = 20.sp,
```
frett, line 89: let's switch to the `bodyLarge` style here and drop the `fontSize`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.airbnb.lottie.compose.rememberLottieComposition
+//import org.cru.godtools.R
+import org.cru.godtools.base.ui.theme.GodToolsTheme
+import java.nio.file.Files.size
+
+@Preview(showBackground = true)
+@Composable
+internal fun TipsLayout() = GodToolsTheme() {
+    Column(
+        modifier = Modifier.fillMaxSize().padding(top = 100.dp),
```
frett, line 31: at some point we will need to work on vertically centering the content and not relying on a fixed top padding to push the content towards the center.

The drawback of using a fixed dimension to push content towards the center is that not all phones are the same dimensions, so this will cause poor centering on pretty much all devices other than the device you are testing with.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
frett: it actually looks like the default value for `isPlaying` is `true`, so you wouldn't need to pass in that argument to `animateLottieCompositionAsState` at all

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
frett: the default value for `speed` is `1f`, so you wouldn't need to pass this argument to `animateLottieCompositionAsState` at all

app/src/main/kotlin/org/cru/godtools/ui/about/AboutActivity.kt
```diff
@@ -22,7 +22,7 @@ class AboutActivity : BasePlatformActivity<ActivityGenericComposeWithNavDrawerBi
     // region Lifecycle
     override fun onBindingChanged() {
         super.onBindingChanged()
-        binding.genericActivity.frame.setContent { AboutLayout() }
+       binding.genericActivity.frame.setContent { AboutLayout() }
```
frett, line 25: I know you are still relatively new to git/github, just pointing this out to hopefully help you mature your git/github practices.

Random code churn like this can lead to several different issues:
1. merge conflicts if your branch accidentally updates a line that another PR also updates (intentionally or not)
2. more complicated git history/git blame
   * When looking at the git history for `AboutActivity` it would now include commits from this PR which is completed unrelated to the AboutActivity
3. Less focused PRs for reviews, instead of being able to focus on just what changed in the PR I now have a bunch of random noise I need to filter over to see if anything functionally changed with that updated code

frett: how I personally avoid this happening is on the command line I use the git staging area to build out my commit before committing it. I use `git add filename.ext` and `git add -p filename.ext` to add files to the staging area, `git reset filename.ext` to remove a file from the staging area, and `git commit` when I'm satisfied with committing the changes I have staged.

It looks like Github Desktop supports selecting individual changes and files when creating a commit as well, in the commit UI, you can use checkboxes to select/deselect files you want to include in a commit. And when previewing the changes in a file you can select/deselect individual lines that you want to include in the commit.

gradle/libs.versions.toml
```diff
@@ -18,7 +18,10 @@ gtoSupport = "3.11.2-SNAPSHOT"
 jacoco = "0.8.7"
 kotlin = "1.6.21"
 kotlinCoroutines = "1.6.2"
-okhttp3 = "4.10.0"
+
+lottie = "5.2.0"
+okhttp3 = "4.9.3"
+
```
frett, line 24: as you were addressing this merge conflict you ended up downgrading `okhttp3` to the previous version. Also the extra spaces around these versions should be removed.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,79 @@
...
+            progress,
+            modifier = Modifier.height(290.dp).fillMaxWidth()
+        )
+
+        //Spacer(modifier = Modifier.height(20.dp))
+        Text(text = stringResource(R.string.tutorial_tips_learn_headline),
+            style = MaterialTheme.typography.titleLarge,
+            //fontSize = 30.sp,
+            textAlign = TextAlign.Center,
+            modifier = Modifier.padding(top = 20.dp)
```
frett, line 67: based on the pre-existing layout, there shouldn't be any padding between the headline text and the animation

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,79 @@
...
+
+        )
+        //Spacer(modifier = Modifier.height(20.dp))
+        Text(stringResource(R.string.tutorial_tips_learn_text),
+            style = MaterialTheme.typography.bodyLarge,
+            //fontSize = 20.sp,
+            textAlign = TextAlign.Center,
+            modifier = Modifier.padding(top = 20.dp))
```
frett, line 75: The pre-existing layout only had 16dp of margin/padding between the title and the text

ui/tutorial-renderer/src/main/res/layout/tutorial_tips_learn.xml
```diff
@@ -19,43 +19,13 @@
...
-            android:breakStrategy="balanced"
-            android:gravity="center_horizontal"
-            android:text="@string/tutorial_tips_learn_text"
-            android:textAppearance="@style/TextAppearance.GodTools.Tutorial.Text"
-            app:layout_constraintEnd_toStartOf="@id/marginEnd"
-            app:layout_constraintStart_toEndOf="@id/marginStart"
-            app:layout_constraintTop_toBottomOf="@id/headlineBottom" />
+            app:layout_constraintStart_toStartOf="@id/marginStart"
+            app:layout_constraintEnd_toEndOf="@id/marginEnd"
+            app:layout_constraintTop_toTopOf="parent" />
```
frett, line 28: This view doesn't have a bottom constraint specified, which is probably part of the reason that the content is pushed down as far as it is. try adding the attribute `app:layout_constraintBottom_toTopOf="@id/action_next"`, that will constrain the bottom of the ComposeView to be aligned with the top of the continue button

frett: you will also want to change `android:layout_height` to be `0dp` so that it defers the height to be defined by the available space between the top and bottom constraints.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,79 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.airbnb.lottie.compose.*
+import org.cru.godtools.base.ui.theme.GodToolsTheme
+
+@Preview(showBackground = true)
+@Composable
+internal fun TipsLayout() = GodToolsTheme() {
+    Column(
+        modifier = Modifier.fillMaxSize().padding(top = 100.dp),
```
frett, line 22: you will want to get rid of the `padding()` modifier here, it's adding 100dp to the top which ends up shifting the vertical center 100dp lower.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,79 @@
+package org.cru.godtools.tutorial
```
frett, line 1: for organization, let's move this composable into the `org.cru.godtools.tutorial.layout` package. It might not live there long term, but it will hold it for now.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,79 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.airbnb.lottie.compose.*
+import org.cru.godtools.base.ui.theme.GodToolsTheme
+
+@Preview(showBackground = true)
+@Composable
+internal fun TipsLayout() = GodToolsTheme() {
```
frett, line 20: let's rename the composable to `TipsTutorialLayout()`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,79 @@
...
+        LottieAnimation(
+            composition,
+            progress,
+            modifier = Modifier.height(290.dp).fillMaxWidth()
+        )
+
+        //Spacer(modifier = Modifier.height(20.dp))
+        Text(text = stringResource(R.string.tutorial_tips_learn_headline),
+            style = MaterialTheme.typography.titleLarge,
+            //fontSize = 30.sp,
```
frett, line 65: one of our code style practices is to not include commented out code within our repo, except in rare one-off occurrences.

If it was code that was previously there that got commented out, it will always be available in the git history if we need to restore it.

ui/tutorial-renderer/build.gradle.kts
```diff
@@ -25,6 +26,7 @@ dependencies {
     implementation(libs.androidx.constraintlayout)
     implementation(libs.androidx.core.ktx)
     implementation(libs.androidx.viewpager2)
+    implementation(libs.androidx.compose.material3)
```
frett, line 29: just a personal preference, but I try to alpha sort dependencies within the different dependency sections. This allows quicker digestion when searching to see if a dependency has already been added or needs to be added.

The rough list of dependency sections are:
1. other GodTools sub projects
2. Any Jetpack (AndroidX) dependencies
3. Any Firebase dependencies
4. Any GTO Support dependencies (this is an internal support library with a bunch of commonly used code in our android apps)
5. Other 3rd party dependencies
6. KAPT dependencies (These dependencies transform/generate code for us)
7. Test dependencies
8. Test KAPT dependencies

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TipsLayout.kt
```diff
@@ -0,0 +1,84 @@
+package org.cru.godtools.tutorial.layout
```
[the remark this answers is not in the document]
frett: it should be moved to `ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsTutorialLayout.kt`

PR #2327, opened 2022-06-15
Adding compose to the last two tutorial tips pages.

JasonHammar:
Compose has been added to the last two pages for the tutorial tips.

### DIFF
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
@@ -57,11 +57,13 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
                     onTutorialAction = {
                         view?.let { it1 ->
-                            findListener<TutorialCallbacks>()?.onTutorialAction(it1)
+                            findListener<TutorialCallbacks>()?.onTutorialAction(it)
                         }
                     },
+                    Page.TIPS_LEARN,
                     R.raw.anim_tutorial_tips_people,
-                    R.string.tutorial_tips_learn_headline, R.string.tutorial_tips_learn_text
+                    R.string.tutorial_tips_learn_headline, R.string.tutorial_tips_learn_text,
+
                 )
             }
             is TutorialTipsLightBinding -> compose.setContent {
@@ -69,9 +71,10 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
                     onTutorialAction = {
                         view?.let { it1 ->
-                            findListener<TutorialCallbacks>()?.onTutorialAction(it1)
+                            findListener<TutorialCallbacks>()?.onTutorialAction(it)
                         }
                     },
+                    Page.TIPS_LIGHT,
                     R.raw.anim_tutorial_tips_tool,
                     R.string.tutorial_tips_light_headline, R.string.tutorial_tips_light_text1,
                     R.string.tutorial_tips_light_text2
@@ -82,9 +85,10 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
                     onTutorialAction = {
                         view?.let { it1 ->
-                            findListener<TutorialCallbacks>()?.onTutorialAction(R.id.action_tips_finish)
+                            findListener<TutorialCallbacks>()?.onTutorialAction(it)
                         }
                     },
+                    Page.TIPS_START,
                     R.raw.anim_tutorial_tips_light,
                     R.string.tutorial_tips_start_headline, R.string.tutorial_tips_start_text
                 )
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
@@ -55,11 +55,7 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
             is TutorialTipsLearnBinding -> compose.setContent {
                 TipsTutorialLayout(
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
-                    onTutorialAction = {
-                        id.let { it1 ->
-                            findListener<TutorialCallbacks>()?.onTutorialAction(it)
-                        }
-                    },
+                    onTutorialAction = { findListener<TutorialCallbacks>()?.onTutorialAction(it) },
                     Page.TIPS_LEARN,
                     R.raw.anim_tutorial_tips_people,
                     R.string.tutorial_tips_learn_headline, R.string.tutorial_tips_learn_text,
@@ -69,11 +65,7 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
             is TutorialTipsLightBinding -> compose.setContent {
                 TipsTutorialLayout(
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
-                    onTutorialAction = {
-                        id.let { it1 ->
-                            findListener<TutorialCallbacks>()?.onTutorialAction(it)
-                        }
-                    },
+                    onTutorialAction = { findListener<TutorialCallbacks>()?.onTutorialAction(it) },
                     Page.TIPS_LIGHT,
                     R.raw.anim_tutorial_tips_tool,
                     R.string.tutorial_tips_light_headline, R.string.tutorial_tips_light_text1,
@@ -83,11 +75,7 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
             is TutorialTipsStartBinding -> compose.setContent {
                 TipsTutorialLayout(
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
-                    onTutorialAction = {
-                        id.let { it1 ->
-                            findListener<TutorialCallbacks>()?.onTutorialAction(it)
-                        }
-                    },
+                    onTutorialAction = { findListener<TutorialCallbacks>()?.onTutorialAction(it) },
                     Page.TIPS_START,
                     R.raw.anim_tutorial_tips_light,
                     R.string.tutorial_tips_start_headline, R.string.tutorial_tips_start_text
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
@@ -56,7 +56,7 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                 TipsTutorialLayout(
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
                     onTutorialAction = { findListener<TutorialCallbacks>()?.onTutorialAction(it) },
-                    Page.TIPS_LEARN,
+                    page,
                     R.raw.anim_tutorial_tips_people,
                     R.string.tutorial_tips_learn_headline, R.string.tutorial_tips_learn_text,
 
@@ -66,7 +66,7 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                 TipsTutorialLayout(
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
                     onTutorialAction = { findListener<TutorialCallbacks>()?.onTutorialAction(it) },
-                    Page.TIPS_LIGHT,
+                    page,
                     R.raw.anim_tutorial_tips_tool,
                     R.string.tutorial_tips_light_headline, R.string.tutorial_tips_light_text1,
                     R.string.tutorial_tips_light_text2
@@ -76,7 +76,7 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                 TipsTutorialLayout(
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
                     onTutorialAction = { findListener<TutorialCallbacks>()?.onTutorialAction(it) },
-                    Page.TIPS_START,
+                    page,
                     R.raw.anim_tutorial_tips_light,
                     R.string.tutorial_tips_start_headline, R.string.tutorial_tips_start_text
                 )
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
@@ -56,7 +56,7 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                 TipsTutorialLayout(
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
                     onTutorialAction = {
-                        view?.let { it1 ->
+                        id.let { it1 ->
                             findListener<TutorialCallbacks>()?.onTutorialAction(it)
                         }
                     },
@@ -70,7 +70,7 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                 TipsTutorialLayout(
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
                     onTutorialAction = {
-                        view?.let { it1 ->
+                        id.let { it1 ->
                             findListener<TutorialCallbacks>()?.onTutorialAction(it)
                         }
                     },
@@ -84,7 +84,7 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                 TipsTutorialLayout(
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
                     onTutorialAction = {
-                        view?.let { it1 ->
+                        id.let { it1 ->
                             findListener<TutorialCallbacks>()?.onTutorialAction(it)
                         }
                     },
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
@@ -14,7 +14,6 @@ import org.cru.godtools.tutorial.databinding.TutorialTipsLearnBinding
 import org.cru.godtools.tutorial.databinding.TutorialTipsLightBinding
 import org.cru.godtools.tutorial.databinding.TutorialTipsStartBinding
 import org.cru.godtools.tutorial.layout.TipsTutorialLayout
-import org.cru.godtools.tutorial.TutorialCallbacks
 import splitties.fragmentargs.arg
 import splitties.fragmentargs.argOrNull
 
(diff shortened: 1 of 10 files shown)


ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
```diff
@@ -50,7 +52,25 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
 
     private fun ViewDataBinding.setupCompose() {
         when (this) {
-            is TutorialTipsLearnBinding -> compose.setContent { TipsTutorialLayout() }
+            is TutorialTipsLearnBinding -> compose.setContent {
+                TipsTutorialLayout(
+                    R.raw.anim_tutorial_tips_people,
+                    R.string.tutorial_tips_learn_headline, R.string.tutorial_tips_learn_text, 0
```
frett, line 58: because the last parameter is only used for one of the pages, we can actually add a default value for it and then you won't have to specify `0` for the screens that don't have it

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -14,71 +14,133 @@ import androidx.compose.ui.Alignment
...
-
-        )
-
-        Text(
-            stringResource(R.string.tutorial_tips_learn_text),
-            style = MaterialTheme.typography.bodyLarge,
-            textAlign = TextAlign.Center,
-            modifier = Modifier.padding(top = 16.dp)
-        )
+internal fun TipsTutorialLayout(anim: Int, title: Int, body: Int, body2: Int) = GodToolsTheme() {
```
frett, line 26: to make `body2` have a default value, use `body2: Int = 0`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -14,71 +14,133 @@ import androidx.compose.ui.Alignment
...
+    if (body2 == 0) {
```
frett, line 27: instead of wrapping the entire column in a conditional and having to duplicate the entire column contents, just wrap the single Text Composable in an if block

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -14,28 +14,27 @@ import androidx.compose.ui.Alignment
...
 import com.airbnb.lottie.compose.LottieConstants
 import com.airbnb.lottie.compose.animateLottieCompositionAsState
 import com.airbnb.lottie.compose.rememberLottieComposition
 import org.cru.godtools.base.ui.theme.GodToolsTheme
-import org.cru.godtools.tutorial.R
 
-@Preview(showBackground = true)
 @Composable
-internal fun TipsTutorialLayout() = GodToolsTheme() {
+internal fun TipsTutorialLayout(anim: Int, title: Int, body: Int, body2: Int = 0) = GodToolsTheme() {
```
frett, line 26: the one other thing I would change about the `body2` param is to make it nullable (and default to null) if you aren't displaying a body2 message.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -14,28 +14,27 @@ import androidx.compose.ui.Alignment
...
         verticalArrangement = Arrangement.Center,
         horizontalAlignment = Alignment.CenterHorizontally
     ) {
 
         val composition by rememberLottieComposition(
+
             LottieCompositionSpec
-                .RawRes(org.cru.godtools.tutorial.R.raw.anim_tutorial_tips_people)
+                .RawRes(anim)
         )
```
frett, line 38: now that `anim` is being passed in as a parameter, there isn't a lot of code here, so it should be able to be condensed to a single line

ui/tutorial-renderer/src/main/res/layout/tutorial_tips_start.xml
```diff
@@ -23,42 +23,7 @@
             app:layout_constraintEnd_toEndOf="@id/marginEnd"
             app:layout_constraintTop_toTopOf="parent" />
```
frett, line 24: This compose view doesn't currently have a bottom constraint specified, it probably needs to be specified.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -23,7 +23,7 @@ import com.airbnb.lottie.compose.rememberLottieComposition
 import org.cru.godtools.base.ui.theme.GodToolsTheme
 
 @Composable
-internal fun TipsTutorialLayout(anim: Int, title: Int, body: Int, body2: Int = 0) = GodToolsTheme() {
+internal fun TipsTutorialLayout(anim: Int, title: Int, body: Int, body2: Int?) = GodToolsTheme() {
```
frett, line 26: `body2: Int? = null` :)

ui/tutorial-renderer/src/main/res/layout/tutorial_tips_start.xml
```diff
@@ -21,7 +21,8 @@
             android:layout_height="wrap_content"
             app:layout_constraintStart_toStartOf="@id/marginStart"
             app:layout_constraintEnd_toEndOf="@id/marginEnd"
-            app:layout_constraintTop_toTopOf="parent" />
+            app:layout_constraintTop_toTopOf="parent"
+            app:layout_constraintBottom_toTopOf="@id/action_next"/>
```
frett, line 25: on this screen it should be laid out above `action_tips_finish`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
JasonHammar: Yeah I just noticed it, Whoops.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -68,17 +79,41 @@ internal fun TipsTutorialLayout() = GodToolsTheme() {
...
+            )
+        }
+        if (anim == R.raw.anim_tutorial_tips_light) {
+            Button(
+                onClick = {
+                    onTutorialAction
+                },
+                Modifier.width(20.dp)
+            ) {
+                Text(text = "Start Training")
```
frett, line 111: this should be using the string resource already defined

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -68,17 +79,41 @@ internal fun TipsTutorialLayout() = GodToolsTheme() {
...
+            }
+        } else {
+            Button(onClick = nextPage) {
+                Text(text = "Continue")
```
frett, line 115: this should be using the string resource already defined

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -68,17 +79,41 @@ internal fun TipsTutorialLayout() = GodToolsTheme() {
...
+
+            Text(
+                // Spacer(modifier = Modifier.minLinesHeight(minLines = 1, textStyle = MaterialTheme.typography.bodyMedium)),
+                stringResource(body2),
+                style = MaterialTheme.typography.bodyMedium,
+                textAlign = TextAlign.Center
+            )
+        }
+        if (anim == R.raw.anim_tutorial_tips_light) {
+            Button(
```
frett, line 105: the reason these are not currently pinned to the bottom is that they are in a `Column` that has `verticalArrangement = Arrangement.Center`. This is causing the Buttons to be vertically centered along with the rest of the content in this `Column`

ui/tutorial-renderer/src/main/res/layout/tutorial_tips_start.xml
```diff
@@ -5,69 +5,19 @@
         <variable name="callbacks" type="org.cru.godtools.tutorial.TutorialCallbacks" />
     </data>
 
-    <androidx.constraintlayout.widget.ConstraintLayout
-        android:layout_width="match_parent"
-        android:layout_height="match_parent"
-        android:clipToPadding="false"
-        android:paddingTop="@dimen/tutorial_page_inset_top"
-        android:paddingBottom="@dimen/tutorial_page_inset_bottom">
```
frett, line 13: @JasonHammar this paddingTop and paddingBottom would need to be present on the column as a whole

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -24,18 +29,27 @@ import com.airbnb.lottie.compose.rememberLottieComposition
 import org.cru.godtools.base.ui.theme.GodToolsTheme
 import org.cru.godtools.tutorial.R
 
-@Preview(showBackground = true)
 @Composable
-internal fun TipsTutorialLayout() = GodToolsTheme() {
+internal fun TipsTutorialLayout(
+    nextPage: () -> Unit = {},
+    onTutorialAction: (Int?) -> Unit = {},
```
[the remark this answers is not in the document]
frett: `onTutorialAction: (Int) -> Unit = {},`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
```diff
@@ -50,7 +52,43 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
...
+                    R.string.tutorial_tips_light_headline, R.string.tutorial_tips_light_text1,
+                    R.string.tutorial_tips_light_text2
+                )
+            }
+            is TutorialTipsStartBinding -> compose.setContent {
+                TipsTutorialLayout(
+                    nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
+                    onTutorialAction = {
+                        view?.let { it1 ->
+                            findListener<TutorialCallbacks>()?.onTutorialAction(R.id.action_tips_finish)
```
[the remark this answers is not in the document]
frett: you should be able to write this as:
```
onTutorialAction = { findListener<TutorialCallbacks>()?.onTutorialAction(it) }
```

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -24,18 +29,27 @@ import com.airbnb.lottie.compose.rememberLottieComposition
 import org.cru.godtools.base.ui.theme.GodToolsTheme
 import org.cru.godtools.tutorial.R
 
-@Preview(showBackground = true)
 @Composable
-internal fun TipsTutorialLayout() = GodToolsTheme() {
+internal fun TipsTutorialLayout(
```
frett, line 33: let's also pass in a `page: Page` parameter for this Compose method. you will have access to the current `Page` object from where you call `TipsTutorialLayout()`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -64,21 +78,60 @@ internal fun TipsTutorialLayout() = GodToolsTheme() {
...
+        if (body2 != null) {
+
+            Text(
+                // Spacer(modifier = Modifier.minLinesHeight(minLines = 1, textStyle = MaterialTheme.typography.bodyMedium)),
+                stringResource(body2),
+                style = MaterialTheme.typography.bodyMedium,
+                textAlign = TextAlign.Center
+            )
+        }
+        if (anim == R.raw.anim_tutorial_tips_light) {
```
frett, line 109: this conditional should be branched based on which `Page` you are rendering, and not based on which animation is used. This will be possible once you pass the `Page` object into `TipsTutorialLayout()`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
```diff
@@ -57,21 +57,24 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
                     onTutorialAction = {
                         view?.let { it1 ->
```
frett, line 59: there is no need to wrap the callback in `view?.let {}`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -26,112 +28,137 @@ import com.airbnb.lottie.compose.LottieCompositionSpec
...
+                    textAlign = TextAlign.Center,
+                    modifier = Modifier.padding(horizontal = 32.dp)
                 )
             }
         }
+        Box(
+            modifier = Modifier.weight(1f)
+        ) {
+            if (page == Page.TIPS_START) {
+                Spacer(Modifier.padding(top = 150.dp))
```
frett, line 134: you shouldn't need to have a `Spacer` in this Box

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -26,112 +28,137 @@ import com.airbnb.lottie.compose.LottieCompositionSpec
...
+                Button(
+                    onClick = {
+                        onTutorialAction(R.id.action_tips_finish)
+                    },
+
+                    modifier = Modifier.padding(horizontal = 32.dp).width(275.dp)
```
frett, line 140: there is a `Modifier.align(Alignment.BottomCenter)` that will put the button on the bottom of the box

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -1,84 +1,164 @@
...
-            style = MaterialTheme.typography.bodyLarge,
-            textAlign = TextAlign.Center,
-            modifier = Modifier.padding(top = 16.dp)
-        )
+                    modifier = Modifier.padding(horizontal = 32.dp).width(275.dp)
+                ) {
+                    Text(
+                        text = stringResource(id = R.string.tutorial_tips_action_start),
+                        fontSize = 30.sp,
+                        modifier = Modifier.padding(horizontal = 32.dp)
```
frett, line 145: oops, I meant put the horizontal padding on all Text nodes that are all "root" elements, it doesn't make sense to put this padding on a Text node within the button.

The horizontal padding was meant to provide margin on both sides of the page. We didn't put it on the root layout node because we wanted the animation/image to be able to fill the entire page width.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
```diff
@@ -50,7 +52,35 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
 
     private fun ViewDataBinding.setupCompose() {
         when (this) {
-            is TutorialTipsLearnBinding -> compose.setContent { TipsTutorialLayout() }
+            is TutorialTipsLearnBinding -> compose.setContent {
+                TipsTutorialLayout(
+                    nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
+                    onTutorialAction = { findListener<TutorialCallbacks>()?.onTutorialAction(it) },
+                    Page.TIPS_LEARN,
```
frett, line 59: for this parameter just use the `page` property that is available on the fragment. same goes for the other compose bindings

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -1,84 +1,157 @@
...
+                        stringResource(body2),
+                        textAlign = TextAlign.Center,
+                        modifier = Modifier.padding(horizontal = 32.dp)
+                    )
+                }
+            }
+        }
+        Box(
+            modifier = Modifier.weight(1f).fillMaxWidth()
+        ) {
```
frett, line 129: In order for the button to show up when scrolling, this box needs to be a Spacer with the weight independent from the Buttons.

```
Spacer(modifier = Modifier.weight(1f))
Button(modifier = Modifier.align(Alignment.CenterHorizontally))
```

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TipsLayout.kt
```diff
@@ -1,84 +1,157 @@
...
+            if (page == Page.TIPS_START) {
```
frett, line 130: It'd be nice if we move the conditional into the `onClick` callback and the `Text` node itself so that we don't have to duplicate the same Modifiers on both buttons.

Something like: 
```
Button(
    onClick = {
        if(page == Page.TIPS_START) onTutorialAction(R.id.action_tips_finish) else nextPage()
    }
) {
    Text(stringResource(if(page == Page.TIPS_START) R.string.tutorial_tips_action_start else R.string.tutorial_tips_action_continue))
}

File: ui/tutorial-renderer/src/main/res/layout/tutorial_tips_start.xml [hunk 24]
```diff
@@ -5,69 +5,19 @@
         <variable name="callbacks" type="org.cru.godtools.tutorial.TutorialCallbacks" />
     </data>
```
frett, line 6: we no longer use `callbacks` within the data binding, so you can remove that and the empty `<data>` xml node as well.

ui/tutorial-renderer/src/main/res/layout/tutorial_tips_start.xml
```diff
@@ -5,69 +5,19 @@
...
-        android:paddingBottom="@dimen/tutorial_page_inset_bottom">
 
-        <include layout="@layout/tutorial_include_margins" />
 
-        <include layout="@layout/tutorial_tips_include_positioning" />
+
         <androidx.compose.ui.platform.ComposeView
             android:id="@+id/compose"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
```
frett, line 14: we should update the `layout_width` and `layout_height` to be `match_parent`

ui/tutorial-renderer/src/main/res/layout/tutorial_tips_start.xml
```diff
@@ -5,69 +5,19 @@
...
             app:layout_constraintStart_toStartOf="@id/marginStart"
             app:layout_constraintEnd_toEndOf="@id/marginEnd"
-            app:layout_constraintTop_toTopOf="parent" />
+            app:layout_constraintTop_toTopOf="parent"
+            app:layout_constraintBottom_toTopOf="@id/action_tips_finish"/>
```
frett, line 18: we can remove all the `app:layout_constraint*` attributes now that this view isn't in a `ConstraintLayout`

ui/tutorial-renderer/src/main/res/layout/tutorial_tips_learn.xml
```diff
@@ -1,36 +1,25 @@
...
-    xmlns:app="http://schemas.android.com/apk/res-auto">
+   >
 
     <data>
         <import type="org.cru.godtools.tutorial.activity.TutorialActivity" />
 
-        <variable name="callbacks" type="org.cru.godtools.tutorial.TutorialCallbacks" />
+
         <variable name="formatArgs" type="android.os.Bundle" />
     </data>
```
frett, line 10: you can also remove this `<data>` xml since it's not used anywhere. I'm not going to hold the PR up on this change though.

frett (approved):

PR #2353, opened 2022-06-24
Compose for Live Share Tutorial

JasonHammar:
Compose is being added for the Live share tutorial

### DIFF
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
@@ -90,8 +90,9 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                     nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
                     onTutorialAction = { findListener<TutorialCallbacks>()?.onTutorialAction(it) },
                     page,
-                    R.raw.anim_tutorial_tips_people,
-                    R.string.tutorial_live_share_description_headline, R.string.tutorial_live_share_description_text
+                    null,
+                    R.string.tutorial_live_share_description_headline, R.string.tutorial_live_share_description_text,
+                    R.drawable.img_tutorial_live_share_people
 
                 )
             }
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
@@ -9,7 +9,13 @@ import androidx.databinding.ViewDataBinding
 import androidx.fragment.app.Fragment
 import org.ccci.gto.android.common.androidx.fragment.app.findListener
 import org.cru.godtools.tutorial.animation.animateViews
-import org.cru.godtools.tutorial.databinding.*
+import org.cru.godtools.tutorial.databinding.TutorialLiveShareDescriptionBinding
+import org.cru.godtools.tutorial.databinding.TutorialLiveShareMirroredBinding
+import org.cru.godtools.tutorial.databinding.TutorialLiveShareStartBinding
+import org.cru.godtools.tutorial.databinding.TutorialOnboardingWelcomeBinding
+import org.cru.godtools.tutorial.databinding.TutorialTipsLearnBinding
+import org.cru.godtools.tutorial.databinding.TutorialTipsLightBinding
+import org.cru.godtools.tutorial.databinding.TutorialTipsStartBinding
 import org.cru.godtools.tutorial.layout.TipsTutorialLayout
 import org.cru.godtools.tutorial.layout.TutorialLiveShareLayout
 import splitties.fragmentargs.arg
--- a/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
+++ b/ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
@@ -9,11 +9,9 @@ import androidx.databinding.ViewDataBinding
 import androidx.fragment.app.Fragment
 import org.ccci.gto.android.common.androidx.fragment.app.findListener
 import org.cru.godtools.tutorial.animation.animateViews
-import org.cru.godtools.tutorial.databinding.TutorialOnboardingWelcomeBinding
-import org.cru.godtools.tutorial.databinding.TutorialTipsLearnBinding
-import org.cru.godtools.tutorial.databinding.TutorialTipsLightBinding
-import org.cru.godtools.tutorial.databinding.TutorialTipsStartBinding
+import org.cru.godtools.tutorial.databinding.*
 import org.cru.godtools.tutorial.layout.TipsTutorialLayout
+import org.cru.godtools.tutorial.layout.TutorialLiveShareLayout
 import splitties.fragmentargs.arg
 import splitties.fragmentargs.argOrNull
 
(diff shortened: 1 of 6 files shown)


ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,153 @@
...
+                            textStyle = MaterialTheme.typography.bodyMedium
+                        )
+                    )
+                    Text(
+                        stringResource(body2),
+                        style = MaterialTheme.typography.bodyMedium,
+                        textAlign = TextAlign.Center,
+                        modifier = Modifier.padding(horizontal = 32.dp)
+                    )
+                }
```
frett, line 93: I don't think there is ever a second paragraph of text in the live share tutorial

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/TutorialPageFragment.kt
```diff
@@ -81,6 +85,36 @@ internal class TutorialPageFragment() : Fragment(), TutorialCallbacks {
                     R.string.tutorial_tips_start_headline, R.string.tutorial_tips_start_text
                 )
             }
+            is TutorialLiveShareDescriptionBinding -> compose.setContent {
+                TutorialLiveShareLayout(
+                    nextPage = { findListener<TutorialCallbacks>()?.nextPage() },
+                    onTutorialAction = { findListener<TutorialCallbacks>()?.onTutorialAction(it) },
+                    page,
+                    R.raw.anim_tutorial_tips_people,
```
frett, line 93: this is a tips animation, and not the image used for live share

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,153 @@
...
+                    // pass false to continue the animation
+                    // at which is was paused
+                    restartOnPlay = false
+
+                )
+                LottieAnimation(
+                    composition,
+                    { progress },
+                    modifier = Modifier
+                        .height(290.dp)
```
frett, line 126: based on the xml, this height should be `R.dimen.tutorial_page_live_share_anim_height`

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,153 @@
...
+                bottom = dimensionResource(R.dimen.tutorial_page_inset_bottom)
+            ).verticalScroll(rememberScrollState())
+    ) {
+        Spacer(modifier = Modifier.weight(1f))
+        Box(modifier = Modifier.fillMaxWidth()) {
+            Column() {
+                Spacer(modifier = Modifier.height(290.dp))
+                Spacer(modifier = Modifier.minLinesHeight(1, MaterialTheme.typography.titleLarge))
+                Spacer(modifier = Modifier.minLinesHeight(8, MaterialTheme.typography.bodyMedium))
+            }
```
frett, line 59: we will need to match this spacing to what the content for this tutorial actually is. Plus because the animation/image is last we will want ot make sure those line up between pages. This means you will probably have to tweak how you do the spacer layout a bit

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,159 @@
...
+                    textAlign = TextAlign.Center,
+                    modifier = Modifier.padding(horizontal = 32.dp),
+                    color = MaterialTheme.colorScheme.primary
+                )
+
+                Text(
+                    stringResource(body),
+                    style = MaterialTheme.typography.bodyMedium,
+                    textAlign = TextAlign.Center,
+                    modifier = Modifier.padding(horizontal = 32.dp)
```
frett, line 79: also add `.padding(top = 16.dp)` for this element

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,159 @@
...
+                    // when paused and play
+                    // pass false to continue the animation
+                    // at which is was paused
+                    restartOnPlay = false
+
+                )
+                LottieAnimation(
+                    composition,
+                    { progress },
+                    modifier = Modifier
```
frett, line 131: also add `.padding(top = 16.dp)` for this element. because the order of modifiers matters, you should add the padding before the height

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,159 @@
...
+                        .height(dimensionResource(R.dimen.tutorial_page_live_share_anim_height))
+                        .fillMaxWidth()
+                )
+            }
+        }
+
+        Spacer(modifier = Modifier.weight(1f).fillMaxWidth())
```
frett, line 138: `.fillMaxWidth()` doesn't matter for the `Spacer` because in this use case we only care about the height

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,159 @@
...
+                )
+            }
+        }
+        /*
+        Spacer for first two elements only
+        Animation will not be in box, will be element of parent column
+
+         */
+        if (page == Page.LIVE_SHARE_DESCRIPTION) {
+            if (img != null) {
```
frett, line 89: I would make this entire if block a when block that doesn't even examine `page` in this case.

```kotlin
when {
  anim != null -> {
    // LottieAnimation logic
  }
  img != null -> Image()
}
```

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,159 @@
...
+            .padding(
+                top = dimensionResource(R.dimen.tutorial_page_inset_top),
+                bottom = dimensionResource(R.dimen.tutorial_page_inset_bottom)
+            ).verticalScroll(rememberScrollState())
+    ) {
+        Spacer(modifier = Modifier.weight(1f))
+        Box(modifier = Modifier.fillMaxWidth()) {
+            Column() {
+                Spacer(modifier = Modifier.minLinesHeight(1, MaterialTheme.typography.titleLarge))
+                Spacer(modifier = Modifier.minLinesHeight(5, MaterialTheme.typography.bodyMedium))
```
frett, line 59: because you will be adding `16dp` of top padding to the `body` text, we should reflect that same vertical spacing here by adding `.padding(top=16.dp)` to this `Spacer`

frett: Remember these hidden spacers are meant to reflect the "typical" vertical size of the page content to help with centering that content consistently between all pages. So, any adjustments made to that vertical size need to be reflected in this spacing as well as the actual content

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,157 @@
...
+            .padding(
+                top = dimensionResource(R.dimen.tutorial_page_inset_top),
+                bottom = dimensionResource(R.dimen.tutorial_page_inset_bottom)
+            ).verticalScroll(rememberScrollState())
+    ) {
+        Spacer(modifier = Modifier.weight(1f))
+        Box(modifier = Modifier.fillMaxWidth()) {
+            Column() {
+                Spacer(modifier = Modifier.minLinesHeight(1, MaterialTheme.typography.titleLarge))
+                Spacer(modifier = Modifier.minLinesHeight(5, MaterialTheme.typography.bodyMedium).padding(top = 16.dp))
```
frett, line 59: you need to define the padding before you set the height of the Spacer. the order of modifiers matters. The order flows from the outer layer to the inner layer. (see: https://stackoverflow.com/a/64216931/4721910 for an example of how order matters)

so in this case you are currently setting the height and then the padding. This causes the padding to be taken out of the allocated height. What you want to do is specify the padding before you specify the height of the Spacer.

for reference: 
`Modifier.size(40.dp).padding(8.dp)` would give you a 40dp square element with 8dp of padding inside the 40dp square.
`Modifier.padding(8.dp).size(40.dp)` would give you 8dp of padding around a 40dp square element

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,157 @@
...
+                LottieAnimation(
+                    composition,
+                    { progress },
+                    modifier = Modifier
+                        .height(dimensionResource(R.dimen.tutorial_page_live_share_anim_height))
+                        .fillMaxWidth().padding(top = 16.dp)
+                )
+            }
+            img != null -> Image(
+                painter = painterResource(img), contentDescription = "",
```
frett, line 128: I think you can set `contentDescription` to null, If that is the case I would rather it be null than an empty string

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,157 @@
...
+                modifier = Modifier
+                    .height(dimensionResource(R.dimen.tutorial_page_live_share_anim_height))
+                    .fillMaxWidth().padding(top = 16.dp)
```
frett, line 131: padding needs to be defined before you set the height of the Image. See my previous comment about how modifier order matters.

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialLiveShareLayout.kt
```diff
@@ -0,0 +1,157 @@
...
+                    // at which is was paused
+                    restartOnPlay = false
+
+                )
+                LottieAnimation(
+                    composition,
+                    { progress },
+                    modifier = Modifier
+                        .height(dimensionResource(R.dimen.tutorial_page_live_share_anim_height))
+                        .fillMaxWidth().padding(top = 16.dp)
```
frett, line 124: padding needs to be defined before you set the height of the animation. See my previous comment about how modifier order matters.

frett (approved):

PR #2370, opened 2022-06-29
Implementing compose for onboarding

JasonHammar:
This pull request will implement compose for the onboarding tutorial.


ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialOnboardingLayout.kt
```diff
@@ -0,0 +1,316 @@
...
+            )
+            .padding(horizontal = 32.dp)
+            .verticalScroll(rememberScrollState())
+    ) {
+        Spacer(modifier = Modifier.weight(1f))
+        Box(modifier = Modifier.fillMaxWidth()) {
+            Column() {
+                Spacer(modifier = Modifier.minLinesHeight(1, MaterialTheme.typography.titleLarge))
+                Spacer(modifier = Modifier.padding(top = 16.dp).minLinesHeight(5, MaterialTheme.typography.bodyMedium))
+            }
```
frett, line 63: There should be no need for wrapping the main content in a box with hidden spacers on the welcome page.

frett: Which will allow you to flatten out the welcome page into a single column

ui/tutorial-renderer/src/main/java/org/cru/godtools/tutorial/layout/TutorialOnboardingLayout.kt
```diff
@@ -0,0 +1,316 @@
...
+                    onClick = { onTutorialAction(R.id.action_onboarding_watch_video) },
+                    // style = MaterialTheme.typography.bodyLarge,
+                    style = TextStyle(
+                        color = MaterialTheme.colorScheme.primary,
+
+                    ),
+                    // color = MaterialTheme.colorScheme.primary,
+                    modifier = Modifier.padding(top = 150.dp),
+
+                )
```
frett, line 110: you actually don't need to use `ClickableText` here, you can use `Text` with a `.clickable()` modifier instead.

The `ClickableText` is useful if you are trying to determine where in a text element the user clicked, but the unfortunate drawback is that it's not part of the material3 widgets, so doesn't use the Material3 theme.

frett (approved):

PR #2561, opened 2022-10-21
GT-1828 Configure OneSky gradle plugin

frett:
This configures the `uploadTranslations` and `downloadTranslations` tasks for all subprojects that contain translation files.

There are a few outstanding issues that prevent us from using a GHA workflow for this currently:
- The plugin doesn't support downloading changes to the base language from OneSky. I have opened a PR with the plugin to address this use case
- OneSky currently uses an older version of the CLDR database than Android. This means that exporting French plurals from OneSky will not pass Android lint checks. Manual intervention is required in these cases to fix the broken French plurals
- You have to use the `--no-parallel` command line option when downloading translations to avoid hitting OneSky API rate limits. It might be possible to use `--max-workers #` to restrict the number of threads and not hit the rate limit, but I didn't test where the # of threads threshold should be

### DIFF
--- a/buildSrc/src/main/kotlin/OneSkyConfiguration.kt
+++ b/buildSrc/src/main/kotlin/OneSkyConfiguration.kt
@@ -1,5 +1,6 @@
 import co.brainly.onesky.OneSkyPluginExtension
 import co.brainly.onesky.task.DEPRECATE_STRINGS_FLAG
+import co.brainly.onesky.task.DownloadTranslationsTask
 import org.gradle.api.Project
 import org.gradle.kotlin.dsl.apply
 import org.gradle.kotlin.dsl.configure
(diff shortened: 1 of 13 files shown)


buildSrc/src/main/kotlin/OneSkyConfiguration.kt
```diff
@@ -0,0 +1,47 @@
...
+private const val PROP_API_KEY = "ONESKY_API_KEY"
+private const val PROP_API_SECRET = "ONESKY_API_SECRET"
+
+fun Project.onesky(configuration: OneSkyPluginExtension.() -> Unit) {
+    extra.set(DEPRECATE_STRINGS_FLAG, true)
+    apply(plugin = "co.brainly.onesky")
+
+    configure<OneSkyPluginExtension> {
+        apiKey = findProperty(PROP_API_KEY)?.toString().orEmpty()
+        apiSecret = findProperty(PROP_API_SECRET)?.toString().orEmpty()
```
frett, line 18: @gyasistory here is how you can use parameters without needing to define a placeholder value in the `gradle.properties` file

buildSrc/src/main/kotlin/OneSkyConfiguration.kt
```diff
@@ -0,0 +1,47 @@
...
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.apply
+import org.gradle.kotlin.dsl.configure
+import org.gradle.kotlin.dsl.extra
+
+private const val PROP_API_KEY = "ONESKY_API_KEY"
+private const val PROP_API_SECRET = "ONESKY_API_SECRET"
+
+fun Project.onesky(configuration: OneSkyPluginExtension.() -> Unit) {
+    extra.set(DEPRECATE_STRINGS_FLAG, true)
```
frett, line 13: this avoids having to set the property on the command line

gyasistory (commented):
@frett Wouldn't you want the English files to upload using GitHub Actions?

frett:
@gyasistory The GodTools product team currently makes changes to english copy directly in OneSky. If we straight upload translations to OneSky it will pave over any changes they made to the english copy

gyasistory (approved):

frett:
What I've historically done to update the available phrases in OneSky is:
1. download the current OneSky translations for English
    a. Use git patch mode to merge changes to pre-existing translations, and keep any new translations coming from the source code
2. Upload the freshly merged files from the app to OneSky

This workflow isn't currently automated, and we will probably want to update the workflow with the product team regarding making changes to source translations. I will probably pursue this once Aaron is back because he could easily make the appropriate changes in the strings files directly.

PR #2579, opened 2022-11-03
GT-1263 GT-1832 Create user activity UI

frett:
- rename Counters singleton to UserActivityManager
- use shared UserCounterNames to set the names for user counters
- expose a UserActivity flow
- add "Your activity" content to Account Activity UI


frett:
This currently doesn't increment user counters for languages being used. This will come in a subsequent PR because it's a larger change involving refactoring parts of the base tool renderer logic

gyasistory (approved):

PR #2585, opened 2022-11-07
Fix a bug with switching to a new isTipCompleteFlow()

frett:
- add a unit test illustrating broken behavior in TipBottomSheetDialogFragmentDataModel
- use combineTransformLatest to fix switching tip complete flows


ui/tips-renderer/src/test/kotlin/org/cru/godtools/tool/tips/ui/TipBottomSheetDialogFragmentDataModelTest.kt
```diff
@@ -0,0 +1,93 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class TipBottomSheetDialogFragmentDataModelTest {
+    private companion object {
+        private const val TOOL = "tool"
+        private const val TIPID1 = "tipId1"
```
wrandall22, line 28: Is this normal naming convention in Kotlin over `TIP_ID_1`?

frett: nope, I can rename it. it's only used internally in this test

wrandall22 (approved):

PR #2609, opened 2022-11-21
GT-1856 Fix a crash if an invalid tool is returned from the API

frett:
In theory the api should never return an invalid tool...


wrandall22 (approved):
Unit tests?

PR #2642, opened 2022-12-15
GT-1791 Fix syncing followups to the API

frett:
In the various migrations from the legacy DAO to Room I missed updating the logic loading the language object from the database which is used to populate the Followup languageId. This issue would have triggered an exception, but I believe that exception was lost because it was running in a background coroutine.

After fixing the issue I added several tests for each part of the bug and fix implemented.

### DIFF
--- a/library/sync/src/main/kotlin/org/cru/godtools/sync/task/FollowupSyncTasks.kt
+++ b/library/sync/src/main/kotlin/org/cru/godtools/sync/task/FollowupSyncTasks.kt
@@ -7,16 +7,15 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
-import org.ccci.gto.android.common.db.find
 import org.cru.godtools.api.FollowupApi
 import org.cru.godtools.db.repository.FollowupsRepository
-import org.keynote.godtools.android.db.GodToolsDao
+import org.cru.godtools.db.repository.LanguagesRepository
 
 @Singleton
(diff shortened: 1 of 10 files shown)


library/sync/src/main/kotlin/org/cru/godtools/sync/task/FollowupSyncTasks.kt
```diff
@@ -26,7 +25,7 @@ internal class FollowupSyncTasks @Inject internal constructor(
                 .map { followup ->
                     async {
                         try {
-                            followup.languageCode?.let { followup.setLanguage(dao.find(it)) }
+                            followup.setLanguage(languagesRepository.findLanguage(followup.languageCode))
```
wrandall22, line 28: It looks to me like the old code only called setLanguage if languageCode existed, but now we are calling it regardless. This is likely my ignorance of Kotlin, but, what happens if you pass in a `null` `languageCode` here? Maybe we should have a test for that case?

frett: it won't be null anymore because `languageCode` is now defined as non-null. The underlying model changed in other refactoring and I noticed the warning on the null-safe operator (`?.`) when working with this code.

If the model ever changes back to having `languageCode` as nullable, this code will throw a compile-time error and we will have to address the null condition before we can compile

frett: The referenced property:

https://github.com/CruGlobal/godtools-android/blob/develop/library/model/src/main/kotlin/org/cru/godtools/model/Followup.kt#L23

wrandall22 (approved):

PR #2651, opened 2023-01-04
GT-1897 Create a ToolsRepository to decouple the database from the rest of the app

frett:


frett:
I'm really sorry about the massive PR, thankfully each commit should be really self-contained if you wanted to just take it a commit at a time instead

gyasistory (approved):

PR #2665, opened 2023-01-13
Disable redundant Unit Tests

frett:
Instead of running the same unit tests on every build variant, we only run them on production debug variants.


frett:
I'll revisit this in the future, it's currently conflicting with jacoco's configuration

PR #2711, opened 2023-02-21
GT-1912 Don't log cancellation exceptions

frett:
when a coroutine is cancelled it triggers a `CancellationException`, this is normal behavior for coroutines.

This specific cancellation was primarily being triggered because we launch a sync of the available tools within the lifecycle of the dashboard activity, if the user leaves the dashboard (e.g. exiting the app) it cancels the sync.

We still schedule the sync to happen in the background later by using workManager

### DIFF
--- a/library/sync/src/main/kotlin/org/cru/godtools/sync/GodToolsSyncService.kt
+++ b/library/sync/src/main/kotlin/org/cru/godtools/sync/GodToolsSyncService.kt
@@ -10,6 +10,7 @@ import java.io.IOException
 import javax.inject.Inject
 import javax.inject.Provider
 import javax.inject.Singleton
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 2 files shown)


wrandall22:
I see how `We still schedule the sync to happen in the background later by using workManager` is happening, but I don't see how `Don't log cancellation exceptions` is happening here.

frett:
> I see how `We still schedule the sync to happen in the background later by using workManager` is happening, but I don't see how `Don't log cancellation exceptions` is happening here.

in the `executeSync` function there was a catch-all `Exception` in the try-catch, it logged any exceptions not previously handled (which included `CancellationException` before this PR.

frett:
https://github.com/CruGlobal/godtools-android/blob/ba7f11bff4c93b5122ae575714f965faa2248b24/library/sync/src/main/kotlin/org/cru/godtools/sync/GodToolsSyncService.kt#L105-L108

wrandall22:
I gotcha.

wrandall22 (approved):

frett:
just want to push up my tests then I'll merge it.

I had to figure out how to make MockK suspend indefinitely when calling a suspendable method so that I could deterministically test the cancel behavior.

PR #2798, opened 2023-04-05
GT-1992 correctly handle exceptions when running syncUserCounters()

frett:
- cleanup GodToolsSyncServiceTest
- correctly handle exceptions when running syncUserCounters()

### DIFF
--- a/library/sync/src/main/kotlin/org/cru/godtools/sync/GodToolsSyncService.kt
+++ b/library/sync/src/main/kotlin/org/cru/godtools/sync/GodToolsSyncService.kt
@@ -137,9 +137,11 @@ class GodToolsSyncService @VisibleForTesting internal constructor(
 
     suspend fun syncDirtyUserCounters() = executeSync<UserCounterSyncTasks> { syncDirtyCounters() }
     suspend fun syncUserCounters(force: Boolean = false) = executeSync<UserCounterSyncTasks> {
-        val resp = syncCounters(force)
-        coroutineScope.launch { syncDirtyCounters() }
-        resp
+        try {
+            syncCounters(force)
+        } finally {
+            coroutineScope.launch { syncDirtyUserCounters() }
+        }
     }
 
     fun syncToolShares(): SyncTask = GtSyncTask(bundleOf(EXTRA_SYNCTYPE to SYNCTYPE_TOOL_SHARES))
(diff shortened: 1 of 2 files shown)


library/sync/src/main/kotlin/org/cru/godtools/sync/GodToolsSyncService.kt
```diff
@@ -137,9 +137,11 @@ class GodToolsSyncService @VisibleForTesting internal constructor(
 
     suspend fun syncDirtyUserCounters() = executeSync<UserCounterSyncTasks> { syncDirtyCounters() }
     suspend fun syncUserCounters(force: Boolean = false) = executeSync<UserCounterSyncTasks> {
-        val resp = syncCounters(force)
-        coroutineScope.launch { syncDirtyCounters() }
-        resp
+        try {
+            syncCounters(force)
+        } finally {
+            coroutineScope.launch { syncDirtyUserCounters() }
```
frett, line 143: I switched from using `UserCounterSyncTasks.syncDirtyCounters()` to using `GodToolsSyncService.syncDirtyUserCounters()` here.

They are named very similarly because the `GodToolsSyncService` provides a common wrapper API for all the underlying sync tasks that are internal. By going through the wrapper method we gain usage of the `executeSync` DSL that handles any thrown exceptions for us.

An alternative solution is to call `executeSync` directly from within `coroutineScope.launch { }`

library/sync/src/test/kotlin/org/cru/godtools/sync/GodToolsSyncServiceTest.kt
```diff
@@ -19,30 +18,38 @@ import kotlinx.coroutines.test.TestScope
...
 import org.cru.godtools.sync.work.scheduleSyncToolsWork
 import org.junit.After
+import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Test
-import org.junit.runner.RunWith
 import timber.log.Timber
 
-@RunWith(AndroidJUnit4::class)
```
wrandall22, line 32: For my own clarification; why is this no longer needed?

wrandall22 (approved):

library/sync/src/test/kotlin/org/cru/godtools/sync/GodToolsSyncServiceTest.kt
frett: Uses the Android test runner loads the Android apis with robolectric. It adds about 10 seconds to test startup.

Robolectric is only necessary if something in the test or objects being tested references/uses an android api

PR #2991, opened 2023-07-14
Workaround flakiness in SSL certificate validation for S3

frett:
This is being caused by an outdated Public Suffix List occassionaly being loaded from the wrong jar file instead of the correct more up to date one


frett:
I'm going to leave this as a draft until it's absolutely necessary to merge this

PR #3131, opened 2023-10-03
GT-2145 create a normalized_app_language

frett:


library/base/src/main/res/values-in/values.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="normalized_app_language">id</string>
```
gyasistory, line 3: Shouldn't this be "in"

gyasistory (changes requested):

library/base/src/main/res/values-in/values.xml
frett: It actually doesn't matter because of how the locale object handles the Indonesian language. But `id` is the correct language tag for Indonesian.

Internally the locale object converts Indonesian to `in` due to old behavior and backwards compatibility. This conversion means that older versions of Android need a `values-in` directory, and newer versions prefer `values-id`. To ensure compatibility we just duplicate the correct `values-id` to the incorrect `values-in`

gyasistory (approved):

PR #3407, opened 2024-03-07
Setup Paparazzi for composable screenshot tests

frett:
Paparazzi provides us the ability to validate UIs against golden values to detect when a UI changes due to a change we make.

This will hopefully provide stability for our UIs as we update underlying data models and rendering logic. 
Also, when we make an intentional change to a UI component, it should be possible to preview the actual UI differences during code review.


frett:
@aaronlaib this might be of interest to you and Jill, you can see screenshots of the various UI elements that paparazzi is tracking [here](https://github.com/CruGlobal/godtools-android/tree/setupPaparazzi/app/src/test/snapshots/images).

gyasistory (approved):

PR #3453, opened 2024-04-09
Convert LanguageSettings to utilize Circuit

frett:
- **update LanguageSettingsLayout to render using LanguageSettingsScreen.State**
- **move the DrawerMenuLayout composable into the LanguageSettingsLayout composable**
- **create a LanguageSettingsPresenter**
- **update LanguageSettingsActivity to render the UI using Circuit**
- **record the language settings feature discovered in the LanguageSettingsPresenter**


frett:
I'm not sure why the accessibility screenshot is overlapping content in the column, my guess it's something to do with how accessibility screenshots manipulate the screen space to include the accessibility data.

it's also pulling in accessibility content from the nav drawer which is currently closed, I think this is because the content is still in the composable tree even though it's not currently visible.

gyasistory (approved):

PR #3620, opened 2024-07-30
Gt-2322  : Add two languages to forced name

gyasistory:


app/src/main/kotlin/org/cru/godtools/ui/languages/LanguageName.kt
```diff
@@ -41,28 +41,24 @@ internal fun LanguageName(language: Language, modifier: Modifier = Modifier) {
 
     LanguageName(
         name = remember(context, language, appLanguage) { language.getDisplayName(context, language.code) },
-        secondName = remember(context, language) {
-            language.takeUnless { it.isForcedName }?.getDisplayName(context, appLanguage)
-        },
+        secondName = remember(context, language) { language.getDisplayName(context, appLanguage) },
```
frett, line 44: I just noticed that the remember keys are backwards for name and secondName.

`name` only references `context` and `language`, so it doesn't need `appLanguage` as a key.
`secondName` references `context`, `language`, and `appLanguage`, but is currently missing `appLanguage` as a key

frett (changes requested):
noticed an adjacent bug that we should just fix in this PR

app/src/main/kotlin/org/cru/godtools/ui/languages/LanguageName.kt
gyasistory: Hi Daniel,

This is the merged in change is this missing the proper support?

https://github.com/dependabot/dependabot-core/pull/7525

gyasistory:
> noticed an adjacent bug that we should just fix in this PR

@frett was this what you commented about?

frett (approved):

PR #3629, opened 2024-08-05
GT:2404   Add a language filter on lesson dashboard

gyasistory:


library/db/src/main/kotlin/org/cru/godtools/db/repository/ToolsRepository.kt
```diff
@@ -23,6 +23,7 @@ interface ToolsRepository {
     fun getDownloadedToolsFlowByTypesAndLanguage(types: Collection<Tool.Type>, locale: Locale): Flow<List<Tool>>
     fun getMetaToolsFlow() = getToolsFlowByType(Tool.Type.META)
     fun getLessonsFlow() = getToolsFlowByType(Tool.Type.LESSON)
+    fun getLessonsFlowByLanguage(locale: Locale): Flow<List<Tool>>
```
frett, line 26: you should write a unit test for this new method

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,92 @@
...
+    val selectedLanguage = _selectedLanguage.asStateFlow()
+    fun updateSelectedLanguage(language: Language) {
+        _selectedLanguage.value = language
+    }
+
+    private var _query = MutableStateFlow("")
+    val query = _query.asStateFlow()
+    fun updateQuery(query: String) {
+        _query.value = query
+    }
```
frett, line 52: why didn't you just use `val query = MutableStateFlow("")`

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,140 @@
...
+                    label = { LanguageName(lang) },
+                    supportingText = pluralStringResource(
+                        R.plurals.dashboard_tools_section_filter_available_tools,
+                        count,
+                        count,
+                    ),
+                    onClick = {
+                        scope.launch {
+                            expanded = false
+                            viewModel.settings.appLanguage = lang.code
```
frett, line 131: why are you updating the appLanguage from the Lessons UI?

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,140 @@
...
+                HorizontalDivider(Modifier.padding(start = ListItemStartPadding, end = ListItemStartPadding))
+                FilterMenuItem(
+                    label = { LanguageName(lang) },
+                    supportingText = pluralStringResource(
+                        R.plurals.dashboard_tools_section_filter_available_tools,
+                        count,
+                        count,
+                    ),
+                    onClick = {
+                        scope.launch {
```
frett, line 129: why are you launching this in a coroutine?

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,92 @@
...
+    @VisibleForTesting
+    private val lessonsFlow = _selectedLanguage.flatMapLatest { languages ->
+        toolsRepository.getLessonsFlowByLanguage(languages.code)
+    }
+    val lessons = lessonsFlow
         .map { it.filterNot { it.isHidden }.sortedBy { it.defaultOrder }.mapNotNull { it.code } }
         .stateIn(viewModelScope, SharingStarted.WhileSubscribed(), emptyList())
 
+    @VisibleForTesting
+    private val languages = languagesRepository.getLanguagesFlow()
```
frett, line 66: `@VisibleForTesting` doesn't make sense here, this property isn't actually visible for testing

gyasistory:
@frett, do you have any idea why the ToolFilterTest would fail when I run all the tests together?  But if I run it separately it passes.

frett:
> @frett, do you have any idea why the ToolFilterTest would fail when I run all the tests together? But if I run it separately it passes.

@gyasistory what is the error when the test fails?

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
frett: by setting the appLanguage you are changing the language for the entire app, that dropdown shouldn't be changing the app language

gyasistory:
> > @frett, do you have any idea why the ToolFilterTest would fail when I run all the tests together? But if I run it separately it passes.
> 
> @gyasistory what is the error when the test fails?

godtools_–_LessonsLayout_kt__godtools_app_main_

frett:
> > > @frett, do you have any idea why the ToolFilterTest would fail when I run all the tests together? But if I run it separately it passes.
> > 
> > 
> > @gyasistory what is the error when the test fails?
> 
> godtools_–LessonsLayout_kt__godtools_app_main

so there is an exception being thrown somewhere else before the test runs, does that exception list which exception caused it?

gyasistory:
> > > > @frett, do you have any idea why the ToolFilterTest would fail when I run all the tests together? But if I run it separately it passes.
> > > 
> > > 
> > > @gyasistory what is the error when the test fails?
> > 
> > 
> > godtools_–LessonsLayout_kt__godtools_app_main
> 
> so there is an exception being thrown somewhere else before the test runs, does that exception list which exception caused it?

I think I found it.  It is in the dashboardActivityTest

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
gyasistory: OK thanks

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,139 @@
...
+import org.cru.godtools.ui.languages.LanguageName
+
+private val DROPDOWN_LESSON_MAX_HEIGHT = 700.dp
+private val DROPDOWN_LESSON_MAX_WIDTH = 350.dp
+
+@Composable
+fun LessonFilters(modifier: Modifier = Modifier) {
+    Column(modifier.fillMaxWidth()) {
+        Row(
+            verticalAlignment = Alignment.CenterVertically,
```
frett, line 54: you should be using a baseline alignment between the label and the dropdown menu and not just centering it vertically

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,139 @@
...
+            horizontalArrangement = Arrangement.SpaceEvenly
```
frett, line 55: you shouldn't need to specify a horizontalArrangement since the items in the row are set to take up the entire space

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,140 @@
...
+
+@Composable
+fun LessonFilters(modifier: Modifier = Modifier) {
+    Column(modifier.fillMaxWidth()) {
+        HorizontalDivider(modifier = Modifier.padding(vertical = 12.dp))
+        Row {
+            Text(
+                stringResource(R.string.dashboard_lessons_section_filter_label),
+                modifier = Modifier
+                    .align(Alignment.CenterVertically)
```
frett, line 57: you need to look into aligning by the text baseline instead of just centering the 2 elements

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsLayout.kt
```diff
@@ -30,18 +30,23 @@ internal sealed interface DashboardLessonsEvent {
 @OptIn(ExperimentalFoundationApi::class)
 internal fun LessonsLayout(viewModel: LessonsViewModel = viewModel(), onEvent: (DashboardLessonsEvent) -> Unit = {}) {
     val lessons by viewModel.lessons.collectAsState()
+    val selectedLanguage by viewModel.selectedLanguage.collectAsState()
+
     LazyColumn(contentPadding = PaddingValues(16.dp)) {
         item("header", "header") { LessonsHeader() }
 
+        item("filter", "filter") { LessonFilters() }
```
frett, line 38: let's just include the composable call to `LessonFilters()` in the `item("header", "header")` item

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,86 @@
...
+    fun updateSelectedLanguage(newLanguage: Language) {
+        selectedLanguage.value = newLanguage
+    }
+
+    val query = MutableStateFlow("")
+    fun updateQuery(newQuery: String) {
+        query.value = newQuery
+    }
+
+    private val allLessons = toolsRepository.getLessonsFlow()
```
frett, line 50: you are only using this flow in one other place in the ViewModel, let's simplify the usage by directly using `toolsRepository.getLessonsFlow()` where you are using `allLessons`

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,86 @@
...
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    private val lessonsFlow = selectedLanguage.flatMapLatest { languages ->
```
frett, line 53: the `languages` variable is an inaccurate name and can lead to confusion. when I first read it I thought it was a list of languages and not just your currently selected language. Rename it to something singular instead of plural

frett: you are also only using `lessonsFlow` in one spot within the ViewModel, please simplify by moving the `selectedLanguage.flatMapLatest {}` to where you are using `lessonsFlow`

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,86 @@
...
+    @ApplicationContext
+    private val context: Context,
+    private val eventBus: EventBus,
+    toolsRepository: ToolsRepository,
+    languagesRepository: LanguagesRepository,
+    private val translationsRepository: TranslationsRepository,
+    val settings: Settings,
+) : ViewModel() {
+
+    val selectedLanguage = MutableStateFlow(Language(settings.appLanguage))
```
frett, line 40: you should be storing the currently selected language in the ViewModel `SavedStateHandle` so that the selection persists through activity re-creation

app/src/test/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModelTest.kt
```diff
@@ -62,9 +95,50 @@ class LessonsViewModelTest {
...
         }
     }
+
+    @Test
+    fun `Property lessons - Filter by language`() = testScope.runTest {
+        val first = randomTool("first", Tool.Type.LESSON, defaultOrder = 1, isHidden = false)
+        val second = randomTool("second", Tool.Type.LESSON, defaultOrder = 2, isHidden = false)
+
+        viewModel.lessons.test {
+            viewModel.updateSelectedLanguage(Language(Locale.ENGLISH))
```
frett, line 110: you should verify that the repository is being properly queried with whatever your current selected language is. Take a look at `coVerify {}` for how to do it

app/src/test/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModelTest.kt
```diff
@@ -62,9 +95,50 @@ class LessonsViewModelTest {
...
+    @Test
+    fun `Property filteredLanguages - Query Filter languages`() = testScope.runTest {
+        val first = Language(Locale.ENGLISH)
+        val second = Language(Locale.FRENCH)
+
+        viewModel.filteredLanguages.test {
+            viewModel.updateQuery("Eng")
+            languageFlow.value = listOf(first, second)
+            runCurrent()
+            assertEquals(listOf(Filter(first, anyInt())), expectMostRecentItem())
```
frett, line 141: you shouldn't be using `anyInt()` here, that is an argument matcher when you are trying to match calls in a Mockito call, not when you are comparing values outside of a Mockito matcher

app/src/test/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModelTest.kt
```diff
@@ -62,9 +95,50 @@ class LessonsViewModelTest {
...
+        }
+    }
+
+    @Test
+    fun `Property filteredLanguages - Filter languages`() = testScope.runTest {
+        val first = Language(Locale.ENGLISH)
+        val second = Language(Locale.FRENCH)
+
+        viewModel.filteredLanguages.test {
+            languageFlow.value = listOf(first, second)
```
frett, line 126: set the value of the `languageFlow` before you start testing `filteredLanguages`, then you shouldn't need to use `runCurrent()` and can using `awaitItem()` instead of `expectMostRecentItem()`

app/src/test/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModelTest.kt
```diff
@@ -62,9 +95,50 @@ class LessonsViewModelTest {
...
+    fun `Property lessons - Filter by language`() = testScope.runTest {
+        val first = randomTool("first", Tool.Type.LESSON, defaultOrder = 1, isHidden = false)
+        val second = randomTool("second", Tool.Type.LESSON, defaultOrder = 2, isHidden = false)
+
+        viewModel.lessons.test {
+            viewModel.updateSelectedLanguage(Language(Locale.ENGLISH))
+            lessonsFlow.value = listOf(first, second)
+            lessonsByLanguageFlow.value = listOf(first)
+            appLanguagesFlow.value = Locale.ENGLISH
+            languageFlow.value = listOf(Language(Locale.ENGLISH))
```
frett, line 114: for organization I like setting up my state before starting the test of the flow

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,140 @@
...
+import org.cru.godtools.ui.dashboard.tools.FilterMenuItem
+import org.cru.godtools.ui.languages.LanguageName
+
+private val DROPDOWN_LESSON_MAX_HEIGHT = 700.dp
+private val DROPDOWN_LESSON_MAX_WIDTH = 350.dp
+
+@Composable
+fun LessonFilters(modifier: Modifier = Modifier) {
+    Column(modifier.fillMaxWidth()) {
+        HorizontalDivider(modifier = Modifier.padding(vertical = 12.dp))
```
frett, line 52: move this horizontal divider out to the overall `LessonsLayout`, that way your `LessonFilters` composable is just the filter drop down row

frett: that will also allow you to get rid of the additional `Column` you are using here

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,140 @@
...
+            expanded = expanded,
+            onDismissRequest = { expanded = false },
+            modifier = Modifier.sizeIn(maxHeight = DROPDOWN_LESSON_MAX_HEIGHT, maxWidth = DROPDOWN_LESSON_MAX_WIDTH)
+        ) {
+            stickyHeader {
+                Surface(color = MaterialTheme.colorScheme.surface) {
+                    SearchBar(
+                        query = query,
+                        onQueryChange = { change ->
+                            scope.launch {
```
frett, line 101: why are you launching this in a coroutine?

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,140 @@
...
+                        onSearch = {},
+                        active = false,
+                        onActiveChange = {},
+                        colors = GodToolsTheme.searchBarColors,
+                        placeholder = {
+                            Text(stringResource(R.string.language_settings_downloadable_languages_search))
+                        },
+                        content = {},
+                        modifier = Modifier
+                            .padding(horizontal = 16.dp, vertical = 16.dp)
```
frett, line 115: I feel like this padding is wrong, that would put a lot of additional vertical whitespace around the search bar in a UI that should probably be more compact

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,86 @@
...
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    private val lessonsFlow = selectedLanguage.flatMapLatest { languages ->
+        toolsRepository.getLessonsFlowByLanguage(languages.code)
+    }
+    val lessons = lessonsFlow
         .map { it.filterNot { it.isHidden }.sortedBy { it.defaultOrder }.mapNotNull { it.code } }
         .stateIn(viewModelScope, SharingStarted.WhileSubscribed(), emptyList())
 
+    private val languages = languagesRepository.getLanguagesFlow()
```
frett, line 60: this is also only used in 1 spot, since this is a simple flow lets simplify by just directly using `languagesRepository.getLanguagesFlow()` when constructing `filteredLanguages`

app/src/main/kotlin/org/cru/godtools/ui/dashboard/tools/ToolFilters.kt
```diff
@@ -222,7 +222,7 @@ private fun FilterMenuItem(
 )
 
 @Composable
-private fun FilterMenuItem(
+fun FilterMenuItem(
```
frett, line 225: make this internal, this function should have no use outside of the tool drop down filters and doesn't need to be exposed as a public API.

It's a good habit to get into to think about the intended usage of functions to determine if it's something that you are intending to be publicly available for everyone or just something that you are using internally in this module only.

frett: I'm also not a total fan of using this Composable in a different layout how you are currently using it.

It's introducing a cross-dependency between different composables, if we need to come back and refactor the tools UI and change functionality/UI there it will have a potential impact on a completely different composable. The fact that you needed to increase the visibility to access this method is an indicator that you are potentially using code outside of it's original intended purpose.

The 2 potential alternative solutions are:
- just duplicate this method or use the `ListItem` composable directly in your `LessonsFilter` composable
- re-organize the code so that there is a common "Filters" package for the dashboard that is used by both the tools and lessons UIs

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,140 @@
...
+                            .fillMaxWidth()
+                            .wrapContentWidth()
+                    )
+                }
+            }
+
+            items(lessonLanguages, key = { (lang) -> lang.code }) { (lang, count) ->
+                HorizontalDivider(Modifier.padding(start = ListItemStartPadding, end = ListItemStartPadding))
```
frett, line 123: There should be no horizontal divider between the searchbar and the first language

One of these is the current language filter from the tools UI, the other is your new Lessons filter drop down
Screenshot 2024-08-12 at 3 43 27 PM
Screenshot 2024-08-12 at 3 43 19 PM

frett: these screenshots also show the issue I mentioned with having extra vertical whitespace

frett:
Changing the language in the dropdown does not update the lesson tool cards to reflect the newly selected language

Screenshot 2024-08-12 at 3 42 50 PM

frett (dismissed):
See the comments

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
frett: Screenshot 2024-08-12 at 3 50 56 PM

This image shows that the text is not aligned by the baseline, there is whitespace between the red line and "Spanish", but no white space between the red line and "Lessons In:"

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,140 @@
...
+                FilterMenuItem(
+                    label = { LanguageName(lang) },
+                    supportingText = pluralStringResource(
+                        R.plurals.dashboard_tools_section_filter_available_tools,
+                        count,
+                        count,
+                    ),
```
frett, line 130: The phrase for this dropdown should be "Lessons available" and not "Tools available"
Screenshot 2024-08-13 at 9 36 30 AM

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,86 @@
...
+        .map { it.mapNotNullTo(mutableSetOf()) { it.code } }
+        .distinctUntilChanged()
+        .flatMapLatest { translationsRepository.getTranslationsFlowForTools(it) }
+        .map { translations ->
+            translations
+                .groupBy { it.languageCode }
+                .mapValues { it.value.distinctBy { it.toolCode }.count() }
+        }
+
+    val filteredLanguages = combine(
```
frett, line 73: you shouldn't be including languages with "0 Lessons available"
Screenshot 2024-08-13 at 9 36 30 AM

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,82 @@
...
+    val query = MutableStateFlow("")
+    fun updateQuery(newQuery: String) {
+        query.value = newQuery
+    }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    val lessons = selectedLanguage
+        .flatMapLatest { language ->
+            toolsRepository.getLessonsFlowByLanguage(language.code)
+        }.map { it.filterNot { it.isHidden }.sortedBy { it.defaultOrder }.mapNotNull { it.code } }
```
frett, line 54:
```suggestion
    val lessons = selectedLanguage
        .flatMapLatest { language -> toolsRepository.getLessonsFlowByLanguage(language.code) }
        .map {
            it.filterNot { it.isHidden }
                .sortedBy { it.defaultOrder }
                .mapNotNull { it.code }
        }
```

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsLayout.kt
```diff
@@ -67,4 +71,6 @@ private fun LessonsHeader() = Column {
         style = MaterialTheme.typography.bodyMedium,
         modifier = Modifier.padding(top = 4.dp)
     )
+    HorizontalDivider(modifier = Modifier.padding(vertical = 12.dp))
+    LessonFilters()
```
frett, line 75: what I meant by the comment about this was to include in the `item("header", "header") {}` block directly, like this:

```
item("header", "header") {
    LessonsHeader()
    HorizontalDivider(modifier = Modifier.padding(vertical = 12.dp))
    LessonsFilters()
}
```

app/src/test/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModelTest.kt
```diff
@@ -25,31 +27,41 @@ import org.cru.godtools.model.Language
...
 
     private val languagesRepository: LanguagesRepository = mockk {
         every { getLanguagesFlow() } returns languageFlow
     }
 
+    private val savedStateHandle: SavedStateHandle = mockk {
+        every { get<Locale>(KEY_SAVED_LESSON_LOCALE) } returns savedLessonLocale.value
+        every { getStateFlow(KEY_SAVED_LESSON_LOCALE, Locale.ENGLISH) } returns savedLessonLocale
+        every { set<Locale>(KEY_SAVED_LESSON_LOCALE, any()) } returns Unit
+    }
```
frett, line 57: for simplicity you should be able to use an actual `SavedStateHandle()` object

app/src/main/kotlin/org/cru/godtools/ui/tools/ToolCardLayouts.kt
```diff
@@ -122,10 +125,16 @@ fun LessonToolCard(
 
                 val appLanguage by viewModel.appLanguage.collectAsState()
                 val appTranslation by viewModel.appTranslation.collectAsState()
+                val selectedLanguage by lessonViewModel.selectedLanguage.collectAsState(initial = appLanguage)
+                val availableLanguages = if (isLessonDashboard) {
```
frett, line 129: this shouldn't be plural

app/src/main/kotlin/org/cru/godtools/ui/tools/ToolCardLayouts.kt
```diff
@@ -98,6 +99,8 @@ fun PreloadTool(tool: Tool) {
 fun LessonToolCard(
     toolCode: String,
     modifier: Modifier = Modifier,
+    isLessonDashboard: Boolean = false,
+    lessonViewModel: LessonsViewModel = viewModel(),
```
frett, line 103: I'm not a fan of accessing the `LessonsViewModel` from here, this goes back to the issue I brought up under the `FilterMenuItem` about cross-feature usage. Another issue that this creates is that we are now creating a `LessonsViewModel` wherever we use the `LessonToolCard` composable, which includes on the favorites UI.

Long term this will need to be updated to be usable by circuit, but a better short-term solution would to be to pass in an optional `inLanguage: Locale? = null` parameter that is prefered over the `appLanguage` in the composable.

frett:
The language selector doesn't properly respect the configured forced language name setting. I selected "Arabic (Middle East)", but it is showing as "Arabic (Bahrain)".

Screenshot 2024-08-15 at 2 52 13 PM

frett:
The lesson cards are still being displayed in the wrong language

Screenshot 2024-08-15 at 3 02 05 PM

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,145 @@
...
+
+@OptIn(ExperimentalMaterial3Api::class, ExperimentalFoundationApi::class)
+@Composable
+internal fun LessonLanguageFilter(modifier: Modifier = Modifier, viewModel: LessonsViewModel = viewModel()) {
+    val context = LocalContext.current
+    val selectedLanguage by viewModel.selectedLanguage.collectAsState(initial = Language(Locale.getDefault()))
+
+    val lessonLanguages by viewModel.filteredLanguages.collectAsState(emptyList())
+    var expanded by remember { mutableStateOf(false) }
+    var query by remember { mutableStateOf("") }
```
frett, line 73: I think we should reset the `query` whenever the dropdown menu is dismissed. it feels like the query should just be localized to that menu.

An easy way to do this is to move the `query` state into the `stickyHeader`

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,145 @@
...
+            onDismissRequest = { expanded = false },
+            modifier = Modifier.sizeIn(maxHeight = DROPDOWN_LESSON_MAX_HEIGHT, maxWidth = DROPDOWN_LESSON_MAX_WIDTH)
+        ) {
+            stickyHeader {
+                Surface(color = MaterialTheme.colorScheme.surface) {
+                    SearchBar(
+                        query = query,
+                        onQueryChange = { change ->
+                            query = change
+                            viewModel.updateQuery(change)
```
frett, line 97: maintaining the same data in 2 locations is prone to bugs as code evolves over time. You should try and refactor this to have a single source of truth

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,115 @@
...
+    val settings: Settings,
+) : ViewModel() {
+    val selectedLanguage = languagesRepository.getLanguagesFlow().combine(
+        savedStateHandle.getStateFlow(
+            KEY_SAVED_LESSON_LANGUAGE_LOCALE,
+            settings.appLanguage
+        )
+    ) { languages, locale ->
+        languages.findLast { it.code == locale } ?: Language(settings.appLanguage)
+    }
```
frett, line 50: you could have also done:

```kotlin
val selectedLanguage = savedStateHandle.getStateFlow(KEY_SAVED_LESSON_LANGUAGE_LOCALE, settings.appLanguage)
    .flatMapLatest { locale ->
        languagesRepository.findLanguageFlow(locale)
            .map { it ?: Language(locale) }
    }
```

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsLayout.kt
```diff
@@ -29,19 +31,28 @@ internal sealed interface DashboardLessonsEvent {
...
     LazyColumn(contentPadding = PaddingValues(16.dp)) {
-        item("header", "header") { LessonsHeader() }
+        item("header", "header") {
+            LessonsHeader()
+            HorizontalDivider(modifier = Modifier.padding(vertical = 12.dp))
+            LessonFilters()
+        }
 
-        items(lessons, { it }, { "lesson" }) { lesson ->
+        items(lessonsWithTranslations, { it }, { "lesson" }) { (lesson, translation) ->
```
frett, line 44: you need to update the `key` for this, by changing the data model now whenever you change languages, or translations reload it will not reuse the same card and you will get strange placement animations

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
frett: this still need to be changed

frett: instead of loading all languages, then searching for the language you want, you should just directly fetch the language you want

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,112 @@
...
+        savedStateHandle.set(KEY_SAVED_LESSON_LANGUAGE_LOCALE, newLanguage.code)
+    }
+
+    var query = MutableStateFlow("")
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    val lessons = selectedLanguage
+        .flatMapLatest { language ->
+            toolsRepository.getLessonsFlowByLanguage(language.code)
+        }
```
frett, line 62:
```suggestion
    val lessons = selectedLanguage.flatMapLatest { toolsRepository.getLessonsFlowByLanguage(it.code) }
```

frett: or
```suggestion
    val lessons = selectedLanguage
        .flatMapLatest { toolsRepository.getLessonsFlowByLanguage(it.code) }
```

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,144 @@
...
+                        count,
+                        count,
+                    ),
+                    onClick = {
+                        expanded = false
+                        viewModel.updateSelectedLanguage(lang)
+                    },
+                    modifier = Modifier.animateItemPlacement()
+                )
+                HorizontalDivider(Modifier.padding(start = ListItemStartPadding, end = ListItemStartPadding))
```
frett, line 128: you now have an extra divider at the end of the popup.
Screenshot 2024-08-21 at 1 50 48 PM

Look into using `itemsIndexed` to solve the problem

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
```diff
@@ -0,0 +1,144 @@
...
+        )
+        ExposedDropdownMenuDefaults.TrailingIcon(expanded = expanded)
+        LazyDropdownMenu(
+            expanded = expanded,
+            onDismissRequest = { expanded = false },
+            modifier = Modifier.sizeIn(maxHeight = DROPDOWN_LESSON_MAX_HEIGHT, maxWidth = DROPDOWN_LESSON_MAX_WIDTH)
+        ) {
+            stickyHeader {
+                Surface(color = MaterialTheme.colorScheme.surface) {
+                    val query by viewModel.query.collectAsState("")
```
frett, line 92: https://github.com/user-attachments/assets/b357276e-180f-4d57-bfc7-fe7346dba513

the query is not being reset when a selection is made because the state is being tracked outside of the composition scope (in the ViewModel).

You will need to clear the query either when you open the menu, or after you make a selection.

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,107 @@
...
+        .flatMapLatest { translationsRepository.getTranslationsFlowForTools(it) }
+
+    private val toolCountsFlow = lessonTranslations
+        .map { translations ->
+            translations
+                .groupBy { it.languageCode }
+                .mapValues { it.value.distinctBy { it.toolCode }.count() }
+        }
+
+    private val lessonsWithAllTranslations = lessons.combine(lessonTranslations) { lessons, translations ->
```
frett, line 78: I think the name of the property could be clearer. right now it sounds like you are associating all translations with a lesson. Maybe `lessonTranslationPairs`.

This data structure could also be simplified to just a list of translation objects because the lessonCode is available on the Translation object under `toolCode`

frett: oh, I see what you were doing here, the variable named `translation` confused me since it made it seem that it was dealing with a single `translation`.

frett: this property is only used by `lessonsWithTranslation`, so maybe combine it to a single property that is more straightforward

app/src/test/kotlin/org/cru/godtools/ExternalSingletonsModule.kt
```diff
@@ -86,6 +88,7 @@ class ExternalSingletonsModule {
     val languagesRepository: LanguagesRepository by lazy {
         mockk {
             every { getLanguagesFlow() } returns flowOf(emptyList())
+            every { findLanguageFlow(any()) } returns flowOf(Language(Locale.ENGLISH))
```
frett, line 91: let's default this behavior to `flowOf(null)`, leave it as the responsibility of individual tests using this mock to define expected objects

app/src/main/kotlin/org/cru/godtools/ui/tools/ToolCardLayouts.kt
```diff
@@ -98,17 +98,26 @@ fun PreloadTool(tool: Tool) {
...
     modifier: Modifier = Modifier,
+    selectedLanguage: Language? = null,
     viewModel: ToolViewModels.ToolViewModel = toolViewModels[toolCode],
     onEvent: (ToolCardEvent) -> Unit = {},
 ) {
-    val state = viewModel.toState()
+    val state = viewModel.toState(selectedLanguage = selectedLanguage)
     val tool by viewModel.tool.collectAsState()
-    val translation by viewModel.firstTranslation.collectAsState()
+    val viewModelTranslation by viewModel.firstTranslation.collectAsState()
```
frett, line 107:
```suggestion
    val translation by rememberUpdateState(state.translation)
```

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsLayout.kt
```diff
@@ -29,19 +31,27 @@ internal sealed interface DashboardLessonsEvent {
...
         items(lessons, { it }, { "lesson" }) { lesson ->
             LessonToolCard(
                 lesson,
+                selectedLanguage = selectedLanguage,
                 onEvent = {
                     when (it) {
                         is ToolCardEvent.OpenTool, is ToolCardEvent.Click -> {
                             viewModel.recordOpenLessonInAnalytics(it.tool)
-                            onEvent(DashboardLessonsEvent.OpenLesson(it.tool, it.lang1))
+                            onEvent(DashboardLessonsEvent.OpenLesson(it.tool, selectedLanguage.code))
```
frett, line 52: this should be launching in the language specified by the LessonToolCard, which now will be the selectedLanguage

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,90 @@
...
+    private val context: Context,
+    private val eventBus: EventBus,
+    toolsRepository: ToolsRepository,
+    languagesRepository: LanguagesRepository,
+    private val translationsRepository: TranslationsRepository,
+    private val savedStateHandle: SavedStateHandle,
+    val settings: Settings,
+) : ViewModel() {
+    @OptIn(ExperimentalCoroutinesApi::class)
+    val selectedLanguage = savedStateHandle.getStateFlow(KEY_SAVED_LESSON_LANGUAGE_LOCALE, settings.appLanguage)
```
frett, line 44: you should make this a `StateFlow` by using `.stateIn(viewModelScope, WhileSubscribed(5_000), settings.appLanguage)`

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsLayout.kt
```diff
@@ -29,19 +31,27 @@ internal sealed interface DashboardLessonsEvent {
 @Composable
 @OptIn(ExperimentalFoundationApi::class)
 internal fun LessonsLayout(viewModel: LessonsViewModel = viewModel(), onEvent: (DashboardLessonsEvent) -> Unit = {}) {
-    val lessons by viewModel.lessons.collectAsState()
+    val lessons by viewModel.lessons.collectAsState(emptyList())
+    val selectedLanguage by viewModel.selectedLanguage.collectAsState(Language(Locale.getDefault()))
```
frett, line 35: the `Language(Locale.getDefault())` isn't good since that won't always be the appLanguage on older versions of Android.

If you make `viewModel.selectedLanguage` a `StateFlow` then you won't need to pass a parameter to `.collectAsState()`

frett:
https://github.com/user-attachments/assets/bcc1e4b4-08b9-4999-9c17-db126aa45ae2

Changing the App Language doesn't update the selected language for the lessons UI

app/src/test/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModelTest.kt
```diff
@@ -61,10 +103,70 @@ class LessonsViewModelTest {
         val first = randomTool("first", Tool.Type.LESSON, defaultOrder = 1, isHidden = false)
         val second = randomTool("second", Tool.Type.LESSON, defaultOrder = 2, isHidden = false)
 
+        languageFlow.value = Language(Locale.ENGLISH)
+        lessonsByLanguageFlow.value = listOf(second, first)
         viewModel.lessons.test {
-            lessonsFlow.value = listOf(second, first)
             runCurrent()
+            coVerify { toolsRepository.getLessonsFlowByLanguage(any()) }
```
frett, line 110: when verifying calls to methods you should be explicit with parameters. otherwise this test will still pass if the code is updated to send the wrong language to the method

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsViewModel.kt
```diff
@@ -1,25 +1,91 @@
...
+        .stateIn(viewModelScope, SharingStarted.WhileSubscribed(5_000), settings.appLanguage)
```
frett, line 45: move this after the `flatMapLatest` so that it actually makes `selectedLanguage` a `StateFlow`

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsLayout.kt
```diff
@@ -29,19 +31,27 @@ internal sealed interface DashboardLessonsEvent {
 @Composable
 @OptIn(ExperimentalFoundationApi::class)
 internal fun LessonsLayout(viewModel: LessonsViewModel = viewModel(), onEvent: (DashboardLessonsEvent) -> Unit = {}) {
-    val lessons by viewModel.lessons.collectAsState()
+    val lessons by viewModel.lessons.collectAsState(emptyList())
+    val selectedLanguage by viewModel.selectedLanguage.collectAsState(Language(viewModel.settings.appLanguage))
```
frett, line 35: now that `viewModel.selectedLanguage` is a `StateFlow` you no longer need to pass an initial value to `collectAsState()`

frett (changes requested):
I think it's just the last couple comments

frett (approved):

PR #3735, opened 2024-10-16
GT-2200 add warning error if attempting login offline

gyasistory:


app/src/main/kotlin/org/cru/godtools/ui/login/LoginLayout.kt
```diff
@@ -167,3 +172,31 @@ private fun LoginError(error: LoginResponse.Error?, onDismiss: () -> Unit) {
         )
     }
 }
+
+@Composable
+private fun NoInternetError(createAccount: Boolean, onDismiss: () -> Unit) {
+    val context = LocalContext.current
+    val isConnected by context.isConnectedFlow().collectAsState(true)
```
frett, line 179: this is actually wrong in a subtle manner. every time `NoInternetError` recomposes it will create a new `isConnectedFlow()`. which will cause `collectAsState` to need to start collecting from a new Flow.

You can solve this by wrapping the creation of the flow itself in a `remember {}` block

frett (dismissed):
The jira ticket has updated text and trigger conditions for the error dialog. You need to update the PR to match what is specified there

app/src/main/kotlin/org/cru/godtools/ui/login/LoginLayout.kt
```diff
@@ -145,14 +153,26 @@ fun LoginLayout(createAccount: Boolean = false, onEvent: (event: LoginLayoutEven
 
 @Composable
 private fun LoginError(error: LoginResponse.Error?, onDismiss: () -> Unit) {
+    val context = LocalContext.current
+    val isConnected by remember { context.isConnectedFlow() }.collectAsState(true)
```
frett, line 157: this is business logic in the view layer, you should move detection of if they are offline to the `GodToolsAccountManager` login logic that produces the `LoginResponse.Error`.

What I had said about this on slack last week:

> we could either try and introduce an `AuthenticationException.DeviceOffline` exception we handle, and/or extend the else logic to be something like:
```
else -> when {
    isOffline -> LoginResponse.Error.DeviceOffline
    else -> LoginResponse.Error()
}
```

frett (changes requested):
business logic in the view

library/account/src/main/kotlin/org/cru/godtools/account/AccountModule.kt
```diff
@@ -27,5 +33,14 @@ abstract class AccountModule {
...
         }
+
+        const val IS_CONNECTED_STATE_FLOW_ACCOUNTS = "STATE_FLOW_IS_CONNECTED_ACCOUNTS"
+
+        @Provides
+        @Reusable
+        @Named(IS_CONNECTED_STATE_FLOW_ACCOUNTS)
+        fun isConnectedStateFlow(@ApplicationContext context: Context, scope: CoroutineScope) =
+            context.isConnectedFlow()
+                .stateIn(scope, WhileSubscribed(5_000), false)
```
frett, line 44: why did you duplicate this here instead of just use the `IS_CONNECTED_STATE_FLOW` that is already on the dagger graph?

library/account/src/main/kotlin/org/cru/godtools/account/GodToolsAccountManager.kt
```diff
@@ -75,12 +85,20 @@ class GodToolsAccountManager @VisibleForTesting internal constructor(
...
+                            exceptionOrNull == AuthenticationException.UserNotFound -> {
+                                LoginResponse.Error.UserNotFound
+                            }
+
+                            exceptionOrNull == AuthenticationException.UserAlreadyExists -> {
+                                LoginResponse.Error.UserAlreadyExists
+                            }
+
+                            !isConnectedFlow.value -> LoginResponse.Error.NotConnected
                             else -> LoginResponse.Error()
```
frett, line 102:
```suggestion
                        else -> when (result.exceptionOrNull()) {
                            AuthenticationException.UserNotFound -> LoginResponse.Error.UserNotFound
                            AuthenticationException.UserAlreadyExists -> LoginResponse.Error.UserAlreadyExists
                            !isConnectedFlow.value -> LoginResponse.Error.NotConnected
                            else -> when {
                                isConnectedFlow.value -> LoginResponse.Error.NotConnected
                                else -> LoginResponse.Error()
                            }
```

frett: I'm actually not a fan of using a `StateFlow` like this, by collecting eagerly when the AccountManager is created you will always be monitoring the online status even if the user never touches the login screen.

There is no need to continuously monitor the online state, it should just be a single check to see if the user is online when an error is encountered

app/src/main/res/values/strings_account.xml
```diff
@@ -13,6 +13,9 @@
     <string name="account_error_user_not_found">User account not found.</string>
     <string name="account_error_unknown">Error logging in.</string>
     <string name="account_error_dialog_dismiss">OK</string>
+    <string name="account_error_no_internet">Please check your internet connection and try again.</string>
+    <string name="account_error_no_internet_title">No Internet Connection</string>
```
frett, line 17: let's change the name of these phrases to use `not_connected` to match the `LoginResponse.Error` they are displayed for. That way naming is consistent

library/account/src/main/kotlin/org/cru/godtools/account/AccountModule.kt
gyasistory: @frett Based on my understanding, I would have had to implement the base module into the account module, and I didn't think you would have wanted that.

frett: the account module is already dependent on `:library:base` via the path: `account -> api -> model -> base`.

The symbols are just not present because `model` defines an `implementation` dependency and not an `api` dependency, which is correct because model doesn't need to expose the base dependency because nothing in base is part of the model public API.

Adding an `implementation` dependency from account to `base` would be the right thing here since the `IS_CONNECTED_STATE_FLOW` is part of the public API of the base module, and you want to use it.

The reason this wasn't the case with the `FilterMenuItem` from a previous PR is because that was not meant to be part of the public API that other modules/features can use

frett: the difference between an `implementation` and an `api` dependency is whether the transitive dependency is exposed as part of the public API of the module you are adding the dependency to.

app/src/main/res/values/strings_account.xml
```diff
@@ -13,6 +13,9 @@
     <string name="account_error_user_not_found">User account not found.</string>
     <string name="account_error_unknown">Error logging in.</string>
     <string name="account_error_dialog_dismiss">OK</string>
+    <string name="account_error_not_connected">Please check your internet connection and try again.</string>
+    <string name="account_error_not_connected_title">No Internet Connection</string>
+    <string name="account_error_title">Authorization Failed</string>
```
frett, line 18: make this `Authentication Failed`

frett: authentication is the process of providing credentials to authenticate.

Authorization is a permission issue after you have already authenticated

library/account/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,3 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+</manifest>
```
frett, line 3: This may not be necessary now since you are using the `StateFlow` provided in `:library:base`

library/account/src/main/kotlin/org/cru/godtools/account/GodToolsAccountManager.kt
```diff
@@ -26,20 +28,28 @@ import org.ccci.gto.android.common.base.Ordered
...
     internal constructor(
         providers: Set<@JvmSuppressWildcards AccountProvider>,
         userApi: Lazy<UserApi>,
-    ) : this(providers.sortedWith(Ordered.COMPARATOR), userApi)
+        @Named(IS_CONNECTED_STATE_FLOW)
+        isConnected: StateFlow<Boolean>,
+        coroutineScope: CoroutineScope = CoroutineScope(SupervisorJob()),
+    ) : this(providers.sortedWith(Ordered.COMPARATOR), userApi, isConnected)
+
+    private val isConnectedFlow = isConnected.stateIn(coroutineScope, SharingStarted.Eagerly, false)
```
frett, line 52: is there a reason you are recreating the `StateFlow` you are already wiring into the `GodToolsAccountManager`?

frett: I'll restate my previous review comment:

> I'm actually not a fan of using a StateFlow like this, by collecting eagerly when the AccountManager is created you will always be monitoring the online status even if the user never touches the login screen.
>
> There is no need to continuously monitor the online state, it should just be a single check to see if the user is online when an error is encountered

gyasistory: Sorry pulled in too many directions and overlooked that part.  My thought on doing this was to place the flow within the same coroutine scope.  Please correct me if I have the wrong understanding of this.

frett: @gyasistory let's talk through this at some point today, there's a few things going on here that would be easier to talk through vs type out

library/account/src/main/kotlin/org/cru/godtools/account/GodToolsAccountManager.kt
```diff
@@ -39,6 +46,7 @@ class GodToolsAccountManager @VisibleForTesting internal constructor(
     internal constructor(
         providers: Set<@JvmSuppressWildcards AccountProvider>,
         userApi: Lazy<UserApi>,
+        coroutineScope: CoroutineScope = CoroutineScope(SupervisorJob()),
```
frett, line 49: this isn't used since we aren't creating a `StateFlow` anymore

library/account/build.gradle.kts
```diff
@@ -12,12 +12,14 @@ android {
 
 dependencies {
     implementation(project(":library:api"))
+    implementation(project(":library:base"))
```
frett, line 15: this isn't needed anymore since we aren't using the `isConnectedStateFlow`

library/account/src/main/kotlin/org/cru/godtools/account/GodToolsAccountManager.kt
```diff
@@ -117,3 +130,14 @@ class GodToolsAccountManager @VisibleForTesting internal constructor(
     internal suspend fun authenticateWithMobileContentApi() = activeProvider?.authenticateWithMobileContentApi(false)
         ?: Result.failure(AuthenticationException.NoActiveProvider)
 }
+
+// TODO Remove when GTO-Support-Util is updated
```
frett, line 134: you should be able to already use this from gto-support since the branch is using gto-support 4.2.3-SNAPSHOT

library/account/src/main/kotlin/org/cru/godtools/account/GodToolsAccountManager.kt
```diff
@@ -73,15 +75,21 @@ class GodToolsAccountManager @VisibleForTesting internal constructor(
         createAccount: Boolean,
         onResponse: (LoginResponse) -> Unit,
     ): ActivityResultLauncher<AccountType> {
+        val context = LocalContext.current
```
frett, line 78: Instead of using Local context, inject the application context into the account manager Singleton.

frett (changes requested):
Almost there, just the 1 more change I think

library/account/src/main/kotlin/org/cru/godtools/account/GodToolsAccountManager.kt
```diff
@@ -33,13 +37,17 @@ class GodToolsAccountManager @VisibleForTesting internal constructor(
     @get:VisibleForTesting
     internal val providers: List<AccountProvider>,
     private val userApi: Lazy<UserApi>,
+    @ApplicationContext
```
frett, line 40: since this isn't the constructor yo are injecting, you don't need to specify the `@ApplicationContext` annotation on it

frett (approved):

PR #3977, opened 2025-04-30
Feature/opt in notification

JasonBennett41:
The Opt In Notification feature is designed as a submodule of the dashboard.  The OptInNotificationController is used to isolate logic specific to the feature as much as can be.

In the controller, shouldPromptNotificationSheet() checks that:

1. The user has completed onboarding
2. The user has not already granted notification permissions
3. The user has not been prompted recently
4. The user has not exceeded the maximum amount of prompts
5. The feature is enabled via FirebaseRemoteConfig
6. The user is running the app with an sdk version of at least 33*

*For users running < 33, notification permissions are automatically granted

```
fun shouldPromptNotificationSheet() {

        val lastPrompted = settings.getLastPromptedOptInNotification() ?: Date(Long.MIN_VALUE)
        val promptCount = settings.getOptInNotificationPromptCount()

        val remoteTimeIntervalLong = remoteConfig.getLong(CONFIG_UI_OPT_IN_NOTIFICATION_TIME_INTERVAL)
        val calendar = Calendar.getInstance()
        calendar.add(Calendar.DAY_OF_YEAR, -remoteTimeIntervalLong.toInt())
        val remoteTimeInterval = calendar.time

        val remotePromptLimit = remoteConfig.getLong(CONFIG_UI_OPT_IN_NOTIFICATION_PROMPT_LIMIT).toInt()

        val remoteFeatureEnabled = remoteConfig.getBoolean(CONFIG_UI_OPT_IN_NOTIFICATION_ENABLED)

        // TODO: Remove sdk version checks for optInNotification logic once minSdk = 33 or greater
        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU || !remoteFeatureEnabled || isOnboardingLaunch || viewModel.permissionStatus == PermissionStatus.APPROVED || promptCount > remotePromptLimit) {
            return
        }

        if (lastPrompted < remoteTimeInterval) {
            viewModel.setShowOptInNotification(true)
            settings.recordOptInNotificationPrompt()
        }
    }
```
If a user meets all the prerequisites, they will encounter the Opt In Notification sheet. The sheet itself responds to user input by showing the system permission request dialog, or sending a user directly to system app settings based on their previous interactions with the android notification permission dialog. The prompt is recorded by storing 'lastPromptedOptInNotification' and 'optInNotificationPromptCount' in shared prefs for use in determining the need for future prompts.

App/util/DeviceUtils.kt is also added to assist with adaptive sizing:

```
@OptIn(ExperimentalMaterial3WindowSizeClassApi::class)
@Composable
private fun detectTablet(): Boolean {
    val context = LocalContext.current
    val activity = context as? Activity ?: return false

    val windowSizeClass = calculateWindowSizeClass(activity = activity)

    val detectedTablet = windowSizeClass.widthSizeClass == WindowWidthSizeClass.Medium ||
            windowSizeClass.widthSizeClass == WindowWidthSizeClass.Expanded

    return detectedTablet
}
```


library/base/src/main/kotlin/org/cru/godtools/base/Settings.kt
```diff
@@ -157,6 +167,40 @@ class Settings internal constructor(private val context: Context, coroutineScope
...
+    }
+
+    fun recordOptInNotificationPrompt() {
+        val currentPromptCount = getOptInNotificationPromptCount()
+        val updatedPromptCount = currentPromptCount + 1
+
+        val dateString = dateFormat.format(Date())
+
+        prefs.edit {
+            putString(LAST_PROMPTED_OPT_IN_NOTIFICATION, dateString)
```
frett, line 198: Let's store the last prompted timestamp as the time millis, that way you don't have to format & parse the date when loading it

app/src/main/kotlin/org/cru/godtools/ui/dashboard/optinnotification/OptInNotificationModalOverlay.kt
```diff
@@ -0,0 +1,230 @@
...
+        LaunchedEffect(transitionState.currentState, transitionState.targetState) {
+            if (!transitionState.currentState && transitionState.isIdle) {
+                navigator.finish(Unit)
+            }
+        }
+
+        Surface(
+            color = Color.Transparent, modifier = Modifier
+                .fillMaxSize()
+                .background(Color.Black.copy(alpha = 0.2f))
```
frett, line 82:
```suggestion
                .background(MaterialTheme.colorScheme.scrim.copy(alpha = 0.2f))
```

app/src/main/kotlin/org/cru/godtools/ui/dashboard/DashboardViewModel.kt
```diff
@@ -61,9 +58,25 @@ class DashboardViewModel @Inject constructor(
...
+    val showOptInNotification: StateFlow<Boolean> = _showOptInNotification
+
+    fun setShowOptInNotification(bool: Boolean) {
+        _showOptInNotification.value = bool
+    }
+    // endregion optInNotification logic
 
     init {
         triggerSync()
     }
```
frett, line 81: let's leave this init block as part of the Sync logic block

frett:
@JasonBennett41 there are some ktlint code style issues to address

gradle/libs.versions.toml
```diff
@@ -65,9 +66,11 @@ androidx-datastore-preferences = { module = "androidx.datastore:datastore-prefer
 androidx-fragment-ktx = "androidx.fragment:fragment-ktx:1.8.6"
 androidx-hilt-compiler = { module = "androidx.hilt:hilt-compiler", version.ref = "androidx-hilt" }
 androidx-hilt-work = { module = "androidx.hilt:hilt-work", version.ref = "androidx-hilt" }
+androidx-hilt-navigation-compose = { module = "androidx.hilt:hilt-navigation-compose", version.ref = "androidx-hilt" }
 androidx-lifecycle-livedata-ktx = { module = "androidx.lifecycle:lifecycle-livedata-ktx", version.ref = "androidx-lifecycle" }
 androidx-lifecycle-runtime-testing = { module = "androidx.lifecycle:lifecycle-runtime-testing", version.ref = "androidx-lifecycle" }
 androidx-lifecycle-viewmodel-compose = { module = "androidx.lifecycle:lifecycle-viewmodel-compose", version.ref = "androidx-lifecycle" }
+androidx-material3-window-size-class1-android = { group = "androidx.compose.material3", name = "material3-window-size-class-android", version.ref = "material3WindowSizeClassAndroid" }
```
frett, line 73: I'd like to share the version ref with the `androidx-compose-material3` dependency already defined.

Also, I haven't checked, but this isn't included transitively when the compose material3 dependency is imported?

gradle/libs.versions.toml
```diff
@@ -65,9 +66,11 @@ androidx-datastore-preferences = { module = "androidx.datastore:datastore-prefer
 androidx-fragment-ktx = "androidx.fragment:fragment-ktx:1.8.6"
 androidx-hilt-compiler = { module = "androidx.hilt:hilt-compiler", version.ref = "androidx-hilt" }
 androidx-hilt-work = { module = "androidx.hilt:hilt-work", version.ref = "androidx-hilt" }
+androidx-hilt-navigation-compose = { module = "androidx.hilt:hilt-navigation-compose", version.ref = "androidx-hilt" }
```
frett, line 69: I'm trying to see where this dependency is being used, we aren't using JetPack Navigation anywhere else in the godtools app

library/base/src/main/kotlin/org/cru/godtools/base/Settings.kt
frett: My own personal preference is to use the java8 time APIs instead of the older `Date` class. They are much more full featured.

Since you are storing just a date, there is a `LocalDate` object that can be used, it provides `LocalDate.toEpochDay()` and `LocalDate.ofEpochDay()` which can be used for storing the value in preferences. It also supports comparison by doing something like: `storedDate.plusDays(5) < LocalDate.now()`

It looks like we have already enabled android desugaring for this library so we would have access to those APIs without any issues

app/src/main/res/values/strings_opt_in_notification.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="opt_in_notification_title">Get Tips and Encouragement</string>
+    <string name="opt_in_notification_body">Stay equipped for conversations.\nAllow notifications today.</string>
+    <string name="opt_in_notification_allow_notifications">Allow Notifications</string>
+    <string name="opt_in_notification_notification_settings">Notification Settings</string>
+    <string name="opt_in_notification_maybe_later">Maybe Later</string>
```
frett, line 7: let's just add this strings to the `strings_dashboard.xml` file under a new "section"

JasonBennett41:
@frett Reviewing the issues/edit requests now 👍🏽

app/src/main/kotlin/org/cru/godtools/ui/dashboard/DashboardLayout.kt
```diff
@@ -87,6 +94,33 @@ internal fun DashboardLayout(onEvent: (DashboardEvent) -> Unit, viewModel: Dashb
     val snackbarHostState = remember { SnackbarHostState() }
     AppUpdateSnackbar(snackbarHostState)
 
+    // region optInNotification
+    val showOverlay by viewModel.showOptInNotification.collectAsState()
+
+    val context = LocalContext.current
+    val activity = context as? Activity
```
frett, line 101: there is a `LocalActivity.current` you can use here instead of casting `context` to an Activity

app/src/main/kotlin/org/cru/godtools/util/DeviceUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+    val detectedTablet = windowSizeClass.widthSizeClass == WindowWidthSizeClass.Medium ||
+            windowSizeClass.widthSizeClass == WindowWidthSizeClass.Expanded
+
+    return detectedTablet
+}
+
+@Composable
+fun isTablet(): Boolean {
+    val hasRun = remember { mutableStateOf(false) }
+    val isTablet = remember { mutableStateOf(false) }
```
frett, line 29: by storing the result of `detectedTablet` in a `remember { mutableStateOf() }` it won't be updated if the UI layout changes, for example if a user unfolds a foldable device.

Comopse takes care of intelligently re-composing functions for you when configuration/state changes, so there is no need to try and manage this yourself

frett: The entirety of this logic could be simplified to:

```
private val TABLET_SIZE_CLASSES = setOf(
    WindowWidthSizeClass.Medium,
    WindowWidthSizeClass.Expanded
)

@OptIn(ExperimentalMaterial3WindowSizeClassApi::class)
@Composable
private fun isTablet(): Boolean {
    val activity = LocalActivity.current ?: return false
    return calculateWindowSizeClass(activity).widthSizeClass in TABLET_SIZE_CLASSES
}
```

frett: the `calculateWindowSizeClass()` composable should already be intelligent about when it gets recomposed and actually re-evaluates the test of if this is a tablet or not.

frett: likewise, the `isTablet()` function will only re-evaluate when some of the state it reads changes (`LocalActivity` changes, or `calculateWindowSizeClass()` needs to recompose because some of the state it's dependent on changes)

gradle/libs.versions.toml
frett: also, you don't need the `-android` part of the library name, the material3 compose libraries are Kotlin Multiplatform modules. Using `-android` just bypasses the multiplatform resolution to get the android platform version only. It's valid, but could be potentially confusing because everywhere that google references the library they don't include the platform modifier.

frett: and name the version catalog entry: `androidx-compose-material3-windowsize`

frett:
```suggestion
androidx-compose-material3-windowsizeclass = { group = "androidx.compose.material3", name = "material3-window-size-class", version.ref = "compose-material3" }
```

app/src/main/kotlin/org/cru/godtools/util/DeviceUtils.kt
frett: also this intelligent recomposition only works because `calculateWindowSizeClass()` is a Composable function that properly uses state to start with. if it was a regular function you would want to use some sort of `remember {}` like you were doing to prevent re-evaluating it every recomposition

app/src/main/kotlin/org/cru/godtools/ui/dashboard/optinnotification/OptInNotificationModalOverlay.kt
```diff
@@ -0,0 +1,230 @@
...
+                        durationMillis = 350,
+                    ),
+                ),
+                exit = slideOutVertically(
+                    targetOffsetY = { it },
+                    animationSpec = tween(
+                        durationMillis = 350,
+                    ),
+                ),
+            ) {
```
frett, line 98: These transitions should be able to be converted to an `enterTransition` and `exitTransition` that you gain from extending `AnimatedOverlay` instead of just `Overlay`.

Then the circuit framework should take care of managing the transition for you and not removing the overlay until the animation finishes

app/src/main/kotlin/org/cru/godtools/ui/dashboard/optinnotification/OptInNotificationModalOverlay.kt
```diff
@@ -0,0 +1,230 @@
...
+                BoxWithConstraints {
+                    val totalWidth = constraints.maxWidth.dp
+                    val isLargeTablet = totalWidth > 1700.dp
+
+                    Card(
+                        modifier = Modifier
+                            .padding(horizontal = if (isTablet) if (isLargeTablet) 200.dp else 150.dp else 16.dp)
```
frett, line 105: this is confusing, let's convert this to a when statement to be more clear.

```suggestion
                            .padding(
                                horizontal = when {
                                    isLargeTablet -> 200.dp
                                    isTablet -> 150.dp
                                    else -> 16.dp
                                }
                            )
```

app/src/main/kotlin/org/cru/godtools/util/DeviceUtils.kt
JasonBennett41: I see, thanks for the insight here

app/src/main/kotlin/org/cru/godtools/ui/dashboard/optinnotification/OptInNotificationModalOverlay.kt
JasonBennett41: So I attempted this approach at first, but the issue is that the dimmed background animates along with the rest of the content.  I could not find a workaround while still keeping the background color within the overlay itself.

Would it be preferable to add the background color in DashboardLayout instead in order to simplify overlay animation logic?  Or if you have any other potential solution let me know.

https://github.com/user-attachments/assets/9f6c662d-9234-4213-9979-940137241a4e

gradle/libs.versions.toml
JasonBennett41: Yes I don't think it's being used at all, not sure how it got thrown in there.  Removed and building successfully, so will nix it

JasonBennett41:
I believe I performed all requested changes, with the exception of changing Overlay to AnimatedOverlay.  Will wait for your response in regards to that one.  Let me know if there is anything else you need. Thanks!

app/src/main/kotlin/org/cru/godtools/ui/dashboard/optinnotification/OptInNotificationModalOverlay.kt
frett: Ah, leave it as it is then

frett: it looks like there is an `Modifier.animateEnterExit()` modifier available in the `AnimatedVisibiltyScope` that would allow you do do a custom animation for the scrim only

frett: the modifier looks like it adds to the already defined animation. their documentation suggests doing `EnterTransition.None`/`ExitTransition.None` and then define the different animations of components when you need varying animations

app/src/main/kotlin/org/cru/godtools/ui/dashboard/optinnotification/OptInNotificationModalOverlay.kt
```diff
@@ -0,0 +1,245 @@
...
+                                }
+                            )
+                            .align(Alignment.BottomCenter)
+                            .offset(y = 4.dp)
+                            .fillMaxWidth()
+                            .wrapContentHeight(),
+                        shape = RoundedCornerShape(16.dp),
+                        border = BorderStroke(4.5.dp, MaterialTheme.colorScheme.primary),
+                        colors = CardDefaults.cardColors(
+                            containerColor = Color.White
```
frett, line 118: we are low-key working on supporting dark mode (currently supported on debug & qa builds, but not production). We don't have explicit designs for what dark mode should look like for this UI, but we do have the freedom to stray from the normal design a bit to better support dark mode as well.

By hardcoding this to white it will have issues when your device is in dark mode, this should be defined as either `MaterialTheme.colorScheme.background` or `MaterialTheme.colorScheme.container` (I think this is the standard container surface color). You will want to update the large image to have a transparent background instead of having the white background baked in.

you can see how we handle a similar image with dark mode in the Delete Account UI. Here are the delete account screenshot tests showing dark vs light themes:

https://github.com/CruGlobal/godtools-android/blob/develop/app/src/test/snapshots/images/org.cru.godtools.ui.account.delete_DeleteAccountLayoutPaparazziTest_DeleteAccountLayout()_-_Deleting%5BPixel_6_Pro%2CNOTNIGHT%2CNO_ACCESSIBILITY%5D.png
https://github.com/CruGlobal/godtools-android/blob/develop/app/src/test/snapshots/images/org.cru.godtools.ui.account.delete_DeleteAccountLayoutPaparazziTest_DeleteAccountLayout()_-_Display%5BNexus_5%2CNIGHT%2CNO_ACCESSIBILITY%5D.png

app/src/main/kotlin/org/cru/godtools/ui/dashboard/DashboardActivity.kt
```diff
@@ -38,21 +53,46 @@ class DashboardActivity : BaseActivity() {
...
+                                        *listOfNotNull(e.lang1, e.lang2).toTypedArray()
+                                    )
+
+                                    is DashboardEvent.OpenToolDetails -> e.tool?.let {
+                                        startToolDetailsActivity(
+                                            it,
+                                            e.lang
+                                        )
+                                    }
+                                }
```
frett, line 92: just a minor nit, the auto-formatter in Android Studio sometimes makes code less readable by introducing a ton of line returns when making a line not exceed the max line length

app/src/main/kotlin/org/cru/godtools/ui/dashboard/DashboardActivity.kt
```diff
@@ -77,10 +118,16 @@ class DashboardActivity : BaseActivity() {
...
-                        viewModel.updateCurrentPage(findPageByUriPathSegment(data.pathSegments.getOrNull(2)))
+                    data?.isDashboardCustomUriSchemeDeepLink() == true -> viewModel.updateCurrentPage(
+                        findPageByUriPathSegment(data.pathSegments.getOrNull(1))
+                    )
+
+                    data?.isDashboardGodToolsDeepLink() == true -> viewModel.updateCurrentPage(
+                        findPageByUriPathSegment(
+                            data.pathSegments.getOrNull(2)
+                        )
+                    )
```
frett, line 129: more auto-formatter shenanigans, I don't think this logic changed at all and was perfectly fine with ktlint rules defined

app/src/main/kotlin/org/cru/godtools/ui/dashboard/DashboardLayout.kt
```diff
@@ -87,6 +97,32 @@ internal fun DashboardLayout(onEvent: (DashboardEvent) -> Unit, viewModel: Dashb
...
+
+    val activity = LocalActivity.current
+
+    LaunchedEffect(showOverlay) {
+        activity?.requestedOrientation = if (showOverlay) {
+            ActivityInfo.SCREEN_ORIENTATION_PORTRAIT
+        } else {
+            ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED
+        }
+    }
```
frett, line 111: like we talked about during our call, this can be removed

app/src/main/kotlin/org/cru/godtools/ui/dashboard/DashboardViewModel.kt
```diff
@@ -8,30 +8,29 @@ import javax.inject.Inject
...
         set(value) = savedState.set(KEY_PAGE_STACK, ArrayList(value))
     private val pageStackFlow = savedState.getStateFlow(KEY_PAGE_STACK, listOf(Page.DEFAULT))
 
-    val hasBackStack = pageStackFlow
-        .map { it.size > 1 }
-        .stateIn(viewModelScope, SharingStarted.Eagerly, false)
-    val currentPage = pageStackFlow
-        .map { it.lastOrNull() ?: Page.DEFAULT }
+    val hasBackStack = pageStackFlow.map { it.size > 1 }.stateIn(viewModelScope, SharingStarted.Eagerly, false)
+    val currentPage = pageStackFlow.map { it.lastOrNull() ?: Page.DEFAULT }
```
frett, line 33: more auto-formatter shenanigans

app/src/main/kotlin/org/cru/godtools/ui/dashboard/DashboardViewModel.kt
```diff
@@ -66,4 +65,20 @@ class DashboardViewModel @Inject constructor(
...
+    // region optInNotification logic
+    var permissionStatus: PermissionStatus? = null
+        private set
+
+    fun setPermissionStatus(status: PermissionStatus) {
+        permissionStatus = status
+    }
+
+    private val _showOptInNotification = MutableStateFlow(false)
+    val showOptInNotification: StateFlow<Boolean> = _showOptInNotification
```
frett, line 78: if you truly want to protect the `StateFlow` from being set elsewhere you can use `_showOptInNotification.asStateFlow()`.

I'm also perfectly fine exposing this property directly as a `MutableStateFlow<Boolean>` and setting it directly instead of going through `setShowOptInNotification()`

app/src/main/kotlin/org/cru/godtools/ui/dashboard/DashboardActivity.kt
JasonBennett41: Ah shoot, I ran ./gradlew ktlintFormat which performed these changes automatically.  Will try to revert and make sure it is still passing ktlintCheck

app/src/main/kotlin/org/cru/godtools/util/DeviceUtils.kt
```diff
@@ -0,0 +1,46 @@
...
+//    val hasRun = remember { mutableStateOf(false) }
+//    val isTablet = remember { mutableStateOf(false) }
+//
+//    if (!hasRun.value) {
+//        val detectedTablet = detectTablet()
+//        isTablet.value = detectedTablet
+//        hasRun.value = true
+//    }
+//    return isTablet.value
+// }
```
frett, line 46: this commented code can be deleted

library/base/src/main/kotlin/org/cru/godtools/base/Settings.kt
```diff
@@ -157,6 +165,32 @@ class Settings internal constructor(private val context: Context, coroutineScope
...
+        val lastPrompted = LocalDate.ofEpochDay(prefs.getLong(LAST_PROMPTED_OPT_IN_NOTIFICATION, 0))
+
+        return lastPrompted
+    }
+
+    fun getOptInNotificationPromptCount(): Int {
+        val promptCount = prefs.getInt(OPT_IN_NOTIFICATION_PROMPT_COUNT, 0)
+
+        return promptCount
+    }
```
frett, line 179: so a fun thing for simple functions like this in Kotlin is that it can be reduced to an expression function like this:

```suggestion
    fun getOptInNotificationPromptCount() = prefs.getInt(OPT_IN_NOTIFICATION_PROMPT_COUNT, 0)
```

frett: `getLastPromptedOptInNotification()` can be simplified in a similar manner

frett (changes requested):
@JasonBennett41 finished up reviewing the code, overall everything looks good, just some outstanding comments for how it could be more succinct or how to better support other functionality in the app

app/src/main/kotlin/org/cru/godtools/ui/dashboard/DashboardActivity.kt
frett: this one probably got caught by the autoformatter since you indented this code when adding the `ContentWithOverlays` block around it

JasonBennett41:
> @JasonBennett41 finished up reviewing the code, overall everything looks good, just some outstanding comments for how it could be more succinct or how to better support other functionality in the app

Thanks very much for the feedback, I certainly appreciate it.  Will reconcile the remaining items and let you know when they are complete 👍🏽

app/src/main/kotlin/org/cru/godtools/ui/dashboard/optinnotification/OptInNotificationModalOverlay.kt
frett: I'm assuming you exported the banner image from figma directly, it's kind of a pain to get it to export without baking in the background color.

What I've done in the past to do this is to turn off the background layers before exporting. I also use the slice tool to set a fixed padding around the image instead of letting figma determine what padding or lack of padding it wants to include in the export

library/base/src/main/kotlin/org/cru/godtools/base/Settings.kt
JasonBennett41: Interesting, and no performance loss without the explicit type definition?

frett: nope, the compiler infers the type from the expression. you can also still specify the type if you need to.

```kotlin
fun getOptInNotificationPromptCount(): Int = prefs.getInt(OPT_IN_NOTIFICATION_PROMPT_COUNT, 0)
```

frett: https://kotlinlang.org/docs/functions.html#single-expression-functions

app/src/main/kotlin/org/cru/godtools/ui/dashboard/optinnotification/OptInNotificationModalOverlay.kt
JasonBennett41: Gotcha sorry I missed that documentation. Was able to get this approach to work, so will remove all the manual transition logic.  Thanks for the help

app/src/main/kotlin/org/cru/godtools/ui/dashboard/DashboardActivity.kt
JasonBennett41: Ah yes I see, so yes if I try to return to previous formatting (or use two lines instead of one for each argument) it does produce an 'each argument should be on a separate line (unless all arguments can fit a single line)' failure.  Have to leave it as is, no?

app/src/main/kotlin/org/cru/godtools/ui/dashboard/optinnotification/OptInNotificationModalOverlay.kt
JasonBennett41: Yes your assumption is correct.  I believe I will need edit access to Figma file for that, should I send a message in Slack channel for that?

frett: I just cloned the figma project to mess with the layers, then deleted the copy after I was done

JasonBennett41: Okay, I went ahead and cloned as you suggested.  However, the background seems to be 'baked' into the image itself, rather than included as a layer...  I don't think I am able to remove the background for this reason, see below


https://github.com/user-attachments/assets/3486d14f-fcb9-4500-9e2a-bca55fa0cba4

frett: hmm, that's annoying

aaronlaib: @frett @JasonBennett41 I'll try to get an updated image from Justin or Lily. I'm not sure where the original file was created, but will track it down.

aaronlaib: @frett @JasonBennett41 was able to track it down. Will this work? (This is a larger image, let me know if you need it reduced in size)

opt-in-image-girl-transparent

JasonBennett41: @aaronlaib @frett thank you both for your help retrieving the graphic, will get that integrated today 👍🏽

JasonBennett41:
@frett okay, with the substitution of the new graphic the requested changes should be complete.  Let me know if there's anything left to attend to here.  Thanks!

frett (approved):

frett:
I want to work through a bit of a build/branch refactor for GodTools before I merge this, I should hopefully be done by the end of today and can merge it then

PR #4062, opened 2025-07-11
Switch to Crowdin for translation management

frett:


crowdin.yml
```diff
@@ -0,0 +1,41 @@
+"project_id": "805338"
```
frett, line 1: @levieggertcru here's what the crowdin CLI config looks like, I think we will just end up using this to sync translations for android, iOS, & KMP projects

frett: I do still need to setup the GHA to push/pull translations using the CLI

dr-bizz: Thank you for tagging us.

PR #4198, opened 2025-11-03
GT-2326 Pass the second language when opening tips from tool details

tjohnson009:
This PR aims to fix what caused the QA failure in the original GT-2326 ticket. The language selector in the top bar should now show for when two languages are selected in the app.


tjohnson009:
@frett I think I need to rebase and squash those two commits together? Is that right?

frett:
@tjohnson009 code looks good, go ahead and squash the 2 commits together. I also updated the PR title to be more descriptive of the actual change instead of just being "QA Fail Fix"

tjohnson009:
It looks like one of the commits went away entirely but the content is the same.... I tried to squash them together but I must have done something wrong. Let me know if it's okay to push how it is. All of the same code changes remained.

frett (approved):
looks good

PR #4223, opened 2025-11-18
GT-2809 Create Custom Sharing Action for Sharesheet

tjohnson009:


ui/base-tool/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ dependencies {
     api(project(":ui:base"))
     implementation(project(":library:base"))
     implementation(project(":library:model"))
+    implementation(project(":ui:qr-code"))
```
frett: I'm assuming you put the dependency here to be able to create the qr code intent, there is a different way I'll want to handle this. I can walk through it with you at some point, but there are a few other changes that we'll want to make first :)

gradle/libs.versions.toml
```diff
@@ -36,6 +36,9 @@ play-app-update = "2.1.0"
 retrofit2 = "3.0.0"
 scarlet = "0.1.12"
 splitties = "3.0.0"
+agp = "8.13.0"
+lifecycleRuntimeKtx = "2.9.4"
+composeBom = "2024.09.00"
```
frett: I know a lot of this was auto-generated by Android Studio when you created the project, this is a lot of duplicated dependencies that we already have defined in a different manner

settings.gradle.kts
```diff
@@ -75,6 +75,7 @@ include("ui:article-aem-renderer")
 include("ui:article-renderer")
 include("ui:base")
 include("ui:base-tool")
+include("ui:qr-code")
```
frett: minor nit, alpha sort the ui module to make it easy to find in the future

ui/base-tool/src/main/kotlin/org/cru/godtools/base/tool/activity/BaseToolActivity.kt
```diff
@@ -214,19 +221,21 @@ abstract class BaseToolActivity<B : ViewDataBinding>(@LayoutRes contentLayoutId:
...
     ) = shareUrl?.let {
         Intent(Intent.ACTION_SEND).apply {
             type = "text/plain"
             putExtra(Intent.EXTRA_SUBJECT, title.orEmpty())
             putExtra(Intent.EXTRA_TEXT, getString(message, shareUrl))
+            putExtra("EXTRA_SHARE_TOOL_URL", shareUrl)
         }
     }
 
+    @RequiresApi(Build.VERSION_CODES.UPSIDE_DOWN_CAKE)
```
frett: putting the `@RequiresApi` here is probably not the right approach, we want the `showShareActivityChooser()` method to be called from all versions of Android, it's just the qr code quick action that isn't available until `UPSIDE_DOWN_CAKE`

ui/base-tool/src/main/kotlin/org/cru/godtools/base/tool/activity/BaseToolActivity.kt
```diff
@@ -171,8 +177,9 @@ abstract class BaseToolActivity<B : ViewDataBinding>(@LayoutRes contentLayoutId:
     protected open val shareMenuItemVisible by lazy { shareLinkUriLiveData.map { it != null } }
 
     protected open val shareLinkTitle get() = activeManifest?.title
+
     @get:StringRes
-    protected open val shareLinkMessageRes get() = R.string.share_tool_message
+    protected open val shareLinkMessageRes get() = R.string.share_tool_message_with_qr_code
```
frett: I think this is the only actual change in this file? everything else looks like auto-formatting changes

ui/base-tool/src/main/kotlin/org/cru/godtools/base/tool/ui/share/model/DefaultShareItem.kt
```diff
@@ -1,15 +1,49 @@
...
     override fun triggerAction(activity: Activity) {
-        activity.startActivity(Intent.createChooser(shareIntent, null))
+        val chooserIntent = Intent.createChooser(shareIntent, null)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+            val options = ActivityOptions.makeBasic().apply {
+                setPendingIntentCreatorBackgroundActivityStartMode(
+                    ActivityOptions.MODE_BACKGROUND_ACTIVITY_START_ALLOWED
+                )
+            }
+            val urlStringForQrCode = shareIntent.getStringExtra("EXTRA_SHARE_TOOL_URL")
```
frett: instead of bundling the share tool url in the intent and extracting it here, provide it as an optional constructor property when creating the `DefaultShareItem` object

frett: it provides a cleaner way of passing the additional option to the `DefaultShareItem`

ui/base-tool/src/main/res/values/strings_tool_renderer.xml
```diff
@@ -4,6 +4,17 @@
...
+    <string name="share_tool_message_with_qr_code">
+    \n\n
+    This QR Code will take you to the web version of what we were talking about today. \n\n
+    If you can\'t scan the QR code, here\'s a website link: %1$s \n\n
+
+    The app can be downloaded from https://godtoolsapp.com \n\n
+
+    Let’s talk more!
+    </string>
+
```
frett: I don't think we actually want to add this string, the message is what is sent via text messages/chat app to another user. it will reference a QR code that they won't see

ui/base-tool/src/main/kotlin/org/cru/godtools/base/tool/activity/BaseToolActivity.kt
frett: after seeing the string you added, I think we actually want to leave this as the original message. I left comments as to why there

ui/qr-code/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+    buildFeatures {
+        compose = true
+    }
+}
+
+dependencies {
+    implementation(libs.androidx.core.ktx)
+    implementation(libs.androidx.lifecycle.runtime.ktx)
+    implementation(libs.androidx.activity.compose)
+    implementation(platform(libs.androidx.compose.bom))
```
frett: we actually are not using the compose bom, we currently use renovate to keep our compose dependencies up to date.

ui/qr-code/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+    implementation(libs.androidx.compose.ui)
+    implementation(libs.androidx.compose.ui.graphics)
+    implementation(libs.androidx.compose.ui.tooling.preview)
+    implementation(libs.androidx.compose.material3)
+
+    implementation("com.google.zxing:core:3.5.1")
```
frett: this dependency needs to be moved to the dependency library (`libs.versions.toml`)

ui/qr-code/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,24 @@
...
+    <application
+        android:allowBackup="false"
+        android:icon="@mipmap/ic_launcher"
+        android:label="@string/app_name"
+        android:roundIcon="@mipmap/ic_launcher_round"
+        android:supportsRtl="true"
+        android:theme="@style/Theme.GodTools">
+        <activity
+            android:name=".activity.QRCodeActivity"
+            android:exported="true"
```
frett: I'm curious if this will work with `exported="false"`.

the exported property indicates if an activity is visible to other apps on the phone or not, I believe you can still target the activity with exported = false from pending intents you create.

ui/qr-code/src/main/kotlin/org.cru.godtools.qrcode/activity/QRCodeActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+        putExtra("EXTRA_SHARE_URL", shareUrl)
+    }
+)
+
+class QRCodeActivity : ComponentActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        enableEdgeToEdge()
+
+        val url = intent.getStringExtra("EXTRA_SHARE_URL") ?: "https://godtoolsapp.com/tools"
```
frett:
```suggestion
        val url = intent.getStringExtra("EXTRA_SHARE_URL")
        if (url == null) {
            finish()
            return
        }
```

frett: since the purpose of this activity is to display a QR code for a provided url, if we don't open the activity correctly let's just finish the activity and return the user to where they came from.

ui/qr-code/src/main/kotlin/org.cru.godtools.qrcode/activity/QRCodeActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+                            }
+                    )
+                }
+            }
+        }
+    }
+}
+
+@Composable
+fun QRCodeScreen(url: String, modifier: Modifier) {
```
frett:
```suggestion
internal fun QRCodeScreen(url: String, modifier: Modifier = Modifier) {
```

frett: 2 changes:
1. make this internal since there should be no need to call this from outside of the `qr-code` module, you could probably make this `private` since it's in the same source file you are calling it from currently
2. Stylistically providing a default `Modifier` makes it so that you aren't required to pass a Modifier if you don't want to change how the composable is rendered

ui/qr-code/src/main/kotlin/org.cru.godtools.qrcode/activity/QRCodeActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+        val bitmap = createBitmap(qrCodeWidth, qrCodeHeight, Bitmap.Config.RGB_565)
+        for (i in 0 until qrCodeWidth) {
+            for (j in 0 until qrCodeHeight) {
+                val color = if (bitMatrix.get(i, j)) -0x1000000 else -0x1 // needs Int?
+                bitmap.setPixel(i, j, color)
+            }
+        }
+        return bitmap
+    } catch (e: Exception) {
+        println(e)
```
frett: `println` shouldn't be used in actual code since it doesn't get printed anywhere. We are using the `Timber` library for logging, you will want to add the timber dependency to build.gradle.kts `implementation(libs.timber)`, then you could do: `Timber.tag("QRCodeActivity").e(e, "Error generating QR Code")`

ui/qr-code/src/main/kotlin/org.cru.godtools.qrcode/activity/QRCodeActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+                    .aspectRatio(1f)
+            )
+            Text(text = url, modifier = Modifier.align(Alignment.BottomCenter))
+        }
+    } else {
+        return
+    }
+}
+
+fun generateQRCode(url: String): Bitmap? {
```
frett: I believe we may want to generate the QR code bitmap on a background thread, how you would change that is to make this a `suspend` function, then when generating it from your composable use the `produceState {}` builder

frett: also, make this function private

ui/qr-code/src/main/kotlin/org.cru.godtools.qrcode/theme/Theme.kt
```diff
@@ -0,0 +1,58 @@
...
+    onPrimary = Color.White,
+    onSecondary = Color.White,
+    onTertiary = Color.White,
+    onBackground = Color(0xFF1C1B1F),
+    onSurface = Color(0xFF1C1B1F),
+    */
+)
+
+@Composable
+fun GodtoolsTheme(
```
frett: don't define a theme here, use the Material3 theme we already have in `:ui:base`

ui/qr-code/src/test [unitTest]/kotlin/org.cru.godtools.qrcode/ExampleUnitTest.kt
```diff
@@ -0,0 +1,17 @@
+package com.example.qr_code
```
Comment (frett) [inline ui/qr-code/src/test [unitTest]/kotlin/org.cru.godtools.qrcode/ExampleUnitTest.kt]: this unit test does nothing and should be removed

ui/qr-code/src/main/res/values/themes.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="Theme.Godtools" parent="android:Theme.Material.Light.NoActionBar" />
```
frett: I believe this theme definition might conflict with a godtools theme we have defined elsewhere, there should be a base theme already defined in `:ui:base`

ui/qr-code/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application
+        android:allowBackup="false"
+        android:icon="@mipmap/ic_launcher"
+        android:label="@string/app_name"
+        android:roundIcon="@mipmap/ic_launcher_round"
+        android:supportsRtl="true"
+        android:theme="@style/Theme.GodTools">
```
frett: you shouldn't be defining any properties on the application in a library module, they will be defined in the app module `:app`

ui/qr-code/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,24 @@
...
+        <activity
+            android:name=".activity.QRCodeActivity"
+            android:exported="true"
+            android:label="@string/app_name"
```
frett: I don't think we will need to define a lable for this activity

ui/qr-code/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+android {
+    namespace = "org.cru.godtools.qrcode"
+
+    configureCompose(project)
+
+    compileSdk {
+        version = release(36)
+    }
```
frett: this is defined for you by `godtools.library-conventions`

ui/qr-code/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+
+    defaultConfig {
+        minSdk = 24
+
+        testInstrumentationRunner = "androidx.test [unitTest].runner.AndroidJUnitRunner"
+    }
```
frett: also defined for you by the library conventions

ui/qr-code/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+
+    buildTypes {
+        release {
+            isMinifyEnabled = false
```
frett: no need to specify minify for the library, it's defined for the app itself in `:app`

ui/qr-code/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+            proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
```
frett: you don't need to include the default proguard rules file here, that is added at the app level when configuring minification

ui/qr-code/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_21
+        targetCompatibility = JavaVersion.VERSION_21
+    }
+    kotlinOptions {
+        jvmTarget = "21"
+    }
```
frett: these are defined by the library-conventions

ui/qr-code/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+    buildFeatures {
+        compose = true
+    }
```
frett: this is configured by the call to `configureCompose(project)`

ui/qr-code/build.gradle.kts
frett: it also looks like the `proguard-rules.pro` file doesn't currently exist, this whole line (and subsequently the `buildTypes {}` block) can be removed.

I can talk through what proguard/minification is if you would like, right now it's not something you should need to worry about

gradle/libs.versions.toml
```diff
@@ -242,3 +248,5 @@ ksp = { id = "com.google.devtools.ksp", version.ref = "ksp" }
 ktlint = { id = "org.jlleitschuh.gradle.ktlint", version.ref = "ktlintGradle" }
 mockposable = { id = "com.jeppeman.mockposable", version = "0.15" }
 paparazzi = { id = "app.cash.paparazzi", version = "2.0.0-alpha02" }
+android-application = { id = "com.android.application", version.ref = "agp" }
+kotlin-android = { id = "org.jetbrains.kotlin.android", version.ref = "kotlin" }
```
frett: these additional plugins aren't being used and can be removed.

I'm guessing they were auto-generated by the AS wizard

gradle/libs.versions.toml
```diff
@@ -36,6 +36,9 @@ play-app-update = "2.1.0"
 retrofit2 = "3.0.0"
 scarlet = "0.1.12"
 splitties = "3.0.0"
+agp = "8.13.0"
```
frett: this duplicates the `android-gradle-plugin` version, and can be removed after you remove the unused android application plugin

gradle/libs.versions.toml
```diff
@@ -36,6 +36,9 @@ play-app-update = "2.1.0"
...
+lifecycleRuntimeKtx = "2.9.4"
```
frett: this duplicates the `androidx-lifecycle` version

ui/base-tool/src/main/kotlin/org/cru/godtools/base/tool/ui/share/model/DefaultShareItem.kt
```diff
@@ -1,15 +1,49 @@
...
+                ChooserAction.Builder(
+                    Icon.createWithResource(activity, R.drawable.ic_qr_code),
+                    "Share by QR Code",
+                    PendingIntent.getActivity(
+                        activity,
+                        42,
+                        showQrCodeIntent,
+                        PendingIntent.FLAG_IMMUTABLE or PendingIntent.FLAG_UPDATE_CURRENT,
+                        options.toBundle()
+                    )
```
frett: I want to move the creation of the PendingIntent to an `Context.createQrCodePendingIntent(data: String)` extension function, it should call a `Context.createQrCodeIntent(data: String)` extension function for creating the QR code intent itself.

building the custom action that gets added to the chooser intent can stay in DefaultShareItem, the Intent & PendingIntent may have uses outside of this logic, and putting them "closer" to the `QRCodeActivity` itself will help with long term maintenance if the parameters need to change at some point

ui/qr-code/src/main/kotlin/org.cru.godtools.qrcode/activity/QRCodeActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+            Image(
+                bitmap = bitmap.asImageBitmap(),
+                contentDescription = "QR Code",
+                modifier = Modifier
+                    .aspectRatio(1f)
+            )
+            Text(text = url, modifier = Modifier.align(Alignment.BottomCenter))
+        }
+    } else {
+        return
```
frett: this else doesn't do anything and can be removed

ui/qr-code/src/main/kotlin/org.cru.godtools.qrcode/activity/QRCodeActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+fun generateQRCode(url: String): Bitmap? {
+    try {
+        val bitMatrix = QRCodeWriter().encode(url, BarcodeFormat.QR_CODE, 250, 250)
+        val qrCodeWidth: Int = bitMatrix.width
+        val qrCodeHeight: Int = bitMatrix.height
+
+        val bitmap = createBitmap(qrCodeWidth, qrCodeHeight, Bitmap.Config.RGB_565)
+        for (i in 0 until qrCodeWidth) {
+            for (j in 0 until qrCodeHeight) {
+                val color = if (bitMatrix.get(i, j)) -0x1000000 else -0x1 // needs Int?
```
frett: I would guess this is a `ColorInt`, there is an android color class that provides a white and black color constant, we should use that if possible to make it clear what this is doing

ui/qr-code/src/main/kotlin/org.cru.godtools.qrcode/activity/QRCodeActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+                bitmap.setPixel(i, j, color)
+            }
+        }
+        return bitmap
+    } catch (e: Exception) {
```
frett: in general it's bad form to catch a wide Exception like this, it could hide unexpected errors. If there is a specific exception that the code can throw you should determine how you want to handle that exception specifically. You can either catch that exception sub-class or let it propagate out if it's more appropriate for the caller to handle that error case.

ui/base-tool/build.gradle.kts
frett: After you create the `createIntent`/`createPendingIntent` extension methods, there is an `Acitivities.kt` file in `:ui:base` that you will want to move those to. That will allow you to trigger the QR code activity from other UI modules without needing to be directly dependent on the qr-code module.

ui/qr-code/src/main/kotlin/org.cru.godtools.qrcode/theme/Theme.kt
tjohnson009: This might have been auto generated as well. I don't recall writing this myself

frett: it probably was :)

ui/base-tool/src/main/kotlin/org/cru/godtools/base/tool/ui/share/model/DefaultShareItem.kt
```diff
@@ -1,15 +1,49 @@
...
+            }
+            val urlStringForQrCode = shareIntent.getStringExtra("EXTRA_SHARE_TOOL_URL")
+            val showQrCodeIntent = Intent(activity, QRCodeActivity::class.java)
+                .apply {
+                    putExtra("EXTRA_SHARE_URL", urlStringForQrCode)
+                }
+            val qrCodeCustomAction = arrayOf(
+                ChooserAction.Builder(
+                    Icon.createWithResource(activity, R.drawable.ic_qr_code),
+                    "Share by QR Code",
```
frett: this string needs to be added to `strings_tool_renderer` or `strings_base_ui`

ui/base-tool/build.gradle.kts
tjohnson009: @frett do you mean move them to an activities file in :ui:base or :ui:base-tool? My gut is telling me :ui:base-tool because that is where I was trying to have them utilized in the first place

frett: I would put them in `:ui:base`, it makes it a bit more visible for use by other share actions in the future

frett: At this point I remembered there were two, I can't remember if I was wanting to consolidate them eventually or keep them separate.

ui/base-tool/src/main/res/values/strings_tool_renderer.xml
```diff
@@ -13,16 +13,17 @@
     <string name="tool_share_sheet_more_apps">More</string>
     <string name="share_tool_message">This link will take you to a web version of what we were talking about today: %1$s \n\n
 
-The app can be downloaded from https://godtoolsapp.com \n\n
+    The app can be downloaded from https://godtoolsapp.com \n\n
```
frett:
```suggestion
The app can be downloaded from https://godtoolsapp.com \n\n
```

ui/base-tool/src/main/res/values/strings_tool_renderer.xml
```diff
@@ -13,16 +13,17 @@
...
 
-Let’s talk more!</string>
+    Let’s talk more!</string>
```
frett:
```suggestion
Let’s talk more!</string>
```

ui/base-tool/src/main/res/values/strings_tool_renderer.xml
frett: the white space change would actually show up in the message being sent when clicking share

frett:
it looks like you changed the `ic_launcher*` icons that were already in the app. I would guess they got changed when you were trying to revert add different versions of them.

ui/qr-code/src/main/kotlin/org/cru/godtools/qrcode/activity/QRCodeActivity.kt
```diff
@@ -0,0 +1,100 @@
+package org.cru.godtools.qrcode.activity
```
frett, line 1: it looks like you put this file in the wrong directory, it's showing as the directory: `org.cru.godtools.qrcode`, when it should be: `org/cru/godtools/qrcode`

ui/qr-code/src/main/res/drawable/ic_launcher_background.xml
```diff
@@ -0,0 +1,170 @@
+<?xml version="1.0" encoding="utf-8"?>
```
frett: this was a new launcher background that you shouldnt' be adding

ui/qr-code/src/main/res/drawable/ic_launcher_foreground.xml
```diff
@@ -0,0 +1,30 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
frett: this is a new launcher icon that shouldn't be added

ui/qr-code/src/main/res/mipmap-anydpi-v26/ic_launcher.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
```
frett: new launcher icon that shouldn't be added

ui/qr-code/src/main/res/mipmap-anydpi-v26/ic_launcher_round.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
```
frett: new launcher icon that shouldn't be added

ui/qr-code/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
```
frett: this file can be deleted, these are generic colors that were added that we shouldnt' be using

ui/qr-code/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="app_name">qr-code</string>
```
frett: this file needs to be deleted, the `app_name` string here will interfere with the correct `app_name` string defined elsewhere

ui/qr-code/src/main/kotlin/org/cru/godtools/qrcode/activity/QRCodeActivity.kt
tjohnson009: I don't totally understand this comment

frett: @tjohnson009 we can connect about this tomorrow if you are around, otherwise it can wait until after thanksgiving

ui/base-tool/src/main/kotlin/org/cru/godtools/base/tool/activity/BaseToolActivity.kt
```diff
@@ -214,22 +219,23 @@ abstract class BaseToolActivity<B : ViewDataBinding>(@LayoutRes contentLayoutId:
...
         title: String? = shareLinkTitle,
         @StringRes message: Int = shareLinkMessageRes,
-        shareUrl: String? = shareLinkUri
+        shareUrl: String? = shareLinkUri,
     ) = shareUrl?.let {
         Intent(Intent.ACTION_SEND).apply {
             type = "text/plain"
             putExtra(Intent.EXTRA_SUBJECT, title.orEmpty())
             putExtra(Intent.EXTRA_TEXT, getString(message, shareUrl))
+            putExtra("EXTRA_SHARE_TOOL_URL", shareUrl)
```
frett: I don't think this is necessary anymore now that we are passing the shareUrl via the `DefaultShareItem` constructor

ui/base-tool/src/main/kotlin/org/cru/godtools/base/tool/ui/share/model/DefaultShareItem.kt
```diff
@@ -2,14 +2,31 @@ package org.cru.godtools.base.tool.ui.share.model
...
+import android.service.chooser.ChooserAction
+import androidx.core.content.ContextCompat.getString
 import kotlinx.parcelize.Parcelize
 import org.ccci.gto.android.common.base.Ordered
+import org.cru.godtools.base.ui.createQrCodePendingIntent
+import org.cru.godtools.tool.R
 
 @Parcelize
-class DefaultShareItem(override val shareIntent: Intent) : ShareItem {
+class DefaultShareItem(override val shareIntent: Intent, var urlStringForQrCode: String? = null) : ShareItem {
```
frett:
```suggestion
class DefaultShareItem(override val shareIntent: Intent, val urlStringForQrCode: String? = null) : ShareItem {
```

ui/base-tool/src/main/kotlin/org/cru/godtools/base/tool/ui/share/model/DefaultShareItem.kt
```diff
@@ -2,14 +2,31 @@ package org.cru.godtools.base.tool.ui.share.model
...
     override val order get() = Ordered.HIGHEST_PRECEDENCE
 
     override fun triggerAction(activity: Activity) {
-        activity.startActivity(Intent.createChooser(shareIntent, null))
+        val chooserIntent = Intent.createChooser(shareIntent, null)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
```
frett:
```suggestion
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE && urlStringForQrCode != null) {
```

ui/qr-code/src/main/res/mipmap-anydpi-v26/ic_launcher_round.xml
frett: The incorrect launcher icons are still present

ui/qr-code/src/main/kotlin/org/cru/godtools/qrcode/activity/QRCodeActivity.kt
```diff
@@ -0,0 +1,100 @@
...
+        enableEdgeToEdge()
+
+        val url = intent.getStringExtra("EXTRA_SHARE_URL")
+        if (url == null) {
+            finish()
+            return
+        }
+
+        setContent {
+            MaterialTheme {
```
frett: this should be `GodToolsTheme`

ui/qr-code/src/main/kotlin/org/cru/godtools/qrcode/activity/QRCodeActivity.kt
```diff
@@ -0,0 +1,100 @@
...
+        value = generateQRCode(url)
+    }
+    val bitmap = bitmapState.value
+    if (bitmap != null) {
+        Box(
+            modifier = modifier.wrapContentSize(Alignment.Center)
+        ) {
+            Image(
+                bitmap = bitmap.asImageBitmap(),
+                contentDescription = "QR Code",
```
frett: we should technically extract this string to a strings resource file, for now let's just set this to null.

ui/qr-code/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+
+    configureCompose(project)
+}
+
+dependencies {
+    implementation(libs.androidx.activity.compose)
+    implementation(platform(libs.androidx.compose.bom))
+    implementation(libs.androidx.compose.ui)
+    implementation(libs.androidx.compose.ui.graphics)
+    implementation(libs.androidx.compose.ui.tooling.preview)
```
frett: this dependency can be removed, it's already defined via `configureCompose` in build-config

ui/qr-code/src/main/kotlin/org/cru/godtools/qrcode/activity/QRCodeActivity.kt
frett: which should be available by adding a dependency on `implementation(project(":ui:base"))`

ui/base/src/main/kotlin/org/cru/godtools/base/ui/Activities.kt
```diff
@@ -103,3 +106,33 @@ private fun Intent.putLanguagesExtra(vararg languages: Locale?) = putExtras(
     //       If the shortcut fails to open then singleString mode is still required
     Bundle().apply { putLocaleArray(EXTRA_LANGUAGES, languages.filterNotNull().toTypedArray(), true) }
 )
+
+private const val ACTIVITY_CLASS_QR_CODE = "org.cru.godtools.qrcode.activity.QRCodeActivity"
+
+fun Context.createQrCodeIntent(data: String?): Intent = Intent().setClassName(this, ACTIVITY_CLASS_QR_CODE)
+    .putExtra("EXTRA_SHARE_URL", data)
```
frett: make the `EXTRA_SHARE_URL` string a constant that you use both here and in `QRCodeActivity`. This helps ensure that any future updates to the extra name will apply to everywhere using it.

ui/qr-code/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+    configureCompose(project)
+}
+
+dependencies {
+    implementation(libs.androidx.activity.compose)
+    implementation(libs.androidx.compose.ui)
+    implementation(libs.androidx.compose.ui.graphics)
+    implementation(libs.androidx.compose.ui.tooling.preview)
+    implementation(libs.androidx.compose.material3)
+    implementation(project(":ui:base"))
```
frett:
```suggestion
    implementation(project(":ui:base"))

    implementation(libs.androidx.activity.compose)
    implementation(libs.androidx.compose.material3)
```

frett: this removes several dependencies that aren't necessary

gradle/libs.versions.toml
```diff
@@ -217,9 +218,12 @@ tinder-statemachine = "com.tinder.statemachine:statemachine:0.2.0"
 turbine = "app.cash.turbine:turbine:1.2.1"
 weakdelegate = "com.github.Karumi:WeakDelegate:1.0.1"
 youtubePlayer = "com.pierfrancescosoffritti.androidyoutubeplayer:core:13.0.0"
+zxing = "com.google.zxing:core:3.5.1"
 
 # HACK: dependencies used to trigger renovate upgrades
 ktlint = { module = "com.pinterest.ktlint:ktlint-cli", version.ref = "ktlint" }
+androidx-compose-bom = { group = "androidx.compose", name = "compose-bom", version.ref = "composeBom" }
+androidx-compose-ui-graphics = { group = "androidx.compose.ui", name = "ui-graphics" }
```
frett: these will no longer be used once you apply the last cleanup to the qr-code module dependencies I just commented

gradle/libs.versions.toml
```diff
@@ -36,6 +36,7 @@ play-app-update = "2.1.0"
 retrofit2 = "3.0.0"
 scarlet = "0.1.12"
 splitties = "3.0.0"
+composeBom = "2024.09.00"
```
frett: this also can be removed

ui/base-tool/src/main/res/values/strings_tool_renderer.xml
```diff
@@ -13,16 +13,17 @@
...
     <string name="share_tool_message_article">I wanted to share this article with you and would like to hear your thoughts about it.\n%1$s</string>
     <string name="feature_discovery_title_share_tool">Share a link</string>
     <string name="feature_discovery_desc_share_tool">Send this tool to a friend by clicking this icon</string>
 
     <!-- Shareables -->
     <eat-comment />
     <string name="tool_shareable_action_share_image">Share image</string>
+    <string name="share_by_qr_code">Share by QR Code</string>
```
frett: for organization purposes, just move this string up under the `Share Tool Functionality` section.

"Shareables" specifically relate to a feature of tools that contain additional custom shareable content

ui/base/src/main/kotlin/org/cru/godtools/base/ui/Activities.kt
```diff
@@ -103,3 +106,33 @@ private fun Intent.putLanguagesExtra(vararg languages: Locale?) = putExtras(
...
+
+fun Context.createQrCodePendingIntent(data: String?): PendingIntent {
+    val intent = createQrCodeIntent(data)
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+        val options = ActivityOptions.makeBasic().apply {
+            setPendingIntentCreatorBackgroundActivityStartMode(
+                ActivityOptions.MODE_BACKGROUND_ACTIVITY_START_ALLOWED
+            )
```
frett: out of curiosity, was there a reason you are setting this start mode? everything seems to work correctly without setting this

ui/qr-code/src/main/kotlin/org/cru/godtools/qrcode/activity/QRCodeActivity.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.google.zxing.qrcode.QRCodeWriter
+import org.cru.godtools.base.ui.theme.GodToolsTheme
+import timber.log.Timber
+
+class QRCodeActivity : ComponentActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        enableEdgeToEdge()
+
+        val url = intent.getStringExtra("EXTRA_SHARE_URL")
```
frett: this should use the constant I suggested defining in another comment

ui/base/src/main/kotlin/org/cru/godtools/base/ui/Activities.kt
frett: if we don't actually need the start mode option, this entire function could be simplified to:

```kotlin
fun Context.createQrCodePendingIntent(
    data: String,
    requestCode: Int = 0,
    flags: Int = PendingIntent.FLAG_IMMUTABLE
): PendingIntent = PendingIntent.getActivity(this, requestCode, createQrCodeIntent(data), flags)
```

frett: hmm, I'm going to do a bit more reading on this, don't do anything about this right now

gradle/libs.versions.toml
tjohnson009: # HACK: dependencies used to trigger renovate upgrades
ktlint = { module = "com.pinterest.ktlint:ktlint-cli", version.ref = "ktlint" }
androidx-compose-bom = { group = "androidx.compose", name = "compose-bom", version.ref = "composeBom" }
androidx-compose-ui-graphics = { group = "androidx.compose.ui", name = "ui-graphics" }

The composeBOM is only present here but it's under the HACK for renovate. Can I still remove it or is it needed for Renovate to run properly?

ui/base/src/main/kotlin/org/cru/godtools/base/ui/Activities.kt
tjohnson009: OK I'll hold off on anything for now. A few iterations ago, the QRCodeActivity would not launch without that setting. I was getting errors messages saying that it could not launch anything in the background.

frett: k, I don't think it will harm anything having that present, the documentation about this seems a bit unclear.

I am going to clean up this function a bit though to reduce some of the duplication that is present.

frett (approved):
looks good, I did simplify the logic of `createQrCodePendingIntent()`. Feel free to use the Squash and Merge option instead of worrying about rebasing to clean up the commit history

PR #4267, opened 2025-12-15
GT-2869 Support parsing knowgod.com lesson deep links

frett:


ui/lesson-renderer/src/main/kotlin/org/cru/godtools/tool/lesson/ui/LessonActivity.kt
```diff
@@ -277,6 +279,14 @@ class LessonActivity :
...
                     dataModel.locale.value = Locale.forLanguageTag(path[2])
                 }
+
+                // Defer to LessonDeepLink processing
+                else -> {
+                    val deepLink = LessonDeepLink.parseKnowGodDeepLink(data) ?: return
+                    dataModel.toolCode.value = deepLink.lesson
+                    dataModel.locale.value = deepLink.locale
+                    initialPage = deepLink.page
+                }
```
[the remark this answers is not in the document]
frett: initialPage is only used to create the lesson pager state initially, after that the current lesson state will be restored and the initialPage property is meaningless

ui/lesson-renderer/src/main/kotlin/org/cru/godtools/tool/lesson/LessonDeepLink.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import android.net.Uri
+import java.util.Locale
+import org.cru.godtools.base.HOST_KNOWGOD_COM
+
+internal data class LessonDeepLink(val lesson: String, val locale: Locale, val page: Int? = null) {
+    companion object {
+        // region https://knowgod.com/en/lesson/lessonhs/1
+        private const val KNOWGOD_PATH_LOCALE = 0
+        private const val KNOWGOD_PATH_LESSON = 2
```
tjohnson009, line 11: I don't think this is hardcoded, but it seems like it is to me. Is this for some part of the deeplink uri? Like a parameter for the final deeplink string that is going to be passed?

frett: These constants are for the index position of path parameters we are extracting from the url.

For the path: `/en/lesson/lessons/7`
We care about:
- 0, `en` this is the locale we want to load the lesson in
- 2, `lessonhs` this is the unique identifier for the lesson we want to load
- 3, `7` this is the page we want to initially load, this is optional

ui/lesson-renderer/src/test/kotlin/org/cru/godtools/tool/lesson/LessonDeepLinkTest.kt
```diff
@@ -0,0 +1,53 @@
...
+
+@RunWith(AndroidJUnit4::class)
+class LessonDeepLinkTest {
+    // region parseKnowGodDeepLink()
+    @Test
+    fun `parseKnowGodDeepLink() - Base`() {
+        assertNotNull(LessonDeepLink.parseKnowGodDeepLink(Uri.parse("https://knowgod.com/en/lesson/lessonhs"))) {
+            assertEquals("lessonhs", it.lesson)
+            assertEquals(Locale.forLanguageTag("en"), it.locale)
+            assertNull(it.page)
```
tjohnson009, line 20: This is null because we are at something like a "homepage" right? There is no page to deep link to yet in this test, correct?

ui/lesson-renderer/src/test/kotlin/org/cru/godtools/tool/lesson/LessonDeepLinkTest.kt
```diff
@@ -0,0 +1,53 @@
...
+        assertNotNull(LessonDeepLink.parseKnowGodDeepLink(Uri.parse("https://knowgod.com/en/lesson/lessonhs/1"))) {
+            assertEquals("lessonhs", it.lesson)
+            assertEquals(Locale.forLanguageTag("en"), it.locale)
+            assertEquals(1, it.page)
+        }
+    }
+
+    @Test
+    fun `parseKnowGodDeepLink() - Non-numeric Page`() {
+        assertNotNull(LessonDeepLink.parseKnowGodDeepLink(Uri.parse("https://knowgod.com/en/lesson/lessonhs/one"))) {
```
tjohnson009: Would this be the scenario if the user tried to type in a page number in the url? If not, what scenario would this be necessary for?

ui/lesson-renderer/src/main/kotlin/org/cru/godtools/tool/lesson/LessonDeepLink.kt
frett: They are also private since they only have meaning when parsing knowgod lesson urls, which should be fully handled by this class

ui/lesson-renderer/src/test/kotlin/org/cru/godtools/tool/lesson/LessonDeepLinkTest.kt
frett: deep link urls for tools/lessons generally have the deep linked page as being optional. if it's not specified the tool/lesson starts at the initial page.

So this `page` property is null because it wasn't present in the URL.

frett: Also of note, the `LessonDeepLink` class is only responsible for parsing a deep link and returning the different properties about it. It doesn't & shouldn't contain any actual loading/business logic around default behavior

ui/lesson-renderer/src/test/kotlin/org/cru/godtools/tool/lesson/LessonDeepLinkTest.kt
frett: I used Copilot to generate tests for this PR, `Non-numeric Page` is a poor name for this test. I'll take a look to see if I can consolidate/cleanup the tests.

The important part about why this test exists is to test corner cases. a url with `/one` for the page number would be invalid based upon how we have defined lesson deeplinks

tjohnson009: helpful to know!

tjohnson009 (approved):
Looks good from what I can tell!

PR #4341, opened 2026-03-02
GT-2676 Offline Warning Dialog for Downloadable Languages

tjohnson009:
- **add offline dialog state and ui event to screen**
- **render offline warning dialog in downloadable languages layout**
- **stop language pin and show offline dialog when not connected**

### DIFF
--- a/app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesLayout.kt
+++ b/app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesLayout.kt
@@ -121,7 +121,7 @@ fun DownloadableLanguagesLayout(state: UiState, modifier: Modifier = Modifier) {
             modifier = Modifier.padding(contentPadding)
         ) {
             itemsIndexed(state.languages, key = { _, it -> it.language.code }) { i, it ->
-                LanguageListItem(it, state.eventSink, isConnected = { isConnected }, Modifier.animateItem())
+                LanguageListItem(it, state.eventSink, { isConnected }, Modifier.animateItem())
                 if (i < state.languages.lastIndex) HorizontalDivider(Modifier.animateItem())
             }
         }
--- a/app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesLayout.kt
+++ b/app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesLayout.kt
@@ -12,7 +12,6 @@ import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.filled.ArrowBack
 import androidx.compose.material.icons.filled.Close
-import androidx.compose.material3.AlertDialog
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.Icon
@@ -22,7 +21,6 @@ import androidx.compose.material3.Scaffold
 import androidx.compose.material3.SearchBar
 import androidx.compose.material3.SearchBarDefaults
 import androidx.compose.material3.Text
-import androidx.compose.material3.TextButton
 import androidx.compose.material3.ripple
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
@@ -39,6 +37,7 @@ import androidx.compose.ui.res.pluralStringResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import com.slack.circuit.codegen.annotations.CircuitInject
+import com.slack.circuit.overlay.LocalOverlayHost
 import dagger.hilt.components.SingletonComponent
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
@@ -69,19 +68,6 @@ fun DownloadableLanguagesLayout(state: UiState, modifier: Modifier = Modifier) {
     }
 
     BackHandler(enabled = searchQuery.isNotEmpty()) { updateSearchQuery("") }
-
-    if (state.showOfflineDialog) {
-        AlertDialog(
-            onDismissRequest = { state.eventSink(UiEvent.DismissOfflineDialog) },
-            title = { Text(stringResource(R.string.account_error_not_connected_title)) },
-            text = { Text(stringResource(R.string.account_error_not_connected)) },
-            confirmButton = {
-                TextButton(onClick = { state.eventSink(UiEvent.DismissOfflineDialog) }) {
-                    Text(stringResource(android.R.string.ok))
-                }
-            },
-        )
-    }
 
     Scaffold(
         topBar = {
@@ -134,7 +120,7 @@ fun DownloadableLanguagesLayout(state: UiState, modifier: Modifier = Modifier) {
             modifier = Modifier.padding(contentPadding)
         ) {
             itemsIndexed(state.languages, key = { _, it -> it.language.code }) { i, it ->
-                LanguageListItem(it, state.eventSink, Modifier.animateItem())
+                LanguageListItem(it, state.eventSink, state.isConnected.value, Modifier.animateItem())
                 if (i < state.languages.lastIndex) HorizontalDivider(Modifier.animateItem())
             }
         }
@@ -142,7 +128,12 @@ fun DownloadableLanguagesLayout(state: UiState, modifier: Modifier = Modifier) {
 }
 
 @Composable
-private fun LanguageListItem(state: UiLanguage, eventSink: (UiEvent) -> Unit, modifier: Modifier = Modifier) = ListItem(
+private fun LanguageListItem(
+    state: UiLanguage,
+    eventSink: (UiEvent) -> Unit,
+    isConnected: Boolean,
+    modifier: Modifier = Modifier
+) = ListItem(
     headlineContent = { LanguageName(state.language) },
     supportingContent = {
         Text(
@@ -155,6 +146,9 @@ private fun LanguageListItem(state: UiLanguage, eventSink: (UiEvent) -> Unit, mo
     },
     trailingContent = {
         var confirmRemoval by rememberSaveable { mutableStateOf(false) }
+        val coroutineScope = rememberCoroutineScope()
+        val overlayHost = LocalOverlayHost.current
+
         LaunchedEffect(confirmRemoval) {
             delay(3_000)
             confirmRemoval = false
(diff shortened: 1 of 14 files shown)


app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesLayout.kt
```diff
@@ -68,6 +70,19 @@ fun DownloadableLanguagesLayout(state: UiState, modifier: Modifier = Modifier) {
...
+        AlertDialog(
+            onDismissRequest = { state.eventSink(UiEvent.DismissOfflineDialog) },
+            title = { Text(stringResource(R.string.account_error_not_connected_title)) },
+            text = { Text(stringResource(R.string.account_error_not_connected)) },
+            confirmButton = {
+                TextButton(onClick = { state.eventSink(UiEvent.DismissOfflineDialog) }) {
+                    Text(stringResource(android.R.string.ok))
+                }
+            },
+        )
```
frett: take a look at using a Circuit overlay to do this

app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesPresenter.kt
```diff
@@ -74,10 +79,16 @@ class DownloadableLanguagesPresenter @AssistedInject constructor(
...
 
-                    is UiEvent.PinLanguage -> coroutineScope.launch(NonCancellable) {
-                        languagesRepository.pinLanguage(it.locale)
+                    is UiEvent.PinLanguage -> if (connected) {
+                        coroutineScope.launch(NonCancellable) {
+                            languagesRepository.pinLanguage(it.locale)
+                        }
+                    } else {
+                        showOfflineDialog = true
                     }
```
frett: let's wait for Jill to respond before we block pinning the language when offline. I feel like blocking it is arbitrarily restricting functionality.

app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesScreen.kt
```diff
@@ -16,12 +16,14 @@ object DownloadableLanguagesScreen : Screen {
     data class UiState(
         val query: MutableState<String> = mutableStateOf(""),
         val languages: ImmutableList<UiLanguage> = persistentListOf(),
+        val showOfflineDialog: Boolean = false,
```
frett: There are 2 ways to approach this with Circuit overlays, and I think there are pros & cons of either solution.

1. Very similar to how you are currently doing this, it uses the `showOfflineDialog` boolean to trigger an `OverlayEffect` that triggers an `eventSink` callback to dismiss the dialog
    - pro: keeps the business logic of deciding when to show a dialog in the Presenter
    - con: a lot of overhead for managing the dialog
        - can be reduced by exposing the state as: `val showOfflineDialog: MutableState<Boolean> = mutableStateOf(false)`
2. Treat it more as an action interrupt
    - Put `isConnected` in the state. (I would actually put this in as `isConnected: State<Boolean>` to prevent re-composition when the users connected state changes)
    - in the pin language onClick use a coroutine scope to trigger the overlay on the `OverlayHost` directly.
    - pro: This would be a more straight forward flow where triggering the dialog is more tightly coupled to the action
    - pro: allows the overlay to modify the original action before firing off the action
    - con: blurring of the separation of business logic from UI logic

I'm kind of leaning towards solution 2 for this, I believe testing is possible by providing a mock OverlayHost that can be used to verify an overlay is triggered or not (and that it returns the expected overlay result)

frett: I guess the question is how do we classify the desire to show an offline dialog overlay, is it just a UI concern when the user is offline? or is it actually more complex business logic that should be in the presenter with the rest of the business logic?

app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesLayout.kt
```diff
@@ -68,6 +70,19 @@ fun DownloadableLanguagesLayout(state: UiState, modifier: Modifier = Modifier) {
...
     BackHandler(enabled = searchQuery.isNotEmpty()) { updateSearchQuery("") }
 
+    if (state.showOfflineDialog) {
+        AlertDialog(
+            onDismissRequest = { state.eventSink(UiEvent.DismissOfflineDialog) },
+            title = { Text(stringResource(R.string.account_error_not_connected_title)) },
+            text = { Text(stringResource(R.string.account_error_not_connected)) },
+            confirmButton = {
+                TextButton(onClick = { state.eventSink(UiEvent.DismissOfflineDialog) }) {
+                    Text(stringResource(android.R.string.ok))
```
frett: I always forget that android provides some base strings like this

app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesLayout.kt
```diff
@@ -68,6 +70,19 @@ fun DownloadableLanguagesLayout(state: UiState, modifier: Modifier = Modifier) {
 
     BackHandler(enabled = searchQuery.isNotEmpty()) { updateSearchQuery("") }
 
+    if (state.showOfflineDialog) {
+        AlertDialog(
+            onDismissRequest = { state.eventSink(UiEvent.DismissOfflineDialog) },
+            title = { Text(stringResource(R.string.account_error_not_connected_title)) },
+            text = { Text(stringResource(R.string.account_error_not_connected)) },
```
frett: so, I'm not a fan of re-using strings from other parts of the app like this. Those strings are specifically around the user authenticating. If there is context added to the strings about it being related to login all of a sudden those strings don't make sense being displayed for pinning a language. This additional context could show up unexpectantly in other languages.

English has plenty of words that have multiple meanings which would translate differently based upon the context they are used in, we want to make sure things are setup so that translators don't have to handle dual meaning for phrases

frett (changes requested):

app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesScreen.kt
tjohnson009: I think after the standup on Tuesday because we are continuing the pin when internet is reestablished that I would look at this as more of a UI concern. Now it really seems more like just a message to the user instead of business logic decision. I also think an overlay feels like it matches with that expectation as well.

app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesLayout.kt
tjohnson009: That makes sense. The thought flashed into my mind if I should create new strings. In light of what you said it seems best to create new strings unless the context is the same or otherwise stated

app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesScreen.kt
frett: good reasoning :)

app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesLayout.kt
frett: yeah, it's easy for Aaron & Jill to copy the pre-existing strings in Crowdin

app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesPresenter.kt
```diff
@@ -49,16 +50,19 @@ class DownloadableLanguagesPresenter @AssistedInject constructor(
...
 ) : Presenter<UiState> {
     @Composable
     override fun present(): UiState {
         val coroutineScope = rememberCoroutineScope()
         val query = rememberSaveable { mutableStateOf("") }
+        val connected = isConnected.collectAsState()
 
         return UiState(
             query = query,
+            isConnected = connected,
```
frett: you can just inline this since you are only using it in 1 place


```suggestion

        return UiState(
            query = query,
            isConnected = isConnected.collectAsState(),
```

app/src/main/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesLayout.kt
```diff
@@ -119,15 +120,20 @@ fun DownloadableLanguagesLayout(state: UiState, modifier: Modifier = Modifier) {
             modifier = Modifier.padding(contentPadding)
         ) {
             itemsIndexed(state.languages, key = { _, it -> it.language.code }) { i, it ->
-                LanguageListItem(it, state.eventSink, Modifier.animateItem())
+                LanguageListItem(it, state.eventSink, state.isConnected.value, Modifier.animateItem())
```
frett: so we can actually defer the read of the `isConnected` State value better by using a closure for reading it.

```kotlin
LanguageListItem(isConnected = { state.isConnected.value })

private fun LanguageListItem(isConnected: () -> Boolean) {
    UiElement(
        modifier = Modifier.clickable {
            if (!isConnected()) {
                 // show overlay
            }
        }
    )
}
```

frett: that way the state value is only read when the user clicks on a  list item and not every time the list item is composed

frett: The other thing you can do the clean up the code more is at the top of the Layout use this:

```kotlin
val isConnected by state.isConnected
```

that way you can just refer to it as `isConnected` when calling `LanguageListItem`

frett: there's lots of ways to do the same thing ;)

app/src/testDebug/kotlin/org/cru/godtools/ui/languages/downloadable/DownloadableLanguagesLayoutPaparazziTest.kt
```diff
@@ -37,20 +39,23 @@ class DownloadableLanguagesLayoutPaparazziTest(
...
             ),
-        )
+        ),
+        isConnected = isConnected,
     )
 
     @Test
     fun `DownloadableLanguagesLayout()`() {
-        snapshot { DownloadableLanguagesLayout(state) }
+        snapshot { ContentWithOverlays { DownloadableLanguagesLayout(state) } }
```
frett: if possible use `ContentWithOverlays` in the `snapshot` function, it'll be used anywhere we do overlays, so having it centralized for testing makes sense

frett (changes requested):
Everything is looking great, just the one comment about using `ContentWithOverlays` in the `snapshot` method so it can be shared across all snapshot tests

tjohnson009:
It seems like the overlay might have affected a screenshot test for another composable. It is the QRCodeComposable.  I'm looking into it, probably won't get to it completely until tomorrow.

ui/base/build.gradle.kts
```diff
@@ -58,6 +58,7 @@ dependencies {
     testImplementation(projects.ui.tractRenderer)
 
     testFixturesApi(libs.testparameterinjector)
+    testFixturesImplementation(libs.circuit.overlay)
```
frett: minor nit: alpha sort in this list

frett (approved):
looks good, just a minor nit and tests need to pass, but you're good to merge

PR #4363, opened 2026-03-26
GT-2978 Menu Localization Settings Setup

tjohnson009:
- **Create and add menu item for localization settings**
- **Add localization country setting to Settings**
- **Add localization settings string resources**
- **Minor comment removal from drawer item source code**
- **Create and structure localization settings layout**
- **Create and structure localization settings presenter**
- **Create and structure localization settings screen**


app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayout.kt
```diff
@@ -0,0 +1,174 @@
...
+                .background(ListItemDefaults.containerColor),
+        ) {
+            item {
+                Box(
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .background(MaterialTheme.colorScheme.primaryContainer)
+                        .padding(32.dp),
+                ) {
+                    CompositionLocalProvider(LocalContentColor provides MaterialTheme.colorScheme.scrim) {
```
frett: the scrim color is meant to be the color that overlays background content when a dialog is displayed.

Use the `onPrimaryContainer` color for the text in the banner

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayout.kt
```diff
@@ -0,0 +1,174 @@
...
+        modifier = modifier,
+    ) { contentPadding ->
+        LazyColumn(
+            modifier = Modifier
+                .padding(contentPadding)
+                .fillMaxSize()
+                .background(ListItemDefaults.containerColor),
+        ) {
+            item {
+                Box(
```
frett: You might be able to use a `Surface` composable here since. It's a base Material 3 component for any container that holds content. I think it has a containerColor and contentColor property that will automatically assign appropriate colors to the container and content.

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayout.kt
```diff
@@ -0,0 +1,174 @@
...
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .background(MaterialTheme.colorScheme.primaryContainer)
+                        .padding(32.dp),
```
frett: try and line up the margins so that it's the same indention as the rest of the content. which means you probably just want `16dp` of padding for the content

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsPresenter.kt
```diff
@@ -0,0 +1,113 @@
...
+            countryCode = settingCountryCode,
+            eventSink = { event ->
+                when (event) {
+                    UiEvent.NavigateBack -> navigator.pop()
+
+                    is UiEvent.SelectCountry -> {
+                        scope.launch {
+                            settings.updateCountrySetting(isoCode = event.isoCode)
+                        }
+                        navigator.pop()
```
frett: the write could actually race popping the screen off the back stack.

When a composable is disposed the coroutine scope returned by `rememberCoroutineScope()` will be canceled, which could cancel any action that is still running in the background. the call to `updateCountrySetting()` may not have finished.

The easy solution is to move the `navigator.pop()` into the `launch { }` block

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayout.kt
```diff
@@ -0,0 +1,174 @@
...
+@Composable
+private fun CountryName(country: CountryItem, modifier: Modifier = Modifier) {
+    val color = LocalTextStyle.current.color.takeOrElse { LocalContentColor.current }
+    Text(
+        remember(country, color) {
+            buildAnnotatedString {
+                withStyle(SpanStyle(color = color)) { append(country.nativeName) }
+                withStyle(SpanStyle(color = color.copy(alpha = color.alpha * 0.60f))) {
+                    append("  ${country.displayName}")
+                }
```
frett: the `LanguageName` composable is using an 8sp gap between the 2 parts of the label, we should duplicate that here instead of using a double space.

It actually looks like the base `LanguageName` composable takes 2 strings, we could probably refactor a bit to maybe have an internal `LocalizedName` composable that both countries & languages use.

frett:
## PR Review: GT-2978 Menu Localization Settings Setup (#4363)

### Summary
Adds a "Localization settings" menu item to the navigation drawer that opens a new `CountrySettingsScreen` — a searchable list of countries backed by a Circuit Presenter/UI pair. Also adds `getCountrySettingFlow()`/`updateCountrySetting()` to `Settings`.

---

### ✅ Looks Good
- `CountrySettingsScreen` correctly annotated with `@Parcelize`, implemented as `data object : Screen`
- `@AssistedInject` / `@AssistedFactory` / `@CircuitInject` structure on `Factory` is correct
- All composable UI events flow through `state.eventSink(UiEvent.*)` — no direct calls
- `GodToolsTheme.searchBarColors` used correctly in `CountrySettingsLayout`
- Both composables (`CountryName`, `CountrySettingsLayout`) have `modifier: Modifier = Modifier`
- Strings extracted to `strings_languages.xml`; native names correctly marked `translatable="false"`
- Drawer item properly calls `eventSink(Event.DismissDrawer)` after navigation
- `Settings.kt` additions follow existing patterns
- ktlint: **PASS**

---

### ❌ Must Fix

- **`CountrySettingsPresenter.kt:262`** — `countries: ImmutableList<CountryItem>` in `UiState` should be plain `List<CountryItem>`. Project convention is no `ImmutableList` in UiState.

- **Mismatched Paparazzi snapshots** — The committed snapshot files are named `org.cru.godtools.ui.languages.localization_LocalizationSettingsLayoutPaparazziTest_LocalizationSettingsLayout()_*` but the test class in the diff is `CountrySettingsLayoutPaparazziTest` in package `org.cru.godtools.ui.languages.country`. These don't match — `verifyPaparazzi` will fail. The orphaned `LocalizationSettingsLayoutPaparazziTest` snapshots need to be deleted and new ones generated via CI for `CountrySettingsLayoutPaparazziTest`.

---

### ⚠️ Minor Issues

- **`CountrySettingsPresenter.kt:270`** — `sealed interface UiEvent` is missing `internal`. Convention: `UiEvent` should be `internal sealed interface`.

- **`CountrySettingsLayout.kt:186`** — `LocalContentColor provides MaterialTheme.colorScheme.scrim` on a `primaryContainer` background is semantically incorrect. `scrim` is a dark overlay color, not a text-on-surface token. The correct pair for `primaryContainer` is `onPrimaryContainer`.

- **`CountrySettingsPresenter.kt:313`** — `rememberSortedCountries` iterates all ~250 ISO country codes, calling `context.resources.getIdentifier()` and `context.getString()` per entry, synchronously inside `remember {}` on the composition thread. Should use `produceState { withContext(Dispatchers.IO) { … } }` per project convention for CPU-heavy work.

---

*Review generated by Claude Code*

app/src/main/kotlin/org/cru/godtools/ui/languages/LanguageName.kt
```diff
@@ -23,23 +23,23 @@ import org.cru.godtools.base.util.getDisplayName
...
+    LocalizedName(
         name = remember(context, locale, appLanguage) { locale.getDisplayName(context, inLocale = locale) },
         secondName = remember(context, locale) { locale.getDisplayName(context, inLocale = appLanguage) },
         modifier = modifier,
     )
 }
 
 @Composable
-internal fun LanguageName(language: Language, modifier: Modifier = Modifier) {
+internal fun LocalizedName(language: Language, modifier: Modifier = Modifier) {
```
frett: my feedback meant only changing the one that took 2 strings, not the methods that take an actual `Language` object

app/src/main/kotlin/org/cru/godtools/ui/languages/LanguageName.kt
```diff
@@ -49,7 +49,7 @@ internal fun LanguageName(language: Language, modifier: Modifier = Modifier) {
 private const val LANGUAGE_NAME_GAP = "[gap]"
 
 @Composable
-private fun LanguageName(name: String, secondName: String, modifier: Modifier = Modifier) {
+internal fun LocalizedName(name: String, secondName: String, modifier: Modifier = Modifier) {
```
frett: let's move this common re-usable composable out to a separate `LocalizedName` file

app/src/main/kotlin/org/cru/godtools/ui/languages/LanguageSettingsLayout.kt
```diff
@@ -153,7 +153,7 @@ internal fun LanguageSettingsLayout(state: UiState, modifier: Modifier = Modifie
 
             itemsIndexed(pinnedLanguages, key = { _, it -> it.code }) { i, it ->
                 if (i == 0) HorizontalDivider(modifier = Modifier.animateItem())
-                LanguageName(
+                LocalizedName(
```
frett: once you fix the `LanguageName` composable so that only the inner one that takes Strings is renamed this change shouldn't be here anymore

app/src/testDebug/kotlin/org/cru/godtools/ui/languages/LanguageNamePaparazziTest.kt
```diff
@@ -27,13 +27,13 @@ class LanguageNamePaparazziTest(@TestParameter nightMode: NightMode) :
...
+            LocalizedName(Locale("ar"))
             HorizontalDivider()
-            LanguageName(Locale.forLanguageTag("bs-BA"))
+            LocalizedName(Locale.forLanguageTag("bs-BA"))
             HorizontalDivider()
-            LanguageName(Locale.ENGLISH)
+            LocalizedName(Locale.ENGLISH)
             HorizontalDivider()
-            LanguageName(language = forcedNameLanguage)
+            LocalizedName(language = forcedNameLanguage)
```
frett: these changes won't exist after you correct the `LanguageName` composable name

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsPresenter.kt
```diff
@@ -0,0 +1,113 @@
...
+                    "string",
+                    context.packageName,
+                )
+                val nativeName = if (resourceId != 0) context.getString(resourceId) else displayName
+                CountryItem(isoCode, displayName, nativeName)
+            }.sortedWith(
+                compareBy(appLocale.getPrimaryCollator()) { it.displayName }
+            ).toImmutableList()
+        }
+    }
```
frett, line 111: there was a comment Claude raised about converting this to a produceState() method to do the generation of the list on a background thread.

that would look something like:

```
produceState(key = appLocale, initial = emptyList()) {
    // generate the list here
    val list = ...

    emit(list)
}
```

frett: it's a good suggestion because resolving 200 country names from internationalization tables and the strings file could trigger multiple disk reads which would lag the UI thread

frett: the `key` would cause it to re-evaluate when the appLocal changes (which should be rare), and the `initial` value is what gets returned before it has a chance to load the actual list

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayout.kt
```diff
@@ -0,0 +1,151 @@
...
+                    }
+                }
+            }
+            itemsIndexed(state.countries, { _, it -> it.isoCode }) { i, country ->
+                if (i > 0) HorizontalDivider(Modifier.padding(horizontal = 16.dp))
+
+                ListItem(
+                    headlineContent = { CountryName(country) },
+                    trailingContent = {
+                        if (country.isoCode == state.countryCode) {
```
frett, line 142: is there a `selected` property available on `ListItem`?

frett: nope, don't see one

frett: I'm not seeing anywhere else that we're using the Material 3 selected item styles, so using the checkmark to indicate selection is fine

app/src/testDebug/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayoutPaparazziTest.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.runner.RunWith
+
+@RunWith(TestParameterInjector::class)
+class CountrySettingsLayoutPaparazziTest(
+    @TestParameter(valuesProvider = DeviceConfigProvider::class) deviceConfig: DeviceConfig,
+    @TestParameter nightMode: NightMode,
+    @TestParameter accessibilityMode: AccessibilityMode,
+) : BasePaparazziTest(deviceConfig = deviceConfig, nightMode = nightMode, accessibilityMode = accessibilityMode) {
+    private val countries = persistentListOf(
+        CountryItem("BR", "Brazil", "Brasil"),
```
frett, line 23: to show how it handles RTL languages, you should add one of the countries that has the native name in Arabic

app/src/testDebug/kotlin/org/cru/godtools/ui/drawer/DrawerMenuLayoutPaparazziTest.kt
```diff
@@ -0,0 +1,21 @@
...
+@RunWith(TestParameterInjector::class)
+class DrawerMenuLayoutPaparazziTest(
+    @TestParameter(valuesProvider = DeviceConfigProvider::class) deviceConfig: DeviceConfig,
+    @TestParameter nightMode: NightMode,
+    @TestParameter accessibilityMode: AccessibilityMode,
+) : BasePaparazziTest(deviceConfig = deviceConfig, nightMode = nightMode, accessibilityMode = accessibilityMode) {
+    @Test
+    fun `DrawerMenuLayout() - Open`() {
+        snapshot { DrawerMenuLayout(state = DrawerMenuScreenStateTestData.open) {} }
+    }
```
frett, line 20: Eventually this will probably be part of the dashboard paparazzi tests. But that UI is not fully setup to use Circuit yet

frett (changes requested):
I think just the produceState change and adding a RTL name to the country list to see how it renders.

app/src/testDebug/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayoutPaparazziTest.kt
frett: it could actually be a separate RTL paparazzi test if you'd like

app/src/testDebug/kotlin/org/cru/godtools/ui/drawer/DrawerMenuLayoutPaparazziTest.kt
tjohnson009: I was wondering about that when I was looking for where to put the tests, but that's helpful to know

frett: technically screenshots for this already show up on the tool details UI since that actually supports the nav drawer. But we probably need to remove the nav drawer from that UI since it's not indicated it's there and doesn't feel like it makes logical sense to have it there.

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsPresenter.kt
```diff
@@ -85,24 +88,26 @@ class CountrySettingsPresenter @AssistedInject constructor(
...
+                    val resourceId = context.resources.getIdentifier(
+                        "country_native_name_$isoCode",
+                        "string",
+                        context.packageName,
+                    )
+                    val nativeName = if (resourceId != 0) context.getString(resourceId) else displayName
+                    CountryItem(isoCode, displayName, nativeName)
+                }.sortedWith(
+                    compareBy(appLocale.getPrimaryCollator()) { it.displayName }
+                ).toImmutableList()
```
frett: you can get rid of the `.toImmutableList()` since we are just returning a `List` anyways

frett (approved):

PR #4372, opened 2026-04-08
GT-2979 Country Settings Confirmation Overlay

tjohnson009:
- **Bring in svg country settings icon from figma**
- **Modify clickable to call the new overlay to confirm country selection**
- **Add country settings overlay string resources with format args**
- **Build country settings confirmation overlay with new string resources and icon**
- **Make country displayName nullable for better null handling**
- **Show the selected country isoCode is displayName is null**


app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsConfirmationOverlay.kt
```diff
@@ -0,0 +1,110 @@
...
+import androidx.compose.ui.unit.dp
+import com.slack.circuit.overlay.Overlay
+import com.slack.circuit.overlay.OverlayNavigator
+import org.cru.godtools.R
+
+class CountrySettingsConfirmationOverlay(val countryName: String?) : Overlay<Boolean> {
+    @OptIn(ExperimentalMaterial3Api::class)
+    @Composable
+    override fun Content(navigator: OverlayNavigator<Boolean>) {
+        BasicAlertDialog(
```
frett: Can we use the `AlertDialog` that lays things out for us and provides styling for the confirm/dismiss buttons to match the rest of the dialogs in the app?

frett: we won't need a close button in the top corner, that should be handled by the cancel/dismiss at the bottom

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsConfirmationOverlay.kt
```diff
@@ -0,0 +1,110 @@
...
+                    val fullText = stringResource(R.string.country_settings_overlay_text, countryName)
+                    val start = fullText.indexOf(countryName)
+                    val annotated = buildAnnotatedString {
+                        append(fullText)
+                        addStyle(
+                            SpanStyle(color = MaterialTheme.colorScheme.primary),
+                            start,
+                            start + countryName.length
+                        )
+                    }
```
frett: take a look at `ConfirmAppLanguageMessage` for an `annotatedStringResource` builder to use instead

frett: and contrary to what is in `ConfirmAppLanguageMessage`, use the primary color like you are currently doing

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsConfirmationOverlay.kt
frett:
```kotlin
AlertDialog(
    icon = { Icon(painterResource(R.drawable.country_setting_overlay_icon) },
    text = { 
        val annotatedString = annotatedStringResource()
        Text(annotatedString)
    },
    confirmButton = { TextButton() },
    dismissButton = { TextButton() },
)
```

app/src/main/res/values/strings_languages.xml
```diff
@@ -27,6 +27,9 @@
     <string name="country_settings_title">Select a country</string>
     <string name="country_settings_subtitle_header">Select your localization</string>
     <string name="country_settings_subtitle_body">Choose a country below to set your localization and get prioritized content and recommendations based on your location.</string>
+    <string name="country_settings_overlay_text">You have selected %1$s as your localization.\nWould you like to continue?</string>
+    <string name="country_settings_overlay_confirm_selection">Continue</string>
+    <string name="country_settings_overlay_dismiss">Edit Localization</string>
```
frett: use `confirm` instead of `overlay`, `overlay` is an internal implementation detail of using Circuit, `confirm` refers to it be a confirmation functionality that is part of the country settings.

`country_settings_confirm_description`
`country_settings_confirm_action_continue`
`country_settings_confirm_action_dismiss`

frett (changes requested):
Good first pass, left a couple comments

app/src/main/res/drawable/country_setting_overlay_icon.xml
```diff
@@ -0,0 +1,19 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="59dp"
+    android:height="67dp"
+    android:viewportWidth="59"
+    android:viewportHeight="67">
```
frett: I'm going to take a look at the icon they are using, the mismatched dimensions defined here are probably not what we actually want

frett: disregard this, they actually did wrap it with padding correctly, so keep these dimensions is fine

app/src/main/res/drawable/country_setting_overlay_icon.xml
```diff
@@ -0,0 +1,19 @@
...
+        android:fillColor="#3BA4DB"
```
frett: use `@color/tintable`, this is currently the primary color. By using the `Icon` composable it will automatically tint the icon with the `iconContentColor` property defined on the `AlertDialog`, which you should use the `primary` color for.

app/src/main/res/drawable/country_setting_overlay_icon.xml
```diff
@@ -0,0 +1,19 @@
...
+      />
+    <path
+        android:pathData="M29.5,2.25C34.311,2.25 38.75,6.069 38.75,10.919C38.75,14.491 36.844,16.493 34.966,18.678C33.09,20.86 31.067,23.381 30.239,28.129H28.761C27.933,23.381 25.91,20.86 24.034,18.678C22.156,16.493 20.25,14.491 20.25,10.919C20.25,6.069 24.69,2.25 29.5,2.25ZM29.5,8.958C28.139,8.958 27.063,10.035 27.063,11.333C27.063,12.631 28.14,13.708 29.5,13.708C30.86,13.708 31.938,12.631 31.938,11.333C31.937,10.035 30.861,8.958 29.5,8.958Z"
+        android:strokeWidth="1.5"
+        android:fillColor="#3BA4DB"
+        android:strokeColor="#ffffff"/>
```
frett: hmm, this may prove to be problematic, but it's an issue with how the icon was created. I believe this is a white outline around the place indicator.

The potential issues I see are:
1. by tinting the icon I believe it will also tint the white stroke color, which will just make it all 1 color
2. if it doesn't tint the stroke color this isn't dark theme compatible because it will still be white when you have dark mode enabled

frett: it really needs to be a transparent cutout so the background color bleeds through

frett: but just making it transparent won't work because the globe is positioned behind it and will bleed through anyways. I wonder if there is anything in figma that could modify the icon for us to correctly handle this

tjohnson009: Yes, this is what I saw as well. I wasn't entirely sure how to deal with it. I tried looking into how to make it transparent but couldn't find anything that did exactly what I was picturing.

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsConfirmationOverlay.kt
tjohnson009: Good to know about the X close button at the top. Put it in there just in case but I figured you might say that.

tjohnson009: The issue I was running into with the AlertDialog is that it didn't let me do a ton of spacing on the buttons. For example, the "Edit Localization" button kept wanting to push the text into 2 lines, even with a maxLines = 1 property on it. AlertDialog only really supported a Row if it was in either the confirmButton property or the dismissButton property ( I couldn't find a way to get both of those buttons in a row for the entire AlertDialog if that makes sense). And I was thinking I didn't want to put both buttons in one of those if they had different functions.

frett: the buttons wrapping to 2 rows is standard dialog behavior for material design, this normally shows up for translations that are longer than normal. Don't try to short-circuit it :)

frett:
@tjohnson009 a couple follow up tickets:

- Add a "I prefer not to say" option at the top of the list
- The currently selected country should be at the top of the list of countries

frett (approved):

PR #4388, opened 2026-04-14
GT-3008 Currently Selected Country At Top Of List

tjohnson009:
- **Add selectedCountry to top of countries list - no overlay when clicked**
- **Record updated snapshots**

### DIFF
--- a/app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayout.kt
+++ b/app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayout.kt
@@ -164,18 +164,11 @@ internal fun CountrySettingsLayout(state: UiState, modifier: Modifier = Modifier
                     },
                     modifier = Modifier.clickable {
                         coroutineScope.launch {
-                            if (state.countryCode == country.isoCode) {
-                                eventSink(UiEvent.SelectCountry(country.isoCode))
-                            } else {
-                                val confirmCountrySelection = overlayHost.show(
-                                    CountrySettingsConfirmationOverlay(
-                                        country.displayName ?: country.isoCode,
-                                    ),
-                                )
-                                if (confirmCountrySelection) {
-                                    eventSink(UiEvent.SelectCountry(country.isoCode))
-                                }
+                            if (country.isoCode != state.countryCode) {
+                                val overlay = CountrySettingsConfirmationOverlay(country.displayName ?: country.isoCode)
+                                if (!overlayHost.show(overlay)) return@launch
                             }
+                            eventSink(UiEvent.SelectCountry(country.isoCode))
                         }
                     },
                 )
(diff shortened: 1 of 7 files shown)


app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayout.kt
```diff
@@ -164,13 +164,17 @@ internal fun CountrySettingsLayout(state: UiState, modifier: Modifier = Modifier
...
                         coroutineScope.launch {
-                            val confirmCountrySelection = overlayHost.show(
-                                CountrySettingsConfirmationOverlay(
-                                    country.displayName ?: country.isoCode,
-                                ),
-                            )
-                            if (confirmCountrySelection) {
-                                eventSink(UiEvent.SelectCountry(country.isoCode))
+                            if (state.countryCode == country.isoCode) {
+                                eventSink(UiEvent.NavigateBack)
```
frett: I probably would still trigger the `SelectCountry()` event here for consistency.

frett:
@tjohnson009 you could update the layout tests to have the selected one first in the data you provide, but I'm honestly not worried about that. The main thing the paparazzi tests are meant to do is detect unexpected rendering changes, the actual data it's rendering is meaningless. So having the checkmark visible on 1 row is enough for the test.

Testing the order of the countries to reflect the currently selected country coming first is the responsibility of a presenter test for this

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayout.kt
tjohnson009: Okay, do I need to write some sort of check in it to see if the country is already selected? I just don't know if rerunning the settings.updateCountrySetting(isoCode = event.isoCode) is a waste or not.

frett: no need to have a check to avoid running it, it won't hurt anything to update to the already set value

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsLayout.kt
```diff
@@ -164,13 +164,17 @@ internal fun CountrySettingsLayout(state: UiState, modifier: Modifier = Modifier
...
                                 eventSink(UiEvent.SelectCountry(country.isoCode))
+                            } else {
+                                val confirmCountrySelection = overlayHost.show(
+                                    CountrySettingsConfirmationOverlay(
+                                        country.displayName ?: country.isoCode,
+                                    ),
+                                )
+                                if (confirmCountrySelection) {
+                                    eventSink(UiEvent.SelectCountry(country.isoCode))
+                                }
```
frett:
```kotlin
coroutineScope.launch {
    if (country.isoCode != state.countryCode) {
        val overlay = CountrySettingsConfirmationOverlay(country.displayName ?: country.isoCode)
        if (!overlayHost.show(overlay)) return@launch
    }
    eventSink(UiEvent.SelectCountry(country.isoCode))
}
```

frett: when you start getting branching logic that seems to duplicate some logic, take a step back and think through how you are structuring the logic :)

How the logic grew to be what it is made sense as you implemented the various features, but it was getting a bit messy and can be made much cleaner and easier to understand :)

frett (approved):

PR #4405, opened 2026-04-27
GT-3017 Onboarding Localization

tjohnson009:
- **Add pop result to app language screen to track result**
- **Modify popResult names and add for presenter**
- **Add pop result to navigator.pop in UiEvents**
- **Modify tests to reflect newly added pop result**
- **Add result interface for pop results to send to presenter**
- **Receive newly added pop results in UiEvents**
- **Sync pager to presenter page state via LaunchedEffect and disable swipe on welcome**
- **Add answering navigators and Next event for language/country flow in presenter**
- **Add initial state and Next event tests to OnboardingPresenterTest**
- **Change event passed to prevent block from continue onboarding**
- **Change test to pass with the previous dismiss select language event blocking onboarding**

### DIFF
--- a/app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingLayout.kt
+++ b/app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingLayout.kt
@@ -12,6 +12,7 @@ import androidx.compose.foundation.pager.rememberPagerState
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Scaffold
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
@@ -40,6 +41,14 @@ fun OnboardingLayout(state: OnboardingPresenter.UiState, modifier: Modifier = Mo
 
     val coroutineScope = rememberCoroutineScope()
     val pagerState = rememberPagerState(pageCount = { 4 })
+    LaunchedEffect(state.currentPage) {
+        if (pagerState.currentPage != state.currentPage) {
+            pagerState.animateScrollToPage(
+                state.currentPage,
+                animationSpec = spring(stiffness = Spring.StiffnessMediumLow)
+            )
+        }
+    }
 
     RecordAnalyticsScreen(OnboardingAnalyticsScreenEvent(pagerState.currentPage))
 
@@ -74,6 +83,7 @@ fun OnboardingLayout(state: OnboardingPresenter.UiState, modifier: Modifier = Mo
             HorizontalPager(
                 key = { it },
                 state = pagerState,
+                userScrollEnabled = state.userScrollEnabled,
                 modifier = Modifier
                     .padding(insets)
                     .consumeWindowInsets(insets)
@@ -90,7 +100,7 @@ fun OnboardingLayout(state: OnboardingPresenter.UiState, modifier: Modifier = Mo
 
                 when (it) {
                     0 -> OnboardingWelcomePageLayout(
-                        nextPage = nextPage,
+                        nextPage = { eventSink(UiEvent.Next) },
                         eventSink = eventSink,
                     )
 
--- a/app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingLayout.kt
+++ b/app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingLayout.kt
@@ -8,11 +8,9 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.navigationBarsPadding
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.pager.HorizontalPager
-import androidx.compose.foundation.pager.rememberPagerState
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Scaffold
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
@@ -40,15 +38,7 @@ fun OnboardingLayout(state: OnboardingPresenter.UiState, modifier: Modifier = Mo
     val eventSink by rememberUpdatedState(state.eventSink)
 
     val coroutineScope = rememberCoroutineScope()
-    val pagerState = rememberPagerState(pageCount = { 4 })
-    LaunchedEffect(state.currentPage) {
-        if (pagerState.currentPage != state.currentPage) {
-            pagerState.animateScrollToPage(
-                state.currentPage,
-                animationSpec = spring(stiffness = Spring.StiffnessMediumLow)
-            )
-        }
-    }
+    val pagerState = state.pagerState
 
     RecordAnalyticsScreen(OnboardingAnalyticsScreenEvent(pagerState.currentPage))
 
--- a/app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingLayout.kt
+++ b/app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingLayout.kt
@@ -1,7 +1,5 @@
 package org.cru.godtools.ui.onboarding
 
-import androidx.compose.animation.core.Spring
-import androidx.compose.animation.core.spring
 import androidx.compose.foundation.layout.consumeWindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
@@ -19,7 +17,6 @@ import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.ui.Modifier
 import com.slack.circuit.codegen.annotations.CircuitInject
 import dagger.hilt.components.SingletonComponent
-import kotlinx.coroutines.launch
 import org.ccci.gto.android.common.androidx.compose.material3.ui.appbar.AppBarAction
 import org.ccci.gto.android.common.androidx.compose.material3.ui.appbar.AppBarActionButton
 import org.cru.godtools.R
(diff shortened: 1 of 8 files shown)


app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingLayout.kt
```diff
@@ -40,6 +41,14 @@ fun OnboardingLayout(state: OnboardingPresenter.UiState, modifier: Modifier = Mo
...
     val coroutineScope = rememberCoroutineScope()
     val pagerState = rememberPagerState(pageCount = { 4 })
+    LaunchedEffect(state.currentPage) {
+        if (pagerState.currentPage != state.currentPage) {
+            pagerState.animateScrollToPage(
+                state.currentPage,
+                animationSpec = spring(stiffness = Spring.StiffnessMediumLow)
+            )
+        }
+    }
```
frett: I'm not a big fan of this, it feels like this would be brittle and potentially fire at unexpected times.

Is it possible to lift the `pagerState` into the `UiState` object so you can just trigger the `animateScrollToPage` from the presenter directly?

app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingPresenter.kt
```diff
@@ -24,21 +31,50 @@ class OnboardingPresenter @AssistedInject constructor(
...
+        val nextButtonNavigator = rememberAnsweringNavigator<AppLanguageScreen.Result>(navigator) { result ->
+            if (result is AppLanguageScreen.Result.LanguageSelected) {
+                languageWasSet = true
+                countrySettingsNavigator.goTo(CountrySettingsScreen)
+            }
+        }
         LaunchedEffect(Unit) { settings.setFeatureDiscovered(FEATURE_TUTORIAL_ONBOARDING) }
 
-        return UiState { event ->
+        val eventSink: (UiEvent) -> Unit = { event ->
```
frett: there isn't a need to extract `eventSink` to a variable like this, you can just add the parameters onto the object being created here.

```kotlin
return UiState(
    currentPage = currentPage,
    userScrollEnabled = currentPage > 0,
) { event ->
```

ui/base/src/main/kotlin/org/cru/godtools/base/ui/circuit/screen/AppLanguageScreen.kt
```diff
@@ -1,7 +1,13 @@
...
+import com.slack.circuit.runtime.screen.PopResult
 import com.slack.circuit.runtime.screen.Screen
 import kotlinx.parcelize.Parcelize
 
 @Parcelize
-data object AppLanguageScreen : Screen
+data object AppLanguageScreen : Screen {
+    sealed interface Result : PopResult {
+        @Parcelize data object LanguageSelected : Result
+        @Parcelize data object Dismiss : Result
```
frett: just keeping the same verb tense for the `Result` type

```suggestion
        @Parcelize data object Dismissed : Result
```

app/src/main/kotlin/org/cru/godtools/ui/languages/country/CountrySettingsScreen.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.slack.circuit.runtime.screen.PopResult
+import com.slack.circuit.runtime.screen.Screen
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+data object CountrySettingsScreen : Screen {
+    sealed interface Result : PopResult {
+        @Parcelize data object CountrySelected : Result
+        @Parcelize data object Dismiss : Result
```
frett:
```suggestion
        @Parcelize data object Dismissed : Result
```

app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingPresenter.kt
```diff
@@ -11,34 +21,73 @@ import dagger.assisted.Assisted
...
+                languageWasSet = true
+                countrySettingsNavigator.goTo(CountrySettingsScreen)
+            }
+        }
         LaunchedEffect(Unit) { settings.setFeatureDiscovered(FEATURE_TUTORIAL_ONBOARDING) }
-
-        return UiState { event ->
+        return UiState(
+            pagerState = pagerState,
+            userScrollEnabled = pagerState.currentPage > 0
```
frett: this should also be true when the user has gone through the language and country UI

app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingPresenter.kt
```diff
@@ -11,34 +21,73 @@ import dagger.assisted.Assisted
...
     sealed interface UiEvent : CircuitUiEvent {
         data object ChangeLanguage : UiEvent
+        data object Next : UiEvent
         data object Skip : UiEvent
         data object Finish : UiEvent
     }
 
     @Composable
     override fun present(): UiState {
+        var languageWasSet by rememberSaveable { mutableStateOf(false) }
```
frett, line 58: you should also track if they have set a country

frett: well, have gone through the country settings UI

frett: I'm also not sure what UX the product team is looking for, if it's just that the user has seen the language/country UIs, or that the user has actually made a selection in them

tjohnson009: If I understand it rightly, I think it would only be that they have seen the countrySettings UI, because right now we start out with the countrySettings being null, and we don't force them to pick one. Maybe I can add a: 
hasSeenCountrySettings val and set it to true when applicable.

tjohnson009: or countryWasSet, but that might not be intuitive if the user keeps the "I prefer not to say" option selected

app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingPresenter.kt
```diff
@@ -11,34 +21,82 @@ import dagger.assisted.Assisted
...
+                languageWasSet = true
+                countrySettingsNavigator.goTo(CountrySettingsScreen)
+            }
+        }
         LaunchedEffect(Unit) { settings.setFeatureDiscovered(FEATURE_TUTORIAL_ONBOARDING) }
-
-        return UiState { event ->
+        return UiState(
+            pagerState = pagerState,
+            userScrollEnabled = pagerState.currentPage > 0 || languageWasSet
```
frett:
```suggestion
            userScrollEnabled = (languageWasSet && hasSeenCountrySettings) || pagerState.currentPage > 0
```

app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingPresenter.kt
```diff
@@ -11,34 +21,82 @@ import dagger.assisted.Assisted
...
+                        scope.launch {
+                            pagerState.animateScrollToPage(
+                                1,
+                                animationSpec = spring(stiffness = Spring.StiffnessMediumLow)
+                            )
+                        }
+                    } else {
+                        nextButtonNavigator.goTo(AppLanguageScreen)
+                    }
+                }
```
frett: This could be restructured as a when block that makes more logical sense:

```kotlin
when {
    !languageWasSet -> Language Settings
    !hasSeenCountrySettings - > Country Settings
    else -> next page
}
```

app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingPresenter.kt
```diff
@@ -11,34 +21,82 @@ import dagger.assisted.Assisted
...
+        var hasSeenCountrySettings by rememberSaveable { mutableStateOf(false) }
+        val pagerState = rememberPagerState(pageCount = { 4 })
+        val scope = rememberCoroutineScope()
+        val countrySettingsNavigator = rememberAnsweringNavigator<CountrySettingsScreen.Result>(navigator) {
+            hasSeenCountrySettings = true
+            scope.launch {
+                pagerState.animateScrollToPage(
+                    1,
+                    animationSpec = spring(stiffness = Spring.StiffnessMediumLow)
+                )
```
frett: we should be able to avoid having to make this multi-line by doing one of the following:

- Create an animation spec constant in the companion object for the spring
- create a private extension method you can call on the pagerState to go to the next page.

This also has the benefit of reducing duplication that may lead to future bugs since you are also triggering a scroll animation in the eventSink.

Additionally, if you work it right by improving the `when` block I suggested, you can make it so you always route the "Next Page" button to the presenter and let it trigger the page scroll animation. That will further reduce duplication because the next page trigger will be in a single place then

frett:
```kotlin
class OnboardingPresenter {
    private suspend fun PagerState.navigateToPage(page: Int = currentPage + 1) {
        animateScrollToPage(page, animationSpec = spring(stiffness = Spring.StiffnessMediumLow))
    }
}
```

app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingPresenter.kt
tjohnson009: The reason I left hasCountrySettings off of this conditional was because I didn't think we were forcing the user to pick a country. If this is here, I suppose they could still hit the back button, but I thought enabling the swipe was a good move too.

frett: the problem is that swiping would be enabled as soon as they opened the app language selector and they wouldn't be forced to use next to see the country selector

app/src/main/kotlin/org/cru/godtools/ui/onboarding/OnboardingPresenter.kt
```diff
@@ -11,34 +21,80 @@ import dagger.assisted.Assisted
...
+                    when {
+                        !languageWasSet -> nextButtonNavigator.goTo(AppLanguageScreen)
+
+                        !hasSeenCountrySettings -> countrySettingsNavigator.goTo(CountrySettingsScreen)
+
+                        else -> scope.launch {
+                            pagerState.navigateToPage()
+                        }
+                    }
+                }
```
frett:
```suggestion
                UiEvent.Next -> when {
                    !languageWasSet -> nextButtonNavigator.goTo(AppLanguageScreen)
                    !hasSeenCountrySettings -> countrySettingsNavigator.goTo(CountrySettingsScreen)
                    else -> scope.launch { pagerState.navigateToPage() }
                }
```

frett (approved):

PR #4424, opened 2026-05-12
GT-2601: Download translation files from CDN with SHA-256 verification

frett:
## Summary
- Introduces `ApiConfig` to consolidate API/CDN base URL configuration, replacing the scattered `@Named` string injection
- Adds `CdnApi` (Retrofit) for downloading published translation files from the CDN, with base URLs configured per flavor (`mobilecontent-stage.cru.org` / `mobilecontent.cru.org`)
- Verifies SHA-256 checksum and byte size when storing downloaded files, rejecting any file that doesn't match the manifest metadata
- Prefers the CDN when downloading published translation files, falling back to the existing `TranslationsApi` on 4xx errors or `IOException`

## Test plan
- [ ] New unit tests cover: CDN preferred over API, fallback to API on CDN 404, fallback on `IOException`, both sources failing returns false, correct/wrong SHA-256 accepted/rejected, correct/wrong size accepted/rejected
- [ ] Run `./gradlew :library:download-manager:test` to verify all download manager tests pass
- [ ] Run `./gradlew :build-logic:ktlintCheck ktlintCheck` to verify code style
- [ ] Smoke test on stage build: download a tool and confirm it completes successfully


library/download-manager/src/main/kotlin/org/cru/godtools/downloadmanager/GodToolsDownloadManager.kt
```diff
@@ -294,9 +298,10 @@ class GodToolsDownloadManager @VisibleForTesting internal constructor(
         val relatedFiles = manifest.relatedFiles
         val completedFiles = AtomicLong(0)
         val successful = coroutineScope {
-            relatedFiles.map {
+            relatedFiles.map { file ->
                 async {
-                    downloadTranslationFileIfNecessary(it).also {
+                    val src = file.src ?: return@async true
+                    downloadTranslationFileIfNecessary(src, sha256 = file.checksumSha256, size = file.size).also {
```
tjohnson009, line 304: I don't understand why we're using the sha256 now. Sha256 is encryption right? What do we need to encrypt while getting the translation file?

library/download-manager/src/main/kotlin/org/cru/godtools/downloadmanager/GodToolsDownloadManager.kt
```diff
@@ -309,22 +314,42 @@ class GodToolsDownloadManager @VisibleForTesting internal constructor(
...
 
         return true
     }
 
-    private suspend fun downloadTranslationFileIfNecessary(fileName: String): Boolean = filesMutex.withLock(fileName) {
+    private suspend fun downloadTranslationFileIfNecessary(
+        fileName: String,
+        sha256: String? = null,
+        size: Long? = null,
+    ): Boolean = filesMutex.withLock(fileName) {
```
tjohnson009, line 329: I needed to look up what a mutex is vs a semaphore again. A mutex is basically we are controlling one gate vs several with a semaphore, right? 

Why do we want/use a mutex here vs a semaphore? Is it because we only deal with one file at a time in this context?

library/download-manager/src/main/kotlin/org/cru/godtools/downloadmanager/GodToolsDownloadManager.kt
frett: Sha256 is a checksum. This allows us to verify the file is what we are expecting and not corrupted

tjohnson009 (approved):
Worked up to understanding a good bit of this. Thanks for letting me see

PR #4430, opened 2026-05-26
GT-2980 No Personalized Lessons UI

tjohnson009:
- **Add dataLoaded flag to delay rendering; Add string resources**
- **Create and add NoPersonalizedLessons UI**
- **Modify exclamation icon for NoPersonalizedLessons composable**
- **Rename onClick param for easier readability and remove unused imports**
- **Swap icon to more closely reflect Figma design**
- **Add closer matching icon for NoPersonalizedLessons composable**


frett:
@tjohnson009 you're going to have to clean up the merge conflicts on this

app/src/main/kotlin/org/cru/godtools/ui/dashboard/tools/ToolsLayout.kt
```diff
@@ -49,6 +49,12 @@ internal fun ToolsLayout(state: UiState, modifier: Modifier = Modifier) {
 
     val columnState = rememberLazyListState()
     LaunchedEffect(state.banner?.type) { if (state.banner != null) columnState.animateScrollToItem(0) }
+    val pinLastItem = rememberPinLastItemBottomArrangement()
+    val verticalArrangement = if (state.mode == UiState.Mode.PERSONALIZATION) {
+        pinLastItem
+    } else {
+        Arrangement.Top
+    }
```
frett: merge remanent, this is no longer used since the arrangement is defined inline

app/src/main/kotlin/org/cru/godtools/ui/dashboard/PinLastItemBottomArrangement.kt
```diff
@@ -0,0 +1,23 @@
+package org.cru.godtools.ui.dashboard
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
+import androidx.compose.ui.unit.Density
+
+@Composable
+internal fun rememberPinLastItemBottomArrangement(): Arrangement.Vertical = remember {
```
frett: merge remanent, this was renamed and moved, so this version shouldn't be actively used anywhere

app/src/main/kotlin/org/cru/godtools/ui/dashboard/tools/ToolsLayout.kt
tjohnson009: Yeah I definitely did something wrong in the merge. It looks like your changes didn't make it. I'll investigate

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/NoPersonalizedLessons.kt
```diff
@@ -0,0 +1,56 @@
...
+                text = stringResource(R.string.dashboard_lessons_section_personalized_no_lessons_description),
+                style = MaterialTheme.typography.bodyMedium,
+                textAlign = TextAlign.Center,
+                modifier = Modifier.padding(top = 8.dp),
+            )
+            Button(
+                onClick = onGoToAllLessons,
+                modifier = Modifier.padding(top = 16.dp),
+            ) {
+                Text(stringResource(R.string.dashboard_lessons_section_personalized_no_lessons_button))
```
frett: I would name this based upon the action the text is for, not the fact that it's a button

```suggestion
                Text(stringResource(R.string.dashboard_lessons_section_personalized_no_lessons_action_all_lessons))
```

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/NoPersonalizedLessons.kt
```diff
@@ -0,0 +1,56 @@
...
+    ) {
+        Column(
+            horizontalAlignment = Alignment.CenterHorizontally,
+            modifier = Modifier.padding(24.dp),
+        ) {
+            Icon(
+                painter = painterResource(R.drawable.ic_priority_high),
+                contentDescription = null,
+                modifier = Modifier.size(64.dp),
+            )
```
frett, line 35: just double checking, does the `Icon` composable automatically use the current content color?

(It probably does, but I don't remember off the top of my head)

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsLayout.kt
```diff
@@ -72,6 +72,15 @@ internal fun LessonsLayout(state: UiState, modifier: Modifier = Modifier) {
             )
         }
 
+        if (state.mode == UiState.Mode.PERSONALIZATION && state.dataLoaded && state.lessons.isEmpty()) {
+            item("no-personalized-lessons", "no-personalized-lessons") {
```
frett, line 76: just checking, have you looked into why we are passing the 2 strings to `item()` before, and what each string does?

app/src/main/res/drawable/ic_priority_high.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="48dp"
+    android:height="48dp"
+    android:viewportWidth="960"
+    android:viewportHeight="960">
+  <path
+      android:pathData="M479.88,796.92q-21.78,0 -37.18,-15.51 -15.39,-15.52 -15.39,-37.3 0,-21.78 15.51,-37.18 15.51,-15.39 37.3,-15.39 21.78,0 37.18,15.51 15.39,15.51 15.39,37.3 0,21.78 -15.51,37.18 -15.51,15.39 -37.3,15.39ZM432.69,600.77v-457.69h94.62v457.69h-94.62Z"
+      android:fillColor="#000000"/>
```
frett: just for standardization with the rest of our vector images, use `@color/tintable` for the color.

(This is assuming the icon gets tinted by it's use in the `Icon` composable)

frett (changes requested):
a couple minor comments, I also want to make sure we have screenshot tests of this UI if they don't already exist

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/NoPersonalizedLessons.kt
tjohnson009: Yes, I looked into it and it does.

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsLayout.kt
tjohnson009: I just looked into that. It seems like a unique key to represent the item. It's for a LazyList so I think it helps maintain scroll position when you scroll through items because for a lazy list items are generated as you scroll? I don't understand why there are two of them though, specifically the contentType.

app/src/main/res/drawable/ic_priority_high.xml
tjohnson009: Yes, I hadn't used an Icon in a while, I forgot to do this.

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsLayout.kt
tjohnson009: This makes me wonder if I need to change the no-personalized-lessons item contentType. The key seems fine, but I'm wondering if theres a more semantic or better contentType to go there

        if (state.mode == UiState.Mode.PERSONALIZATION && state.dataLoaded && state.lessons.isEmpty()) {
            item("no-personalized-lessons", "no-personalized-lessons") {
                NoPersonalizedLessons(
                    onGoToAllLessons = { state.eventSink(UiEvent.ChangeMode(UiState.Mode.ALL_LESSONS)) },
                    modifier = Modifier.padding(top = 16.dp, horizontal = MARGIN_LESSONS_LAYOUT_HORIZONTAL)
                )
            }

tjohnson009:
It seems that the NoPersonalizedLessons UI will be covered by a test that is already present in the test suite. Unless we want one for that UI element specifically when I rerecord the screenshots it should be covered as well.

tjohnson009:
Or actually I could add some lessons so that the localization settings box shows with some lessons present as opposed to always with the no personalized lessons UI

app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonsLayout.kt
frett: Leaving both strings there is fine, I can talk through the difference between them at some point

app/src/testDebug/kotlin/org/cru/godtools/ui/dashboard/DashboardLayoutPaparazziTest.kt
```diff
@@ -316,6 +316,15 @@ class DashboardLayoutPaparazziTest(
 
     @Test
     fun `LessonsLayout() - Personalization - Localization Settings Box`() {
+        assumeTrue(locale == null)
+        lessonsState = lessonsState.copy(
+            mode = LessonsPresenter.UiState.Mode.PERSONALIZATION,
+        )
+        snapshotDashboardLayout(state.copy(initialPage = LessonsScreen))
+    }
```
frett, line 325: this test now no longer shows the Localization Settings Box we are trying to test.

Image

frett: include a single tool/lesson so that the box is still visible and doesn't get pushed off the bottom

app/src/testDebug/kotlin/org/cru/godtools/ui/dashboard/DashboardLayoutPaparazziTest.kt
```diff
@@ -316,6 +316,16 @@ class DashboardLayoutPaparazziTest(
 
     @Test
     fun `LessonsLayout() - Personalization - Localization Settings Box`() {
+        assumeTrue(locale == null)
+        lessonsState = lessonsState.copy(
+            mode = LessonsPresenter.UiState.Mode.PERSONALIZATION,
+            lessons = lessonsState.lessons.take(1)
```
frett: this works fine, I believe in this scenario doing `lessonsState.lessons.subList(0, 1)` would convey the meaning of what you are doing better. (I believe there is a `subList()` method available)

frett (commented):
just one minor comment about something that might convey meaning better, and the behavior question I posted in Slack.

app/src/testDebug/kotlin/org/cru/godtools/ui/dashboard/DashboardLayoutPaparazziTest.kt
frett: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/-list/sub-list.html

tjohnson009: Yeah I can see how this method is more true to what is happening.

frett (approved):

PR #4440, opened 2026-05-28
Migrate CategoriesFragment to Compose

frett:
## Summary

- Rewrites `CategoriesFragment` to use Jetpack Compose (`LazyColumn` + `RenderArticleCategory`) in place of DataBinding/RecyclerView
- Removes `CategoriesAdapter` and all associated XML layouts/drawables/styles
- Cleans up unused dependencies (`constraintlayout`, `recyclerview`, `viewBinding`, `gtoSupport.picasso`, etc.) from `article-renderer`

## Test plan

- [ ] Open the article renderer categories screen and verify categories render correctly
- [ ] Tap a category and verify navigation to the articles list works
- [ ] Verify analytics screen event (`SCREEN_CATEGORIES`) fires on resume


ui/article-renderer/src/main/kotlin/org/cru/godtools/article/ui/categories/CategoriesFragment.kt
```diff
@@ -3,56 +3,106 @@ package org.cru.godtools.article.ui.categories
...
+    @Composable
+    private fun CategoriesContent(modifier: Modifier = Modifier) {
+        val manifest by toolDataModel.manifestFlow.collectAsState(null)
+        val categories = manifest?.categories.orEmpty()
+
+        LazyColumn(
+            verticalArrangement = Arrangement.spacedBy(1.dp),
+            modifier = modifier.fillMaxSize(),
+        ) {
+            items(categories, key = { it.id.orEmpty() }) { category ->
```
frett: `Category.id` is nullable (confirmed in the KMP model: `val id: String?`). Using `orEmpty()` as the key means any two categories with `null` IDs both receive key `""`, which will throw a `IllegalArgumentException: Key "" was already used` crash in LazyColumn.

Either omit the `key` parameter entirely (falls back to position — safe), or guard with the object itself as fallback:
```kotlin
items(categories, key = { it.id ?: it }) { category ->
```

frett:
@tjohnson009 this is maintenance work of trying to move away from databinding

tjohnson009:
Why do we want to move away from databinding?

PR #4447, opened 2026-06-03
Replace local SyncTaskRegistry with gto-support-sync library

frett:
## Summary
- Deletes the local `SyncTaskRegistry` class and migrates all three dashboard presenters (`DashboardPresenter`, `LessonsPresenter`, `ToolsPresenter`) to use `rememberSyncTaskRegistry` / `rememberSyncTask` from the `gto-support-sync` library
- Updates tests to use the library `SyncTaskRegistry` directly via `putTag()` instead of the now-`internal` extension property, and deletes `SyncTaskRegistryTest` (covered by the library's own tests)
- Fixes a pre-existing issue in `ToolsPresenter` where the sync locale fell back to `settings.appLanguage` (a one-shot read) — it now uses `settings.appLanguageFlow.collectAsState()` so locale changes are reflected reactively

## Test plan
- [ ] `./gradlew test` passes
- [ ] `./gradlew :build-logic:ktlintCheck ktlintCheck` passes


frett (commented):
## Review

Removes the hand-rolled `SyncTaskRegistry` from `:app` and replaces it with `rememberSyncTaskRegistry` / `rememberSyncTask` from `gto-support-sync` across all three dashboard presenters. Also fixes a pre-existing bug in `ToolsPresenter` where the fallback sync locale was a one-shot read instead of a reactive flow.

### ✅ Looks Good

- **ktlint** — passes clean
- **Presenter patterns** — `DashboardPresenter` correctly calls `syncRegistry.rememberSyncTask { }` directly on the returned registry (not via `circuitContext.rememberSyncTask { }`), which avoids the `DisposableEffect` timing gap where the context tag isn't set until after first composition
- **`LessonsPresenter` / `ToolsPresenter`** — `circuitContext.rememberSyncTask(locale) { }` is the right child-presenter pattern; behavior is equivalent to the old `DisposableEffect` registration
- **Test updates** — `putTag(syncTaskRegistry)` correctly replaces the `internal` extension setter for cross-module test setup; `circuitContext.tag<SyncTaskRegistry>()` correctly replaces the `internal` getter in `DashboardPresenterTest`
- **`SyncTaskRegistryTest` deletion** — correct; the library's own tests cover this behavior
- **`appLanguageFlow` fix** — `collectAsState()` on the `StateFlow` provides a stable initial value immediately; recomposition on locale change is intentional and correct
- **Initialization order in tests** — `syncTaskRegistry` is declared before `circuitContext` in both test classes, so property initialization is safe

PR #4568, opened 2026-08-20
GT-3014 Account for hidden pages in lesson share links & deep links

tjohnson009:
Fixes [GT-3014](https://jira.cru.org/browse/GT-3014)

## Problem

The lesson pager only contains visible pages, but iOS and knowgod.com reference pages by their position within the **full** list of manifest pages (including hidden pages). Android was using the pager index in both directions, so whenever a lesson contains hidden pages (e.g. the Listening Lesson in English goes 0 → 1 → 5):

- share links/QR codes generated on Android pointed at the wrong page on other platforms
- share links generated on iOS/web opened to the wrong page on Android

## Changes

**Share links** (`GT-3014 Build the lesson share link from the current page position`)
- The settled-page tracker now records the current page's `position` within the full list of manifest pages (`currentPagePosition`) instead of the pager index, and the share link is built from that. The tracker also observes the pager's visible pages, so the share link stays correct if a hidden page is revealed mid-lesson.
- `pageReached` (lesson feedback) intentionally keeps its existing pager-index semantics.

**Deep links** (`GT-3014 Resolve lesson deep link page numbers as manifest page positions`)
- Incoming knowgod.com lesson deep link page numbers are now resolved as manifest page positions and converted to a pager index before being handed to the lesson pager.
- Hidden pages resolve to the closest previous visible page (matching the existing lesson resume behavior), since hidden pages are only revealed by content events. Invalid/out-of-range positions resolve to the first page, so malformed URLs degrade safely.

## Testing

- New unit tests for `Manifest.lessonPagerIndexForPagePosition()` covering visible pages, hidden pages, invalid positions, and a hidden first page
- `./gradlew :ui:lesson-renderer:test` and ktlint pass on each commit

## Notes

- Deep linking *to* a hidden page (revealing it) is intentionally not supported — hidden pages are event-driven and `LessonPagerState` re-hides them when they're not the current page. If that's ever wanted (per the open question on the ticket), it would need a shared-library change.
- The tract share link issue from the same Slack thread (always sharing the first page, broken by GT-1446) is a separate ticket and not addressed here.


ui/lesson-renderer/src/main/kotlin/org/cru/godtools/tool/lesson/ui/LessonActivity.kt
```diff
@@ -164,14 +165,18 @@ class LessonActivity :
...
                     val pagerState = lessonPagerState.pagerState
 
                     // record the highest page reached for feedback functionality
+                    // and the current page position for the share link
                     LaunchedEffect(pagerState) {
-                        snapshotFlow { pagerState.settledPage }.collect { page ->
+                        snapshotFlow { pagerState.settledPage to lessonPagerState.pages }.collect { (page, pages) ->
                             dataModel.pageReached.update { maxOf(it, page) }
-                            dataModel.currentPage.value = page
+                            dataModel.currentPagePosition.value = pages.getOrNull(page)?.position ?: 0
```
frett: a better way to handle this is to probably expose a `settledPage: Page` property on the `LessonPagerState` object that is derived from from `pages[pagerState.settledPage]`.

I'm not fully sure if it needs to be nullable or not, `pagerState` is tightly coupled to the rest of `LessonPagerState`, so it should be capped to exactly the number of pages available in `LessonPagerState.pages`. Which means that `.getOrNull()` may not be necessary to protect against an IndexOutOfBoundsException.

You may also want to separate the highest page reached logic from the logic to determine the current position into 2 separate `LaunchedEffects` to keep the 2 flows simple and not have to rely on combining multiple values in a `Pair`

tjohnson009: I've gone back and forth with Claude on this and I'm not sure either. But I think the null check feels like a very small price to pay to avoid a potential crash. I asked Claude if there was ever a scenario where a lesson has no pages, because that doesn't seem like a possibility fro what I can tell and it listed a few, but I don't know how realistic it sounds.

tjohnson009: I'll get on the other things first, and then let me know where you land the plane on the nullable part. I think it is a small price to pay, but I might not be seeing this fully

frett: I didn't post a followup, but thinking about it more we do want to keep it nullable because of the empty pages scenario.

I'm about to open a PR on godtools-shared with a settledPage: LessonPage? property you can use

tjohnson009: Okay cool. I was mid-flight on doing that, but I'll tell Claude to pause and then look for it in a bit.

frett: https://github.com/CruGlobal/kotlin-mpp-godtools-tool-parser/pull/1234

tjohnson009: Done in 74e758c46 — consumed `settledPage` from CruGlobal/kotlin-mpp-godtools-tool-parser#1234 (verified against snapshot build 58) and split the page tracking into 2 separate `LaunchedEffect`s.

ui/lesson-renderer/src/main/kotlin/org/cru/godtools/tool/lesson/util/LessonPageUtils.kt
```diff
@@ -0,0 +1,17 @@
...
+ * index of the corresponding page within the initially visible lesson pages. Hidden pages resolve to the closest
+ * previous visible page, and invalid positions resolve to the first page.
+ */
+internal fun Manifest.lessonPagerIndexForPagePosition(position: Int): Int {
+    val page = pages.getOrNull(position)
+        ?.let { page -> generateSequence(page) { it.previousPage }.firstOrNull { !it.isHidden } }
+    return pages.filterIsInstance<LessonPage>().filterNot { it.isHidden }
+        .indexOfFirst { it == page }
+        .coerceAtLeast(0)
+}
```
frett: I get why this exists, but it feels like it's trying to pound a square peg into a round hole.

You are getting a deep link to a specific page, but then trying to find a page that will actually show up with how the LessonPagerState currently works.

This goes against one of the things that came up as I was pressing the product team about if they wanted to support deep links to hidden pages. Paraphrased: "I want to be able to hit share on some content and have that content sent to someone else". In that scenario this code wouldn't do that, it would land the user on a previous page, not the explicit page shared

frett: what you probably need to do is expand or overload `rememberLessonPagerState` to take a `LessonPage` for the `initialPage` and setup the internal state properly for that

frett (approved):

PR #4583, opened 2026-09-01
GT-2786 Replace deprecated TabRow with SecondaryTabRow

tjohnson009:
Resolves [GT-2786](https://jira.cru.org/browse/GT-2786)

Migrates the two remaining usages of the deprecated Material3 `TabRow` composable (`AccountLayout` and `ToolDetailsLayout`) to `SecondaryTabRow`, which renders the same full-width secondary indicator. The custom `pagerTabIndicatorOffset` indicator from gto-support (code inherited from the discontinued Accompanist library) is no longer needed and its usages are removed.

## Behavior change to be aware of

`SecondaryTabRow` uses its built-in indicator, which spring-animates between tabs when the selected page changes. The previous custom indicator tracked the pager's scroll offset fractionally, so it slid pixel-for-pixel with the finger during a swipe. If we want to keep the fractional tracking behavior, it would need a new indicator helper in gto-support built on the `TabIndicatorScope` API — happy to do that as a follow-up if preferred.

## Visual parity

The static render is visually identical: verifying the pre-existing `ToolDetailsLayout` goldens against the migrated code failed only with a subpixel shift of the tab label (the indicator itself had zero delta). Goldens were re-recorded via the Record Snapshots workflow — the 30 re-recorded ToolDetails images correspond exactly to the 30 previously-failing comparisons, and no unrelated goldens changed.

## Tests

- New `AccountLayoutPaparazziTest` covering `AccountLayoutHeader` (both tabs selected), which previously had no snapshot coverage. `AccountLayoutHeader` is now `internal` + `@VisibleForTesting` to make it snapshotable, since `AccountLayout` itself requires a live ViewModel.
- `ToolDetailsLayout` already had full Paparazzi coverage; its goldens were re-recorded.
- `verifyPaparazzi` passes locally for both classes against the recorded goldens; app unit tests and ktlint pass.


app/src/testDebug/kotlin/org/cru/godtools/ui/account/AccountLayoutPaparazziTest.kt
```diff
@@ -0,0 +1,40 @@
...
+    @Test
+    fun `AccountLayoutHeader() - Global Activity tab selected`() {
+        snapshot {
+            AccountLayoutHeader(
+                user = user,
+                pages = pages,
+                pagerState = rememberPagerState(initialPage = 1) { pages.size },
+            )
+        }
+    }
```
frett, line 39: I would prefer that these paparazzi tests were for the entire `AccountLayout`, and not just the header.

Have we converted `AccountLayout` to Circuit yet? if not you should have claude check for a jira ticket or create a jira ticket to do that migration. That conversion shouldn't happen in this PR, it should be a separate unit of work.

app/src/main/kotlin/org/cru/godtools/ui/account/AccountLayout.kt
```diff
@@ -111,8 +110,9 @@ internal fun AccountLayout(viewModel: AccountViewModel = viewModel(), onEvent: (
 }
 
 @Composable
+@VisibleForTesting
 @OptIn(ExperimentalFoundationApi::class, ExperimentalMaterial3Api::class)
-private fun AccountLayoutHeader(
+internal fun AccountLayoutHeader(
```
frett, line 115: If we can generate the AccountLayout paparazzi tests instead of testing the header only, then this can stay private.

Although, thinking about it more I believe the badges we display in the `AccountLayout` are tested independently, that should be changed too to be part of an overall `AccountLayout` paparazzi tests.

Maybe a better solution is to create another followup jira ticket for consolidating the `AccountLayout` paparazzi tests to be full screen and not just components of the screen. (Some component tests might remain to account for visualizing component state)

frett: let's just create the follow up jira ticket to consolidate the AccountLayout screenshot tests. You can probably ask Claude to generate the jira ticket for you

app/src/testDebug/kotlin/org/cru/godtools/ui/account/AccountLayoutPaparazziTest.kt
frett: double check this and create a jira ticket for converting AccountLayout to Circuit if necessary, you can ask claude to do this for you.

(There is a jira MCP server within our organization account that makes it easy for Claude to create tickets)

frett (approved):
This is good to merge, there are potentially 2 followup jira tickets that should be created

PR #4586, opened 2026-09-02
GT-3101 Hide the Featured Tools section without country specific data

tjohnson009:
Resolves [GT-3101](https://jira.cru.org/browse/GT-3101)

Per the MVP Content Data Flow decision tree in the [Figma design](https://www.figma.com/design/gdsgXh4cAeBAVrJzH7YoEOU6/GodTools-app-master-file?node-id=14815-135), the Featured Tools section on the personalized dashboard should only be visible when the user has selected a country **and** featured data exists for their language & country combo. Every branch of the tree where no country is selected, or no featured data exists for the combo, ends in "Featured Section Hidden".

## Changes

- `FeaturedToolsFlowProducer` no longer falls back to the language-only featured list (which aggregates featured tools across all countries) when the country-specific list is empty. It now emits an empty list when no country is selected, and only the exact language+country featured list otherwise.
- The section already hides itself when the list is empty via the existing `isNotEmpty()` check in `ToolsLayout`, so no UI changes were needed.
- Tests updated: the fallback test is replaced with two tests pinning the new behavior (no country → empty without querying; country with no data → empty without falling back).

## Note for the sync side (GT-3011)

`ToolsPresenter` still triggers `syncFeaturedTools(locale, country)` when no country is selected, which downloads the language-only featured data into a bucket nothing reads anymore. Left untouched here since sync behavior belongs to GT-3011 — flagging in case it's worth pruning there.

### DIFF
--- a/app/src/test/kotlin/org/cru/godtools/ui/dashboard/tools/FeaturedToolsFlowProducerTest.kt
+++ b/app/src/test/kotlin/org/cru/godtools/ui/dashboard/tools/FeaturedToolsFlowProducerTest.kt
@@ -72,6 +72,7 @@ class FeaturedToolsFlowProducerTest {
     // region PERSONALIZATION mode
     @Test
     fun `getFlow - Personalization - uses getFeaturedToolsFlow`() = runTest {
+        countryFlow.value = "US"
         producer.getFlow(mode = Mode.PERSONALIZATION).first()
         verify { toolsRepository.getFeaturedToolsFlow(any(), any()) }
         verify(exactly = 0) { toolsRepository.getNormalToolsFlow() }
@@ -80,12 +81,14 @@ class FeaturedToolsFlowProducerTest {
     @Test
     fun `getFlow - Personalization - uses appLanguageFlow when no language provided`() = runTest {
         appLanguageFlow.value = Locale.FRENCH
+        countryFlow.value = "US"
         producer.getFlow(mode = Mode.PERSONALIZATION).first()
         verify { toolsRepository.getFeaturedToolsFlow(Locale.FRENCH, any()) }
     }
 
     @Test
     fun `getFlow - Personalization - uses provided language`() = runTest {
+        countryFlow.value = "US"
         producer.getFlow(mode = Mode.PERSONALIZATION, language = Locale.GERMAN).first()
         verify { toolsRepository.getFeaturedToolsFlow(Locale.GERMAN, any()) }
         verify(exactly = 0) { toolsRepository.getFeaturedToolsFlow(Locale.ENGLISH, any()) }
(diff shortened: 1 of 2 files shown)


tjohnson009 (commented):
## PR Review: GT-3101 (self-review, pre-push)

### Summary
Removes the language-only fallback from `FeaturedToolsFlowProducer` so the personalized featured section shows only exact language+country featured data — emitting an empty list (section hidden by the layout's existing `isNotEmpty` check) when no country is selected or the combo has no data. Behavior verified branch-by-branch against the MVP Content Data Flow decision tree in Figma.

### ✅ Looks Good
- Repo-wide ktlint passes; `FeaturedToolsFlowProducerTest` 14/14; full `:app` unit suite passes (nothing else depended on the fallback)
- All three flowchart branches map to code paths, each pinned by a test — including two negative-space assertions (`verify(exactly = 0)`) proving the fallback query is gone and no query fires without a country
- Turbine for flow assertions, explicit mocks, ticket-referenced test names
- Incidental improvement: with no country selected the flow emits immediately instead of waiting on a DB query

### ⚠️ Minor
- `ToolsPresenter` still runs `syncFeaturedTools(locale, country)` with a null country, downloading language-only featured data into a bucket nothing reads anymore. Not touched here — sync belongs to GT-3011, so flagging for that work instead (also noted in the PR description).

### ❌ Must Fix
- None

**Verdict: APPROVE** (posted as comment — self-review)

app/src/test/kotlin/org/cru/godtools/ui/dashboard/tools/FeaturedToolsFlowProducerTest.kt
```diff
@@ -72,6 +72,7 @@ class FeaturedToolsFlowProducerTest {
     // region PERSONALIZATION mode
     @Test
     fun `getFlow - Personalization - uses getFeaturedToolsFlow`() = runTest {
+        countryFlow.value = "US"
```
frett: I feel like a minor cleanup could be defaulting countryFlow to `US` since this producer now only works when there is a country. The corner-cases that test when country is null should be the ones that set the value to `null` before executing the test

frett (changes requested):
just a minor tweak to the tests

app/src/test/kotlin/org/cru/godtools/ui/dashboard/tools/FeaturedToolsFlowProducerTest.kt
tjohnson009: Done in ada10b62d — `countryFlow` now defaults to `"US"` and only the no-country corner cases set it to `null` (including `updates when country changes`, which now starts from an explicit `null`). All 14 tests pass.

🤖 [Claude Code](https://claude.ai/code)

frett (approved):

PR #4587, opened 2026-09-03
GT-3102 Adjust the personalized tools language filter

tjohnson009:
Resolves [GT-3102](https://jira.cru.org/browse/GT-3102)

Gives the Personalized tools language filter its own persisted setting so the Personalized and All Tools filter selections are independent, per the ticket spec and the [Figma design](https://www.figma.com/design/gdsgXh4cAeBAVrJzH7YoEOU6/GodTools-app-master-file?node-id=14815-135).

## Changes

- New `dashboardPersonalizedFilterLocale` setting. Unset resolves to the app language, which satisfies the first-run default with no migration — while the existing All Tools setting keeps its unset = "Any language" behavior.
- `DefaultToolFiltersStateProducer` is mode-aware: it reads/writes the per-tab setting, resolves the personalized default, and omits the "Any language" item in personalized mode.
- `LanguageFilter` hides the "# Tools available" supporting text in personalized mode, and never renders the "Any language" label there (including the brief state before the stored language loads, matching how Lessons behaves).
- The Personalization Paparazzi scenarios now set a selected language to reflect the realistic filter state — those goldens are re-recorded via the Record Snapshots workflow.

## Intentionally unchanged

- The category filter remains shared between tabs — the ticket only covers the language filter. Asking Jillian whether the category dropdown's counts should also be hidden, since the "unwieldy number" rationale could apply there too.
- The Figma mockup still shows "Any language" on the personalized button; the ticket text explicitly supersedes that.
- The filter's position (between Featured and the ranked list) already matched the design — no layout change needed.

### DIFF
--- a/app/src/main/kotlin/org/cru/godtools/ui/dashboard/tools/ToolFilters.kt
+++ b/app/src/main/kotlin/org/cru/godtools/ui/dashboard/tools/ToolFilters.kt
@@ -37,6 +37,7 @@ import org.cru.godtools.ui.dashboard.filters.FilterMenu
 import org.cru.godtools.ui.dashboard.filters.FilterMenuItem
 import org.cru.godtools.ui.dashboard.filters.LazyFilterMenu
 import org.cru.godtools.ui.dashboard.tools.ToolFiltersStateProducer.Filters
+import org.cru.godtools.ui.dashboard.tools.ToolsPresenter.UiState.Mode
 import org.jetbrains.annotations.VisibleForTesting
 
 private val DROPDOWN_MAX_HEIGHT = 700.dp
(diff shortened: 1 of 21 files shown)


tjohnson009 (commented):
## PR Review: GT-3102 (self-review, pre-push)

### Summary
Per-tab persisted language filter selection, app-language default, no "Any language" option, and no availability counts in personalized mode — every ticket bullet pinned by a test.

### ✅ Looks Good
- Repo-wide ktlint passes; both app unit-test variants and `library:base` tests pass
- Per-mode setting writes guarded by cross-contamination assertions (`coVerify(exactly = 0)` on the opposite key)
- The new Settings pair is an exact pattern-copy of its sibling, with a mirrored `SettingsTest`
- The "Any language" string can never render in personalized mode, including the pre-load blink (matches Lessons)
- Paparazzi Personalization scenarios updated to the realistic selected-language state; All Tools goldens verified untouched

### ⚠️ Minor
- `ToolFiltersTest` uses the file's pre-existing `createComposeRule()` pattern rather than `runComposeUiTest` — converting the file is a separate chore; the new tests follow local convention.
- The 13 Personalization goldens fail verify until the Record Snapshots workflow commit lands on this branch (workflow triggered).

### ❌ Must Fix
- None

**Verdict: APPROVE** (posted as comment — self-review)

PR #4588, opened 2026-09-04
Generate share links for CYOA tools (GT-3080)

tjohnson009:
## Summary
CYOA tools currently have no share link at all — `CyoaActivity` inherits the empty `shareLinkUriLiveData` from `BaseToolActivity`, so the share button never appears. This adds outgoing share link generation matching the knowgod.com URL pattern the incoming deep link parser (`CyoaDeepLink`) already supports:

```
https://knowgod.com/{locale}/tool/v2/{tool}/{page}?icid=gtshare
```

## Changes
- Track the currently displayed page fragment in `pageFragmentLiveData`, updated via `FragmentLifecycleCallbacks` (covers forward navigation, back navigation, and page dismissal)
- Override `shareLinkUriLiveData` to combine the active manifest with the current page, so the link updates on both language changes and page navigation
- The share menu item and QR code action appear automatically via the existing `BaseToolActivity` wiring once the link is non-null

The `{position}` path segment supported by `CyoaDeepLink` for card/page-collection pages is intentionally not included — the ticket's example URL omits it, and it can be added as a follow-up if needed.

## Tests
- Share link contains the current page id
- Link updates when navigating forward and back
- Generated links round-trip through `CyoaDeepLink.parseKnowGodDeepLink` (same tool/page/locale)
- No link before a manifest is loaded

🎫 [GT-3080](https://jira.cru.org/browse/GT-3080)

### DIFF
--- a/ui/cyoa-renderer/src/main/kotlin/org/cru/godtools/tool/cyoa/ui/CyoaActivity.kt
+++ b/ui/cyoa-renderer/src/main/kotlin/org/cru/godtools/tool/cyoa/ui/CyoaActivity.kt
@@ -6,19 +6,26 @@ import android.os.Bundle
 import android.view.MenuItem
 import androidx.activity.viewModels
 import androidx.annotation.VisibleForTesting
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentManager
 import androidx.fragment.app.FragmentManager.POP_BACK_STACK_INCLUSIVE
 import androidx.fragment.app.commit
+import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.asFlow
 import androidx.lifecycle.asLiveData
+import androidx.lifecycle.switchMap
 import dagger.hilt.android.AndroidEntryPoint
 import java.util.Locale
 import kotlinx.coroutines.flow.combine
 import org.ccci.gto.android.common.androidx.fragment.app.backStackEntries
 import org.ccci.gto.android.common.androidx.fragment.app.hasPendingActions
+import org.ccci.gto.android.common.androidx.lifecycle.ImmutableLiveData
+import org.ccci.gto.android.common.androidx.lifecycle.combineWith
 import org.ccci.gto.android.common.androidx.lifecycle.toggleValue
 import org.cru.godtools.base.HOST_DYNALINKS
 import org.cru.godtools.base.HOST_GODTOOLSAPP_COM
 import org.cru.godtools.base.SCHEME_GODTOOLS
+import org.cru.godtools.base.URI_SHARE_BASE
 import org.cru.godtools.base.tool.activity.MultiLanguageToolActivity
 import org.cru.godtools.base.tool.model.Event
 import org.cru.godtools.shared.tool.parser.model.Manifest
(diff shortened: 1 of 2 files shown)


ui/cyoa-renderer/src/main/kotlin/org/cru/godtools/tool/cyoa/ui/CyoaActivity.kt
```diff
@@ -181,6 +196,31 @@ class CyoaActivity :
...
+
+    private val activePageLiveData by lazy {
+        pageFragmentLiveData.switchMap { it?.page ?: ImmutableLiveData<Page?>(null) }
+    }
+
+    private fun Manifest.buildShareLink(page: Page?): Uri.Builder? {
+        val tool = code ?: return null
+        val locale = locale ?: return null
+        return URI_SHARE_BASE.buildUpon()
+            .appendEncodedPath(locale.toString().lowercase(Locale.ENGLISH))
```
frett, line 215: `locale.toString()` might be wrong for locales with regions, if it's a java `Locale` object you will want to use `locale.toLanguageTag()`, if it's the fluidsonic `Locale` object, I'm not totally sure what format it uses for `.toString()`

ui/cyoa-renderer/src/main/kotlin/org/cru/godtools/tool/cyoa/ui/CyoaActivity.kt
```diff
@@ -190,6 +230,13 @@ class CyoaActivity :
         }
     private val activePage get() = pageFragment?.page?.value
 
+    private val pageFragmentLiveData = MutableLiveData<CyoaPageFragment<*, *>?>(null)
+
+    private fun updatePageFragmentLiveData() {
+        val fragment = supportFragmentManager.primaryNavigationFragment as? CyoaPageFragment<*, *>
+        if (pageFragmentLiveData.value !== fragment) pageFragmentLiveData.value = fragment
+    }
```
frett, line 238: you shouldn't be storing the `fragment` in a `LiveData`. I don't think your current implementation would leak anything, but storing an active fragment like this has potential to leak the fragment if it's not handled carefully.

You should probably just store what page the fragment is for

ui/cyoa-renderer/src/main/kotlin/org/cru/godtools/tool/cyoa/ui/CyoaActivity.kt
```diff
@@ -181,6 +196,31 @@ class CyoaActivity :
...
+
+    private fun Manifest.buildShareLink(page: Page?): Uri.Builder? {
+        val tool = code ?: return null
+        val locale = locale ?: return null
+        return URI_SHARE_BASE.buildUpon()
+            .appendEncodedPath(locale.toString().lowercase(Locale.ENGLISH))
+            .appendPath("tool")
+            .appendPath("v2")
+            .appendPath(tool)
+            .apply { page?.let { appendPath(it.id) } }
```
frett, line 219: I don't think you are properly handling page collection pages here or card collection pages

frett:
This might need to be put on hold, I posted a message in the godtools channel about the web not supporting page collection pages, which means that those pages would be shared incorrectly.

tjohnson009:
Going to wait in this until further notice

PR #4589, opened 2026-09-04
Adjust personalized lessons language filter (GT-3097)

tjohnson009:
## Summary
The Lessons page currently has a single language filter shared by both the Personalized and All Lessons tabs, held only in memory (reset on app restart, and reset whenever the app language changes). This makes the two filters independent and persistent per GT-3097:

- **Independent per tab** — the selection on Personalized and All Lessons no longer affect each other
- **Persisted** — each tab's selection is stored in DataStore (`dashboardLessonsFilterLocale` / `dashboardPersonalizedLessonsFilterLocale`) and survives app restarts. Until the user makes a selection, each tab defaults to the app language
- **Hidden counts** — the "# Lessons available" phrase is hidden in the language dropdown on Personalized (still shown on All Lessons). Counts are still used internally to hide languages with no lessons
- **Layout prep for GT-3096** — `LessonFilters` moved out of the header item into its own LazyColumn item, so the Featured Lessons section can be inserted between the header and the filter without restructuring

## Behavior change to note
Previously, changing the app language reset the filter to the new app language. Now an explicit user selection persists through app-language changes (matching the ticket's persistence requirement and the GT-3102 behavior for tools). A tab with no explicit selection still follows the app language.

## Known minor
On a tab switch, the filter briefly shows the app language until DataStore delivers the stored selection — same class of transient accepted in GT-3102 (#4587).

## Coordination
The `Settings.kt`/`SettingsTest.kt` additions sit adjacent to GT-3102's (#4587); whichever PR merges second will have a trivial keep-both conflict.

## Tests
- `SettingsTest`: save/read/clear round-trips for both new keys
- `LessonsPresenterTest`: independent selection per mode (with per-key write verification), follows app language when unset, keeps user selection on app-language change, persistence through state save & restore
- `LessonsLayoutTest`: counts hidden in Personalized dropdown, shown in All Lessons
- `verifyPaparazzi` passes — the layout split is pixel-identical, no golden changes

🎫 [GT-3097](https://jira.cru.org/browse/GT-3097)

### DIFF
--- a/app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
+++ b/app/src/main/kotlin/org/cru/godtools/ui/dashboard/lessons/LessonFilters.kt
@@ -15,6 +15,7 @@ import org.cru.godtools.R
 import org.cru.godtools.base.LocalAppLanguage
 import org.cru.godtools.ui.common.LanguageName
 import org.cru.godtools.ui.dashboard.filters.LazyFilterMenu
+import org.cru.godtools.ui.dashboard.lessons.LessonsPresenter.UiState.Mode
 
 @Composable
 @OptIn(ExperimentalLayoutApi::class)
@@ -33,7 +34,12 @@ internal fun LessonFilters(state: LessonsPresenter.UiState, modifier: Modifier =
             itemKey = { (it) -> it.code },
             itemLabel = { LanguageName(it.item) },
             itemSupportingText = { (_, count) ->
-                pluralStringResource(R.plurals.dashboard_lessons_section_filter_available_lessons, count, count)
+                when (state.mode) {
+                    Mode.PERSONALIZATION -> null
+
+                    Mode.ALL_LESSONS ->
+                        pluralStringResource(R.plurals.dashboard_lessons_section_filter_available_lessons, count, count)
+                }
             },
             modifier = Modifier
                 .weight(1f)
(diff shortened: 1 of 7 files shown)


tjohnson009:
## PR Review (pre-merge self-review via /pr-review)

### Summary
Makes the Lessons language filter independent per tab and persisted across app restarts (two new DataStore keys defaulting to app language), hides the "# Lessons available" dropdown counts on Personalized, and splits the filter into its own LazyColumn item to leave an insertion slot for GT-3096's Featured section.

### ✅ Looks Good
- **ktlint**: clean (full `ktlintCheck` passed)
- **Presenter conventions**: coroutine launched inside the `eventSink` callback (not bare in `present()`); the write uses `launch(start = UNDISPATCHED) { withContext(NonCancellable) { … } }` — the project-mandated pattern for committing state on a user action, so a save can't be dropped if composition disposes mid-write
- **Flow scoping**: `remember(mode)` rebuilds the stored-locale flow on tab change, and `remember(selectedLocaleFlow)` re-keys the selected-item chain off it — no stale-flow capture
- **Settings**: new keys/accessors are exact structural copies of the existing `dashboardFilterLocale` idiom; `null` = follow-app-language gives the "default on first download" requirement for free
- **Tests at all three layers**: storage round-trips (`SettingsTest`), per-mode independence with write-key verification plus follows/keeps app-language behavior (`LessonsPresenterTest`), count visibility per mode (`LessonsLayoutTest`)
- **Layout split is inert**: `verifyPaparazzi` passes — pixel-identical; the float-to-bottom arrangement counts from the list end, so the extra item doesn't affect it
- **Hygiene**: 7 files, all on-topic; full app + base unit tests green

### ⚠️ Minor (accepted, not changed)
- `LessonsPresenter.kt` — on a tab switch, `collectAsState(Language(settings.appLanguage))` shows the app language for a frame or two until DataStore delivers the stored locale — a brief filter-label blink for users whose saved choice differs from the app language. Only visible on tab switch; same class of transient accepted in GT-3102 (#4587), so shipping as-is
- Coordination: `Settings.kt`/`SettingsTest.kt` additions sit adjacent to GT-3102's (#4587) — whichever merges second resolves a trivial keep-both conflict

### ❌ Must Fix
- None

### Verdict
**APPROVE** — faithful to the ticket and the GT-3102 precedent, tested at every layer, zero visual regressions, and it clears the path for GT-3096.

PR #4592, opened 2026-09-08
GT-3107 Convert AccountLayout to Circuit

tjohnson009:
Resolves [GT-3107](https://jira.cru.org/browse/GT-3107)

Converts the Account screen to a Circuit Presenter/UI pair, following the same structure as the Dashboard and Tool Details conversions. Follow-up from the review of #4583.

## Changes

- **`AccountScreen`** — new `@Parcelize data object` `ParcelableScreen`.
- **`AccountPresenter`** — replaces `AccountViewModel`, with `UiState`/`UiEvent` nested in the presenter (`internal`, `@ConsistentCopyVisibility`) matching `DashboardPresenter`. It also absorbs `AccountActivityViewModel` and `GlobalActivityViewModel` (each was a single `stateIn` flow), so the child pager pages are driven from the same state object. Syncing uses the shared `SyncTaskRegistry`/`SyncTracker` mechanism from gto-support — the same pattern as `DashboardPresenter` — so the initial sync fires on task registration, pull-to-refresh flows through `triggerSyncTasks(force = true)`, and sync failures are caught and logged by `SyncTracker`. Up navigation is `navigator.pop()`.
- **`AccountLayout`** — now a `@CircuitInject` UI that is a pure function of `UiState`, wrapped in `DrawerMenuLayout` (preserving the drawer previously provided by `AccountActivity`). Rendering is unchanged — the pre-existing `AccountLayoutHeader` Paparazzi goldens verify without modification. `AccountLayoutHeader` also gained a `modifier` parameter.
- **Navigation** — the drawer menu's Profile item launches `startCircuitActivity(AccountScreen)` through the shared `CircuitActivity` host. `AccountActivity`, `startAccountActivity()`, `AccountLayoutEvent`, and the manifest entry are removed.

## Why child ViewModels were converted in the same pass

`AccountActivityLayout`/`GlobalActivityLayout` previously instantiated ViewModels inside composition, which is what made the assembled screen unrenderable in Paparazzi. With the whole screen state-driven, GT-3108 (consolidating the Account Paparazzi tests to full-screen snapshots) is unblocked.

## Tests

- New `AccountPresenterTest` covering state production (user, pages incl. remote-config gating, activity data, drawer), the launch-time sync, forced sync via pull-to-refresh, and up navigation. Sync tests gate the mocked sync service on a `CompletableDeferred` so the `isSyncRunning` transitions are asserted deterministically.
- `verifyPaparazzi` passes against the existing goldens; full app unit tests and ktlint pass.

https://claude.ai/code/session_01Mm8jFZh3uaPNENLcbnuxow


app/src/main/kotlin/org/cru/godtools/ui/account/AccountPresenter.kt
```diff
@@ -0,0 +1,106 @@
...
+                    add(AccountPage.ACTIVITY)
+                    if (remoteConfig.getBoolean(CONFIG_UI_GLOBAL_ACTIVITY_ENABLED)) add(AccountPage.GLOBAL_ACTIVITY)
+                }.toImmutableList()
+            },
+            userActivity = remember { userActivityManager.userActivityFlow }
+                .collectAsState(UserActivity(emptyMap())).value,
+            globalActivity = GlobalActivityScreen.UiState(
+                activity = remember { globalActivityRepository.getGlobalActivityFlow() }
+                    .collectAsState(GlobalActivityAnalytics()).value
+            ),
```
frett, line 82: let's put these changes on hold, I think I want global activity and account activity to have their own sub-presenters and not just be wrapped up into a large Account Presenter

frett: I need to do some research to see if SubCircuit fits this use case or not

